PR #116, opened 2015-01-08
Fix display when there are no products / orders

ShayneP:
Changing `.blank?` to `.none?` so that the proper table shows on the default index page when there are no products or no orders.

Changes this : 

... to this : 

@jamesmacaulay @silverstreaked

### DIFF
--- a/lib/generators/shopify_app/templates/app/views/home/index.html.erb
+++ b/lib/generators/shopify_app/templates/app/views/home/index.html.erb
@@ -11,21 +11,21 @@
 
 <div class="row">
   <div class="span3">
-    <div class="well">  
+    <div class="well">
       <h3>Check out the <code>home_controller</code></h3>
       <hr style="margin:2px 0 7px"/>
       <p>The controller in this demo application fetches the newest 5 orders and products and makes them available as instance variables <code>@orders</code> and <code>@products</code></p>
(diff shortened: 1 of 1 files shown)


csaunders:
A+++++

silverstreaked:
:+1:

jamesmacaulay:
Great work :+1: 

I'm confused why `blank?` doesn't work though, doesn't `blank?` return true for empty arrays? Or is `@products` some ActiveRecord collection object that behaves a bit differently?

ShayneP:
Yeah - @products is a ActiveRecord collection object that also includes the original params and name of the resource class, so `blank?` always returns false.

jamesmacaulay:
Cool, good to merge :+1: 

I see now that they are actually [_ActiveResource_ collections](https://github.com/rails/activeresource/blob/816faef88cc0d17c27e27feefce17240cd118348/lib/active_resource/collection.rb), which makes it even less surprising that they wouldn't  behave like arrays :/

ShayneP:
Ah, right. Makes sense! Merged and closed.

ckhatton:
Good info to know about the difference of `blank?` between ActiveRecord and ActiveResource, I didn't know, thanks! :octocat:

PR #122, opened 2015-03-06
Solves Error: A copy of Shop has been removed from the module tree but is still active!

shalilawaley:
when ShopifySessionRepository.storage = Shop is enabled in
    config/initializers/shopify_session_repository.rb
    and the Shop model is modified
    it gives error :
    ArgumentError (A copy of Shop has been removed from the module tree but is still active!)
    in development.
    
    to solve the error Shop is replaced with identifier self.
    this solves the error in a simple manner

PS: I have never sent pull request. So, please merge to the branch you think suitable. (If mergeable)

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,13 +88,13 @@ Your ActiveRecord model would look something like this:
 ```ruby
 class Shop < ActiveRecord::Base
   def self.store(session)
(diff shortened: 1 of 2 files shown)


csaunders:
Cool beans.

PR #124, opened 2015-04-19
Make ShopifyApp a proper Rails Engine

kevinhughes27:
This PR effectively re-writes this gem as a proper Rails engine (major version bump to 6.0.0). I started this last hack days and found time this weekend to :shipit: 

Why?
-------
* Enable some degree of code sharing from this gem - the old generator simply generated code once and then individual app owners had to update code. This got us once when there was an error in `sanitize_shop_param`. There are still apps running the wrong function I'm sure
* Rails engines are pretty powerful and this PR is really only the beginning of what we and our community can do here. I am looking forward to PRs to this repo which can really improve Shopify app development

Changes
------------
there are a lot...

* Create modular generators. Install is essentially what we had before except it uses the engine version of files. Shop model generates a shop model the way we typically do it. The other generators copy defaults to the host app which will override the engine files
* Moves a lot of code from generated files into concerns that are part of the lib. This lets us do code re-use
* Adds a shopify_app.rb initializer for users of the gem
* Generates embedded apps now by default!
* added cmd args to the generator (like whether it should be embedded)
* updates the readme and changelog
* adds tests
* adds travis

Upgrading existing shopify_apps is pretty easy (especially since there were no tests in this repo all of our apps usually have a full suite covering what we expect here). Basically you just need to bump the version and then look at your controllers include the concern and remove any methods that are now included in this gem. I did this a while back for Qbo and it was super easy:
https://github.com/Shopify/qbo/compare/new-shopify-app-gem

@warrendunlop we briefly discussed once but you are cool with the deleting of all the design files from this gem? It looks like a lot of work so I would suggest we make another repo or gist of it. It might still be valuable to some user and we could link to it from this repo.

@stephenminded re: when you tried this a while back on Heroku - did you set the gem path for shopify_app to use the git branch gem in the generated app? I'll try a generated app on heroku to make sure it works before shipping. 

for review: @csaunders @Hammadk @garymardell


ckhatton:
:open_mouth: Wow @kevinhughes27 this is cool! Great work +1

I know you directed the question to Warren, but I would like to pipe up here, as I have been meaning to sort out the design files (for ages), and it is something that is constantly discussed (mostly as the bootstrap installation doesn't work / conflicts). It has been discussed to create a bootstrap-esque style that matches the Shopify admin, and now that this defaults to an embedded app, it makes more sense to do so. It could match the one from Timber?

Would it be something you (@warrendunlop included) would like me to sort out, when this PR is accepted? I can chuck the old style into a gist - I cannot see that it would be valuable as the html classes will possibly change :octocat:

kevinhughes27:
yeah that sounds good! I think the best solution would be to add a new generator that gives you some basic css/design assets etc if you want it. I really want people to start contributing all they can as generators here to avoid a lot of duplicated effort when starting new apps and this is a great example of just that.

Also down the line I'd like to make an internal gem with generators for things specific to internal apps - like setting up our logging and datadog etc properly. These are things that we shouldn't need to do over and over again

lib/generators/shopify_app/install/templates/shopify_app_ready_script.html
```diff
@@ -0,0 +1,8 @@
+<script type="text/javascript">
+  ShopifyApp.ready(function(){
+    ShopifyApp.Bar.initialize({
+      title: "Home",
+      icon: "<%= asset_path('faveicon.png') %>"
```
garymardell: `favicon` ?

kevinhughes27: this probably shouldn't be here - we used to copy over some assets including this png with the old generator. Options - add this png to the new generator or remove this line of code.

nice catch!

lib/shopify_app/login_protection.rb
```diff
@@ -1,44 +1,46 @@
...
-      ensure
-        ShopifyAPI::Base.clear_session
+    def shopify_session
+      if shop_session
+        begin
+          ShopifyAPI::Base.activate_session(shop_session)
+          yield
+        ensure
+          ShopifyAPI::Base.clear_session
+        end
```
stephenminded: Instead of activating and clearing, you might want to consider using `ShopifyAPI::Session.temp`.

kevinhughes27: this is a good point - this is how it runs now but yeah it really should use `ShopifyAPI::Session.temp`

warrendunlop:
It was a bunch of work but that was like 4 years ago so I have no problem deleting it. So long as a designer has taken a look at it. It should follow our current design guide and feel like a Shopify app out of the gate.

ckhatton:
@warrendunlop I'll be that designer :smiley:

kevinhughes27:
looping in @atleeclark @beckwith @joshubrown -relevant for app developers

Hammadk:
There is fairly tough to review, but everything looks good so far :+1: 

Turning this into an engine is a great idea.

kevinhughes27:
Yeah I agree an in depth review here is not super feasible. As long as a few people take a look through and everyone is in agreement then I think we should be okay to ship. It will be a new major version so no one should get this accidentally and even if they do the gem won't do anything unless they include the new concerns so its pretty safe imo

kevinhughes27:
@csaunders can you take a look through here (not expecting a detailed review). I am going to test a new app on Heroku when I find time and then :shipit: unless there are any objections

garymardell:
If you could convert https://github.com/Shopify/embedded-app-example which is our public example as the test app that would be awesome.

kevinhughes27:
@garymardell great idea

kevinhughes27:
@garymardell  https://github.com/Shopify/embedded-app-example/tree/update-shopify_app

I also tested on Heroku to make sure it works there. I originally ran into the same issue as @stephenminded and turns out it was related to the InMemorySessionStore. I knew this wouldn't *work* but thought it would work enough for me to test that the app runs on heroku. Turns out this fails in a super weird way because heroku just always returns nothing from memory and the app tries to use it as a session. I may strengthen the warning against the InMemorySessionStore for anything other than playing around locally.

last chance to look though, I am shipping tomorrow!

gavinballard:
Just wanted to say thanks to @kevinhughes27 - bundling this into an Engine has made Rails development much nicer. Much appreciated, five stars, would buy again :+1:.

kevinhughes27:
thanks! @gavinballard

PR #125, opened 2015-04-30
add configurable myshopify_domain

kevinhughes27:
This PR adds a configurable myshopify_domain to the ShopifyApp gem for internal use (and I suppose external use if an app dev has made a mock local shopify server).

@shawnfrench @stephenminded


shawnfrench:
Nice :+1:

stephenminded:
:cool:

PR #126, opened 2015-04-30
add utils module

kevinhughes27:
in Qbo we use the sanitize function outside of the SessionsController so in order to make the best use of this new gem configuration I am moving this function into a utils module which can more easily be used in other places.

@stephenminded @shawnfrench


stephenminded:
Sweet :shipit:

shawnfrench:
Very nice :+1:

PR #127, opened 2015-05-07
Remove hard requirement on shopify_api ~>3.2.0

stephenminded:
And bump gem to `6.0.5`

@kevinhughes27 @shawnfrench

### DIFF
--- a/shopify_app.gemspec
+++ b/shopify_app.gemspec
@@ -10,7 +10,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency('rails', '>= 3.1', '< 5.0')
 
-  s.add_runtime_dependency('shopify_api', '~> 3.2.0')
+  s.add_runtime_dependency('shopify_api', '>= 3.2.0')
   s.add_runtime_dependency('omniauth-shopify-oauth2', '~> 1.1.4')
 
   s.add_development_dependency('rake')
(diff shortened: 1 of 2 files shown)


shopify_app.gemspec
```diff
@@ -10,7 +10,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency('rails', '>= 3.1', '< 5.0')
 
-  s.add_runtime_dependency('shopify_api', '~> 3.2.0')
+  s.add_runtime_dependency('shopify_api', '> 3.2.0')
   s.add_runtime_dependency('omniauth-shopify-oauth2', '~> 1.1.4')
```
kevinhughes27: can we also remove the hard req here

shopify_app.gemspec
```diff
@@ -10,7 +10,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency('rails', '>= 3.1', '< 5.0')
 
-  s.add_runtime_dependency('shopify_api', '~> 3.2.0')
+  s.add_runtime_dependency('shopify_api', '> 3.2.0')
```
shawnfrench: Should this be `>=`?

stephenminded: Yeah, strictly speaking that's probably the best way to specify it. I'll change it

stephenminded:
OK, updated... another set of :eyes: ?

PR #132, opened 2015-05-13
passes the options through to the install generator

### LINKED ISSUE #131: Add cmd args to the `all` generator
The install generator takes arguments and while this is the main one lots of users will want to run every generator and currently this means they can't use CMD args to set their defaults before running the generator.

Would be a nice feature to have.

kevinhughes27:
closes #131 

This passes the arguments down into the install generator which uses them. Most users of this gem will probably use the simple `generate shopify_app` which does everything so it would be nice if cmd args worked here.

I wish passing args in between generators was cleaner but this works at least

for review:
@celsodantas @stephenminded

### DIFF
--- a/lib/generators/shopify_app/shopify_app_generator.rb
+++ b/lib/generators/shopify_app/shopify_app_generator.rb
@@ -2,8 +2,13 @@ module ShopifyApp
   module Generators
     class ShopifyAppGenerator < Rails::Generators::Base
 
+      def initialize(args, *options)
+        @opts = options.first
+        super(args, *options)
+      end
+
       def run_all_generators
-        generate "shopify_app:install"
+        generate "shopify_app:install #{@opts.join(' ')}"
         generate "shopify_app:shop_model"
 
         generate "shopify_app:controllers"


stephenminded:
Can you give an example of what args might be passed in?

kevinhughes27:
the args are -api_key=sghfdhsdh... -secret=dsfhsdfhdsfh... -embedded=true

lib/generators/shopify_app/shopify_app_generator.rb
```diff
@@ -2,8 +2,13 @@ module ShopifyApp
   module Generators
     class ShopifyAppGenerator < Rails::Generators::Base
 
+      def initialize(args, *options)
+        @opts = options.first
```
stephenminded: Did you consider using `extract_options!`? https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/array/extract_options.rb#L22

kevinhughes27:
they are used to create the shopify_app initializer for you

lib/generators/shopify_app/shopify_app_generator.rb
kevinhughes27: I didn't know about that but I am not sure it helps me here. I am just grabbing what I need here and sending downwards - here is where it actually gets used https://github.com/Shopify/shopify_app/blob/master/lib/generators/shopify_app/install/install_generator.rb#L12

stephenminded:
:sunglasses:

lib/generators/shopify_app/shopify_app_generator.rb
```diff
@@ -2,8 +2,13 @@ module ShopifyApp
...
+        super(args, *options)
```
celsodantas: you can just do `super`. The args are implicit.

celsodantas:
:+1:

lib/generators/shopify_app/shopify_app_generator.rb
kevinhughes27: yeah - the code is kind of weird looking though so I want to leave this anyways.

PR #133, opened 2015-05-13
draw routes on the app - no need to mount use the routes boolean

### LINKED ISSUE #129: Rails Engine Routes
Currently the default routes as part of the engine are drawn on the engine itself. This means they have to be mounted in order to become part of the hose app. They are also behind a config var called `routes` on ShopifyApp configuration.

The routes should actually be drawn on Rails.application and then the `routes` config will be the only thing controlling the routes in the host app. This is cleaner and makes the engine more transparent (some engines make sense to be mounted some make more sense to be more seamless and use config). This change will also help make the tests cleaner.

btw @stephenminded this answers the engine mounting confusion we had when upgrading everyone. I want to wait until the next version after the security bump to fix this since this adds an extra layer of safety before this app will add routes.

kevinhughes27:
closes #129 

Currently the routes are drawn on the engine and guarded by a config setting. The idea behind the config setting is that if you want the routes they appear naturally and not mounted at some location. For some rails engine it makes sense to mount the whole thing somewhere but this is more of an offloading to the engine type situation where it should be seamless. It was my mistake when doing the upgrade and still learning about engines.

However since everyone had to mass upgrade recently the extra work required to get the routes was beneficial and safer. I still think that actually using the routes provided by the gem will be an atypical use case as most will run the global generator which disables and copies over the routes (and most people will want their own session controller with some extra things).

I'll version bump after this one to keep things super safe for 6.0.5.

for review @celsodantas @stephenminded

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,5 @@
 if ShopifyApp.configuration.routes_enabled?
-  ShopifyApp::Engine.routes.draw do
+  Rails.application.routes.draw do
     controller :sessions do
       get 'login' => :new, :as => :login
       post 'login' => :create, :as => :authenticate
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,10 +1,8 @@
-if ShopifyApp.configuration.routes_enabled?
-  Rails.application.routes.draw do
-    controller :sessions do
-      get 'login' => :new, :as => :login
-      post 'login' => :create, :as => :authenticate
-      get 'auth/shopify/callback' => :callback
-      get 'logout' => :destroy, :as => :logout
-    end
+ShopifyApp::Engine.routes.draw do
+  controller :sessions do
+    get 'login' => :new, :as => :login
+    post 'login' => :create, :as => :authenticate
+    get 'auth/shopify/callback' => :callback
+    get 'logout' => :destroy, :as => :logout
   end
 end
--- a/lib/shopify_app/login_protection.rb
+++ b/lib/shopify_app/login_protection.rb
@@ -42,5 +42,11 @@ module ShopifyApp
       redirect_to login_path
     end
 
+    def login_path(params = {})
+      main_app.login_path(params)
+    rescue NoMethodError => e
+      shopify_app.login_path(params)
+    end
+
   end
 end
--- a/lib/generators/shopify_app/install/install_generator.rb
+++ b/lib/generators/shopify_app/install/install_generator.rb
@@ -70,6 +70,7 @@ module ShopifyApp
       end
 
       def add_home_index_route
+        route "mount ShopifyApp::Engine, at: '/'"
         route "root :to => 'home#index'"
       end
 
--- a/lib/generators/shopify_app/install/templates/home_controller.rb
+++ b/lib/generators/shopify_app/install/templates/home_controller.rb
@@ -1,4 +1,5 @@
 class HomeController < ApplicationController
+  before_action :login_again_if_different_shop
   around_filter :shopify_session
 <% if embedded_app? -%>
   layout 'embedded_app'
--- a/lib/generators/shopify_app/install/templates/shopify_app.rb
+++ b/lib/generators/shopify_app/install/templates/shopify_app.rb
@@ -3,5 +3,4 @@ ShopifyApp.configure do |config|
   config.secret = "<%= opts[:secret] || '<secret>' %>"
   config.scope = "<%= opts[:scope] || 'read_orders, read_products' %>"
   config.embedded_app = <%= opts[:embedded_app] || true %>
-  config.routes = true
 end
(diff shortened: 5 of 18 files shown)


stephenminded:
:sunglasses:

config/routes.rb
```diff
@@ -1,5 +1,5 @@
 if ShopifyApp.configuration.routes_enabled?
-  ShopifyApp::Engine.routes.draw do
+  Rails.application.routes.draw do
```
celsodantas: so this will automatically generate the routes for any app that adds `shopify_app` gem?

kevinhughes27: yeah the default is to have the routes which will forward to the basic controller that is part of the app. Any user of the gem will need an initializer and they can set routes false there. 

Most users will probably run the full generator which creates the controllers and copies the routes then sets the routes config to false.

I am going to do a 6.1.0 release after this so no one upgrade accidentally. (aside there is still small risk here because if the routes will be overriden by any routes in the host app. All that can happen is that a request that should have 404ed 500s)

celsodantas: I kinda prefer to be explicit then just magically have the session routes working once you install the gem.
If we want to be minimalistic maybe have something similar to Devise and have a method that can be called in the `routes.rb` to define the routes?

stephenminded: I kinda like that, or make the default config for `routes_enabled` false so they have to opt-in?

kevinhughes27: Okay I'll make the default in the config to false but then set it in the
initializer so it's a visible setting - how does that sound?

celsodantas: What do you mean about set it in the initializer?

kevinhughes27:
Okay I think I found the optimal solution here:

configuration.routes is no longer initialized by the configuration class itself. This means that simply adding the gem will have no affect. (config.routes will be false)

Running the install generator will create a default 'shopify_app' initializer which will include the line `config.routes = true` which will enable the routes.

Running the `routes` or global generator (which includes the routes generator) will copy the routes into the host app (overriding the engine) and also set the `config.routes` to false.

stephenminded:
You modelled the recent changes after Clearance, right? You've probably seen this, but this is their recommendation for routes (https://github.com/thoughtbot/clearance#routes):

> See config/routes.rb for the default set of routes.
>
> As of Clearance 1.5 it is recommended that you disable Clearance routes and take full control over routing and URL design.
>
> To disable the routes, set config.routes = false. You can optionally run rails generate clearance:routes to dump a copy of the default routes into your application for modification.

Sounds like you're following a similar strategy

kevinhughes27:
yeah I talked to @celsodantas a bit this morning and I am working to remove the `routes` config and just mount the engine routes

kevinhughes27:
Okay I go this working with mounting the Rails engine now. Heres the lo down:

1. The engine needs to be mounted explicitly in the routes file for the routes to affect the host app. Thus simply adding the gem does nothing.

2. The install generator adds the `mount` line to the host app routes file

3. Routes generator (and therefore the global generator) copies the routes into the host app routes file and removes the mount line

4. Testing mountable engines is weird - here is what I came up with which I think is the best. The dummy app has the routes explicitly copied (not mounted) this is because Rails testing can't handle the mounted routes and the host app routes at the same time and things get weird. However this leaves us with no tests testing that the engine mounts as expected - this is where my routes tests come in. In the routes test file I set the routes to the engine routes and ensure they route to the right spot. Maybe this will get better in a future version of rails but there is no shortage of people with this problem on the interwb (http://stackoverflow.com/questions/26820691/rails-4-engine-controller-redirect-fails-unit-testing,  http://stackoverflow.com/questions/7691594/how-to-test-routes-in-a-rails-3-1-mountable-engine, http://stackoverflow.com/questions/21294260/rails-engine-4-0-2-routing-methods-fail-in-controllers-using-minitest, http://stackoverflow.com/questions/22924752/no-route-matches-rails-engine)

another look please!

celsodantas:
:+1: from me. This looks much better imo =)

@csaunders mind taking a look too?

kevinhughes27:
thanks! I agree its deffs better now

kevinhughes27:
ping @csaunders

kevinhughes27:
I appended a small fix for an issue @clayton-shopify found when only running the `install` generator. This lets `login_path` work regardless of who owns the route.

lib/shopify_app/login_protection.rb
```diff
@@ -42,5 +42,11 @@ def close_session
       redirect_to login_path
     end
 
+    def login_path(params = {})
+      main_app.login_path(params)
+    rescue NoMethodError => e
+      shopify_app.login_path(params)
+    end
```
kevinhughes27: for some reason using `.try` here didn't work. It would work the first time but if you tried to login in to a different shop like described in #134 it would raise `NoMethodError` again. I opted for this fix which always works.

kevinhughes27:
@csaunders is probably busy with mobile buy sdk launch. @stephenminded can you take another look through here plz

csaunders:
Nah I'm fine now. Had to deal with some stuff yesterday. Getting on this now

csaunders:
:+1:

kevinhughes27:
thanks!

PR #135, opened 2015-05-19
Fix embedded flag

kevinhughes27:
This PR moves all the defaults into one place so it is super clear. Also fixes a bug with how the embedded flag was set.

for review:
@clayton-shopify @christhomson


christhomson:
:ship:

clayton-shopify:
Seems to be working properly with these changes. :+1:

PR #138, opened 2015-06-01
add shopify domain to session

kevinhughes27:
This PR adds the shopify shop name / domain to the session so its easier to see which shop the session is for. The main use for this is with exception catching tools that include the session with the exception. Currently all thats in the session is the id of the shop in the apps database which means you have to lookup which shop it actually is. With this addition you can see this right away.

for review
@shawnfrench @stephenminded

### DIFF
--- a/lib/shopify_app/sessions_controller.rb
+++ b/lib/shopify_app/sessions_controller.rb
@@ -12,8 +12,10 @@ module ShopifyApp
 
     def callback
       if response = request.env['omniauth.auth']
-        sess = ShopifyAPI::Session.new(params[:shop], response['credentials']['token'])
+        shop_name = response.uid
+        sess = ShopifyAPI::Session.new(shop_name, response['credentials']['token'])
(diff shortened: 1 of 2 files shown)


lib/shopify_app/sessions_controller.rb
```diff
@@ -12,8 +12,10 @@ def create
 
     def callback
       if response = request.env['omniauth.auth']
-        sess = ShopifyAPI::Session.new(params[:shop], response['credentials']['token'])
+        shop_name = params[:shop]
```
stephenminded: Pretty sure there was a security issue taking the shop name from the params instead of from the omniauth hash. Ping @Shopify/security @dylanahsmith  (for reference: https://github.com/Shopify/shopify-theme-store/pull/494)

kevinhughes27: I was wondering about this actually

dylanahsmith:
:+1:

stephenminded:
:shipit:

PR #139, opened 2015-06-02
Bump omniauth-shopify-oauth2 gem and use the default setup lambda.

dylanahsmith:
@stephenminded & @kevinhughes27 for review

Require the latest version of omniauth-shopify-oauth2 gem and use the default setup lambda so that it doesn't need to be generated for each app.


stephenminded:
:+1:

kevinhughes27:
:+1:

PR #145, opened 2015-06-23
Authenticated controller

kevinhughes27:
I've noticed that its a pretty common error to refactor code into the ApplicationController and not realize that it affects the sessions controller. A pattern that I use a lot is to create an AuthenticatedController which has the required before filters for the authenticated part of the app. This helps prevent accidentally breaking auth for your app while refactoring into the ApplicationController.

I also pulled the layout config into AuthenticatedController and out of the generated HomeController. This is an improvement since now if you inherit AuthenticatedController everything just works

It seems a bit weird to me that then ApplicationController still gets ShopifyApp::Controller included since only its derived classes use these methods. The alternative would be to not include ShopifyApp::Controller in the ApplicationController and instead include it in both the SessionsController and the AuthenticatedController - thoughts?

@qq99 @AWaselnuk @stephenminded

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,12 @@ ShopifyApp::SessionRepository.storage = 'Shop'
 If you run the `shop_model` generator it will create the required code to use the generated Shop model as the SessionRepository and update the initializer.
 
 
+AuthenticatedController
+-----------------------
+
+The engine includes a controller called `AuthenticatedController` which inherits from `ApplicationController`. It adds some before_filters which ensure the user is authenticated and will redirect to the login page if not. It is best practice to have all controllers that belong to the Shopify part of your app inherit from this controller. The HomeController that is generated already inherits from AuthenticatedController.
+
+
 Questions or problems?
 ----------------------
 http://api.shopify.com <= Read up on the possible API calls!
--- a/lib/generators/shopify_app/install/templates/shopify_provider.rb
+++ b/lib/generators/shopify_app/install/templates/shopify_provider.rb
@@ -2,4 +2,4 @@
     ShopifyApp.configuration.api_key,
     ShopifyApp.configuration.secret,
 
-    :scope => ShopifyApp.configuration.scope,
+    :scope => ShopifyApp.configuration.scope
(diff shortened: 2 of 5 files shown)


qq99:
cc @nt3rp

qq99:
:+1: just used this in our project, works very well

AWaselnuk:
I think `ShopifyApp::Controller` is fine included in `ApplicationController`. That is where I would naturally go to look for that. Good refactor here :)

lib/generators/shopify_app/install/templates/shopify_provider.rb
```diff
@@ -2,4 +2,4 @@
     ShopifyApp.configuration.api_key,
     ShopifyApp.configuration.secret,
 
-    :scope => ShopifyApp.configuration.scope,
+    :scope => ShopifyApp.configuration.scope
```
kevinhughes27: this is a bug fix btw. the generator is broken with this atm

stephenminded:
:rocket: 

Maybe consider a `README.md` update to document that people should inherit from the generated `AuthenticatedController` if they want the batteries included?

kevinhughes27:
great point

kevinhughes27:
added to the readme. Can someone give my words a look please

README.md
```diff
@@ -143,6 +143,12 @@ ShopifyApp::SessionRepository.storage = 'Shop'
 If you run the `shop_model` generator it will create the required code to use the generated Shop model as the SessionRepository and update the initializer.
 
 
+AuthenticatedController
+-----------------------
+
+The engine comes with a controller called AuthenticatedController which inherits from ApplicationController and adds some before_filters which ensure the user is authenticated or it redirects to the login page. Its best practice to have all controllers that belong to the Shopify part of your app inherit from this controller. The HomeController that is generated already inherits from AuthenticatedController.
```
stephenminded: Maybe split the first sentence up into 2, it's a little long. 

"The engine includes a controller called `AuthenticatedController` which inherits from `ApplicationController`. It adds some before_filters which ensure the user is authenticated and will redirect to the login page if not."

Also, "Its" should be "It's" or "It is"

stephenminded:
Wording suggestions, otherwise :rocket:

kevinhughes27:
updated!

PR #148, opened 2015-06-30
Fix session glitch

### LINKED ISSUE #134: How to use filter :login_again_if_different_shop
Hi,

I want to use __login_again_if_different_shop__ in my app. 
```ruby
  if shop_session && params[:shop] && params[:shop].is_a?(String) && shop_session.url != params[:shop]
    redirect_to_login
  end
end
```
I've added it like following:
```ruby
class ApplicationController < ActionController::Base
  before_filter :login_again_if_different_shop
```
```ruby
class SessionsController < ApplicationController
  skip_before_filter :login_again_if_different_shop
```
for now, I opened my app in Shop A, then open a new window(Chrome tab) to access my app in Shop B. 
the code always capture the url of Shop A from params[:shop], so redirect to Shop A.

If I remove __skip_before_filter__ from SessionsController then get the __Too many redirects__ error.
How to make it switch to the Shop B automatically?

qq99:
Fixes #134 (I'm 99% sure)

Fixes a session glitch where another return_to URL would be stored from a previous request by the same browser

cc @kevinhughes27

### DIFF
--- a/lib/shopify_app/sessions_controller.rb
+++ b/lib/shopify_app/sessions_controller.rb
@@ -42,7 +42,7 @@ module ShopifyApp
     end
 
     def return_address
-      session[:return_to] || main_app.root_url
+      session.delete(:return_to) || main_app.root_url
     end
 
     def sanitized_shop_name


lib/shopify_app/sessions_controller.rb
```diff
@@ -16,6 +16,7 @@ def callback
         sess = ShopifyAPI::Session.new(shop_name, response['credentials']['token'])
         session[:shopify] = ShopifyApp::SessionRepository.store(sess)
         session[:shopify_domain] = shop_name
+        session[:return_to] = nil
```
kevinhughes27: err now `redirect_to return_address` won't work at all. We need to pop this out of the session so we can use it and delete it

qq99: won't it fall back to `main_app.root_url`?

kevinhughes27: yeah but if return_to was actually set now we have unset and not used it.

I think the solution here is just to clean this up after we've used it is all.

qq99:
```ruby
    def return_address
    end
```
like such?

qq99: I'm not really familiar with OAuth

qq99:
Tweaked @kevinhughes27 how is this?

lib/shopify_app/sessions_controller.rb
kevinhughes27: Its complicated stuff because there are so many possible states. I don't think what you've got there is correct because we only want to clear it in the callback endpoint. We want it to carry through the other ones

kevinhughes27:
:+1: ignore my other comment this works since we only call `return_address` when we are done an oauth flow (or exit early)

qq99:
cool I also moved to a `.delete` and didn't `nil` it out anymore

PR #150, opened 2015-07-20
Close session in login_again_if_different_shop

shawnfrench:
@kevinhughes27 @stephenminded I think this fixes the issue I found with `login_again_if_different_shop` not actually redirecting to the new shop.

There might be a better way to test this. Thoughts?

### DIFF
--- a/lib/shopify_app/login_protection.rb
+++ b/lib/shopify_app/login_protection.rb
@@ -27,6 +27,7 @@ module ShopifyApp
     def login_again_if_different_shop
       if shop_session && params[:shop] && params[:shop].is_a?(String) && shop_session.url != params[:shop]
         session[:shopify] = nil
+        session[:shopify_domain] = nil
         redirect_to_login
       end
     end
@@ -40,6 +41,7 @@ module ShopifyApp
 
     def close_session
       session[:shopify] = nil
+      session[:shopify_domain] = nil
       redirect_to login_path
     end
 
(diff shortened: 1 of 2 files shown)


kevinhughes27:
lgtm

lib/shopify_app/login_protection.rb
```diff
@@ -26,6 +26,7 @@ def shop_session
 
     def login_again_if_different_shop
       if shop_session && params[:shop] && params[:shop].is_a?(String) && shop_session.url != params[:shop]
+        session[:shopify] = nil
```
stephenminded: We also set `:shopify_domain` in the session here: https://github.com/Shopify/shopify_app/blob/dd7feffd03ab047cfbf4fad4187f11ca2aa5cec3/lib/shopify_app/sessions_controller.rb#L18 

Might be worth doing a `reset_session` instead?

shawnfrench: @kevinhughes27 you added all the shopify_domain stuff - what was the reason?

kevinhughes27: I added it so that when error handlers (like errbit) dumped the session we had the shop domain not just the id. We had an issue you made for this

shawnfrench: lol well I guess I should be careful what I ask for! 
It's not so terrible, but I didn't understand since in code this can be retrieved. Understand now it's mostly for when session gets dumped blindly by something like errbit :+1:

kevinhughes27: yeah its also irrelevant now with bugsnag. If we leave it in we should add a comment

shawnfrench:
Forgot about this - you guys good to :ship: ?

kevinhughes27:
:rocket:

stephenminded:
:money_with_wings:

PR #152, opened 2015-08-05
add redirect uri to shopify app

kevinhughes27:
since https://github.com/Shopify/shopify/pull/46095 redirect_uri is a required param for apps. This PR adds redirect_uri to the ShopifyApp config and the omniauth provider. It can be passed into the generators the same as api_key and other params.

@ShayneP and I top hatted a new app locally and were successful.

ToDo
- [ ] Update appinfive screencast somehow - what are our options here? @joshubrown. Seems like lots of people are using it and then running into issues since it is now out of date
- [ ] Update shopify_api and shopify_python_api READMEs to indicate that this param is no longer optional. @ShayneP can you take this please

for review @EiNSTeiN- @awd

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ $ rails generate shopify_app:install
 
 # or optionally with arguments:
 
-$ rails generate shopify_app:install -api_key=<your_api_key> -secret=<your_app_secret>
+$ rails generate shopify_app:install -api_key=<your_api_key> -secret=<your_app_secret> -redirect_uri=<your_redirect_uri>
 ```
 
 Other options include:
@@ -101,6 +101,7 @@ The `install` generator places your Api credentials directly into the shopify_ap
 ShopifyApp.configure do |config|
   config.api_key = ENV['SHOPIFY_CLIENT_API_KEY']
   config.secret = ENV['SHOPIFY_CLIENT_API_SECRET']
+  config.redirect_uri = "<%= your_redirect_uri %>"
   config.scope = 'read_customers, read_orders, write_products'
   config.embedded_app = true
 end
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -101,6 +101,7 @@ The `install` generator places your Api credentials directly into the shopify_ap
 ShopifyApp.configure do |config|
   config.api_key = ENV['SHOPIFY_CLIENT_API_KEY']
   config.secret = ENV['SHOPIFY_CLIENT_API_SECRET']
+	config.redirect_uri = "<%= your_redirect_uri %>"
```
EiNSTeiN-: indent is weird

EiNSTeiN-:
The change looks good to me. Do you know why all of our own apps already send a redirect_url to the oauth authorize phase if the redirect_url was not set by default here?

kevinhughes27:
I don't - is that we see on splunk? strange

kevinhughes27:
ping @awd

awd:
This LGTM, any way to try and merge this and the other repos around the same time to prevent weird fragmentation of documentation and functionality?

kevinhughes27:
The other repos work fine they just need the readme updated to say that redirect_uri is required not optional. @ShayneP was gonna take this

awd:
okay cool, :shipit: :+1:

kevinhughes27:
I don't think many people use the api libs to do oauth manually anyways (why would you) and this is broken right now so we should ship a working version

joshubrown:
@kevinhughes27 Yep I guess we need to update the video unfortunately. The first one you did was amazing – would you like to record it again?

kevinhughes27:
I recorded it in small segments. I might be able to just add one

kevinhughes27:
@joshubrown did you figure out how to do annotations? why don't we just add an annotation saying now you need to set redirect uri when creating your app on the partners page and then mention it again when running the generator

PR #153, opened 2015-08-06
Return an HTTP 401 for XHRs that aren't logged in

nickhoffman:
## Problem

`session[:return_to]` shouldn't be set if the request is an XHR, because browsing sessions rarely mean to be able to return to XHR paths.

## Solution

Don't set it if the request is an XHR.

/related Shopify/facebook-commerce#3525
/review @kevinhughes27, @peterjm

### DIFF
--- a/test/shopify_app/login_protection_test.rb
+++ b/test/shopify_app/login_protection_test.rb
@@ -6,6 +6,7 @@ class LoginProtectionController < ActionController::Base
   include ShopifyApp::LoginProtection
   helper_method :shop_session
 
+  around_action :shopify_session, only: [:index]
   before_action :login_again_if_different_shop, only: [:second_login]
 
   def index
@@ -61,6 +62,20 @@ class LoginProtectionTest < ActionController::TestCase
       assert_redirected_to @controller.send(:login_path, shop: 'other_shop')
       assert_nil session[:shopify]
       assert_nil session[:shopify_domain]
+    end
+  end
+
+  test '#shopify_session with no Shopify session, redirects to the login path' do
+    with_application_test_routes do
+      get :index, shop: 'foobar'
+      assert_redirected_to @controller.send(:login_path, shop: 'foobar')
+    end
+  end
+
+  test '#shopify_session with no Shopify session, sets session[:return_to]' do
+    with_application_test_routes do
+      get :index, shop: 'foobar'
+      assert_equal '/?shop=foobar', session[:return_to]
     end
   end
 
--- a/test/shopify_app/login_protection_test.rb
+++ b/test/shopify_app/login_protection_test.rb
@@ -79,6 +79,13 @@ class LoginProtectionTest < ActionController::TestCase
     end
   end
 
+  test '#shopify_session with no Shopify session, when the request is an XHR, returns an HTTP 401' do
+    with_application_test_routes do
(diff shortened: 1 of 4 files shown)


nickhoffman:
Note: There're no tests for setting `session[:return_to]` . I tried to add them, but it looks like it'll take more time than it's worth.

kevinhughes27:
yeah thats the same issue we ran into last time there was an issue with this (tests for that are hard to write)

kevinhughes27:
:+1:

nickhoffman:
AFAICT, this will break the return-to functionality of single-page apps that rely heavily on Ajax and manipulating the browser's `history` object.

peterjm:
This'll solve the cookie issue, but if the request is XHR, the redirect will be processed by javascript and the merchant experience will still be broken, won't it?

nickhoffman:
@peterjm, I'm afraid I don't follow...

peterjm:
We're redirecting a javascript request to the login path; that won't get translated to the merchant's browser being redirected to the login path. From the user experience perspective, it'll seem like the Javascript request just silently failed.

nickhoffman:
@peterjm, yes, that's true. But that's already the case and not what this PR is addressing. This PR prevents `session[:return_to]` from being set to a path that was requested via an XHR.

peterjm:
:+1: if you really want to ship this as is; it does fix a potential crash.

But it seems like we shouldn't fix something, only to have it fail on the very next line. Wouldn't something like this be more generally helpful:

```
def redirect_to_login
  if request.xhr?
    head :unauthorized
  else
    ...
  end
end
```

nickhoffman:
Yup, that'd be much better. However, I don't know if that would break any apps. Some apps might rely on the existing bad behaviour...

kevinhughes27:
change log it. apps don't have to upgrade

nickhoffman:
Updated with @peterjm's suggestion , and bumped the version. Could you guys give this another review, please?

kevinhughes27:
looks fine to me. A test would be nice ..

peterjm:
:+1: with test

nickhoffman:
Sorry for the delay, guys. Tests added. Another quick :eyes, please, @peterjm and @kevinhughes27?

test/shopify_app/login_protection_test.rb
```diff
@@ -7,6 +7,7 @@ class LoginProtectionController < ActionController::Base
   helper_method :shop_session
 
   before_action :login_again_if_different_shop, only: [:second_login]
+  around_action :shopify_session, only: [:with_a_shopify_session]
```
kevinhughes27: does index not require a shopify_session? can we do this without another route

nickhoffman: Neither of the existing controller actions calls `#shopify_session` . I might be able to restructure the test controller and the tests, but that's a lot more work than I'd like to put into this.

kevinhughes27: interesting. Can we re-use index or will that break tests?. Can we rename the route `internal` or something to indicate that it is behind auth and not related to it

nickhoffman: Good suggestion! I was able to throw `#shopify_session` around the `index` action.

kevinhughes27:
:+1:

Issue #154, opened 2015-08-11
Running the app from a sub-directory


ramananbalakrishnan:
Hi, 
I am trying to run my (non-embedded) app from a sub-directory, "/myapp" instead of "/" . 

#### issue
After following the instructions on [Rails guides](http://guides.rubyonrails.org/configuring.html#deploy-to-a-subdirectory-relative-url-root) and also modifying the deployment configurations, I noticed that this relative_url_root is not being respected in [`SessionsController`](https://github.com/Shopify/shopify_app/blob/22f73e1dbb5d10b1c67549bd51725119ab3f2cfb/lib/shopify_app/sessions_controller.rb) when redirecting a user who enters a shop_name in `/login`, 
i.e. the person is sent to "/auth/shopify?shop=test.myshopify.com"
instead of "/myapp/auth/shopify?shop=test.myshopify.com" 

#### possible fix
I am not sure if this issue is related to some earlier issues, but changing the redirect destination in [`SessionsController#authenticate`](https://github.com/Shopify/shopify_app/blob/22f73e1dbb5d10b1c67549bd51725119ab3f2cfb/lib/shopify_app/sessions_controller.rb#L38) from `"/auth/shopify?shop=#{shop_name}"` to 
`"#{main_app.root_url}auth/shopify?shop=#{shop_name}"` seems to fix the specific problem that I had faced. 

Not sure if it would break other parts though.

kevinhughes27:
Nice find - yeah its tough to think through auth stuff. I think it will be fine, do you want to write up a PR and then we'll see if the test suite likes it.

In the meantime I would just overwrite that method in your app so you are not blocked by this.

ramananbalakrishnan:
After an initial misstep, looks like my PR now passes the tests and also solves my problem. You can merge when suitable.

I have corrected my local version to fix things in the meantime

PR #155, opened 2015-08-11
Use root_url from main_app when redirecting the user in authenticate

ramananbalakrishnan:
The `main_app.root_url` parameter is used to set the appropriate leading value for the path. This should fix issue #154 .

### DIFF
--- a/lib/shopify_app/sessions_controller.rb
+++ b/lib/shopify_app/sessions_controller.rb
@@ -35,7 +35,7 @@ module ShopifyApp
 
     def authenticate
       if shop_name = sanitize_shop_param(params)
-        fullpage_redirect_to "#{main_app.root_url}auth/shopify?shop=#{shop_name}"
+        fullpage_redirect_to "#{main_app.root_path}auth/shopify?shop=#{shop_name}"
       else
         redirect_to return_address
       end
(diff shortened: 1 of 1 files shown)


kevinhughes27:
Cool! Thanks for the PR - I just want to do a sanity check with some of our apps locally before merging.

kevinhughes27:
I tested locally and everything seemed to still work with one of our apps. @stephenminded can you take a look here as well

kevinhughes27:
actually @stephenminded is pretty busy, @christhomson for another review!

lib/shopify_app/sessions_controller.rb
```diff
@@ -35,7 +35,7 @@ def destroy
 
     def authenticate
       if shop_name = sanitize_shop_param(params)
-        fullpage_redirect_to "/auth/shopify?shop=#{shop_name}"
+        fullpage_redirect_to "#{main_app.root_path}auth/shopify?shop=#{shop_name}"
```
stephenminded: I guess it's safe to rely on `root_path` always being there? In a vanilla rails app it isn't, but the first thing you're told to do is set it in `config/routes.rb`.

stephenminded:
:+1: Thanks for the PR @ramananbalakrishnan

lib/shopify_app/sessions_controller.rb
kevinhughes27: I already rely on it else where in the gem

stephenminded: Awesome, then we're good

christhomson:
I'm a bit late to the party :tada:, but looks good to me!

PR #159, opened 2015-09-14
Update install instructions

andrewromanenco:
In case rails application was running before shopify_app generator execution, the generation process get stuck. Spring should be stopped (or confirm that is is not running).

cc: @kevinhughes27

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,9 @@ Generators
 ### Install Generator
 
 ```sh
+# stop Spring gem (or confirm it's not running)
+$ bundle exec spring stop
+
 $ rails generate shopify_app:install
 
 # or optionally with arguments:
--- a/README.md
+++ b/README.md
@@ -53,9 +53,6 @@ Generators
 ### Install Generator
 
 ```sh
-# stop Spring gem (or confirm it's not running)
-$ bundle exec spring stop
-
 $ rails generate shopify_app:install
 
 # or optionally with arguments:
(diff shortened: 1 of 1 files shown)


kevinhughes27:
I was envisioned a new section called Troubleshooting at the bottom but maybe it is better closer to where you need to run the command. Can we at least make this more of an after note though? like if this does nothing try stopping spring?

andrewromanenco:
Kevin, I just made a test, here is the only one way when I can reproduce this "feature"
Create rails app, run rails app, run shopify_app generator -> problem
if it goes like:
Create rails app, run shopify_app generator, run rails app -> no problem

I would recommend to have '''spring stop''' executed every time and let it be part of the instruction. The reason is that in case of bad luck, the process get stuck and ctrl+c is ignored. The process actually has to be killed by ````kill -9 pid``` command.

So I would either let stop command to be part of generator steps (like it is in this PR right now). Or add troubleshooting section at the end.

What do you think?

andrewromanenco:
Now I think that troubleshooting session is better one. Technically, if one follows instructions from readme, there is no need to use ```spring stop```. Let me update the doc

README.md
```diff
@@ -157,6 +157,17 @@ AuthenticatedController
 
 The engine includes a controller called `AuthenticatedController` which inherits from `ApplicationController`. It adds some before_filters which ensure the user is authenticated and will redirect to the login page if not. It is best practice to have all controllers that belong to the Shopify part of your app inherit from this controller. The HomeController that is generated already inherits from AuthenticatedController.
 
+Troubleshooting
+---------------
+
+### Generator shopify_app:install hangs
+
+Shopify_app uses spring gem to speed up the development. Spring must be stopped for generator to proceed. Kill 'shopify_app:install' process and stop Spring:
```
kevinhughes27: `Rails uses spring by default to speed up development` ... Sometimes Spring must be stopped`...

Would be how I would word it.

andrewromanenco: haha, diplomacy rules

kevinhughes27:
lgtm

Issue #165, opened 2015-10-12
config.redirect_uri setting ignored in authentication flow


lulessa:
1. In config/initializers/shopify_app.rb, set `config.redirect_uri` to any URL other than the standard https://myapp.example.com/auth/shopify/callback.
2. Submit a request to your app's /login route and watch your network logs' query string parameters.
3. The `redirect_uri` request param sent to the /admin/oauth/authorize endpoint is *always* equal to the /auth/shopify/callback route on the host and port where the app is running rather than the `config.redirect_uri` setting.

This is a corner case, for sure. Always submitting the /auth/shopify/callback route as the redirect_uri param may work just fine for your standard app setup, but when developing and using a different port for https vs http (#160), for example, it rears its ugly head.

The setting is there, but it has no effect. A search for `redirect_uri` in this repo shows ~~zero~~ evidence of the redirect_uri URL param being set during Shopify API's authentication flow. (Edit: It's in config/initializers/omniauth.rb)

It looks like `create_permission_url(scope, redirect_uri)` has to be called on the ShopifyAPI session at some point, as seen here:
https://github.com/Shopify/shopify_api/blob/4fb8f572136b9641157ecc177551e12744c27959/test/session_test.rb#L86

kevinhughes27:
Its used here: https://github.com/Shopify/shopify_app/blob/master/lib/generators/shopify_app/install/templates/shopify_provider.rb#L5

Can you confirm that it is present in your generated shopify oauth provider? if you updated the gem and didn't add this yourself then it would be missing (the generators don't re-run on update)

lulessa:
Yes, it is present in config/initializers/omniauth.rb. Here's the full file, unmodified since generation:
```ruby
Rails.application.config.middleware.use OmniAuth::Builder do
  provider :shopify,
    ShopifyApp.configuration.api_key,
    ShopifyApp.configuration.secret,

    :redirect_uri => ShopifyApp.configuration.redirect_uri,

    :scope => ShopifyApp.configuration.scope
end
```
I started a new Rails app and ran `rails generate shopify_app` using shopify_app 6.2.0

lulessa:
Take a look at https://github.com/Shopify/omniauth-shopify-oauth2/issues/29

They were getting the same message I was getting "Oauth error invalid_request: The redirect_uri is missing or invalid" — even though my `ShopifyApp.configuration.redirect_uri` matches what's set in the partner dashboard.

[...]

Issue #160, opened 2015-09-14
Redirects in the app mess with http/https


andrewromanenco:
cc @kevinhughes27 no sure if this is qualifies as an issue

Use case:
Run rails app on port 3000
Have stunnel running to map https port 8888 to http port 3000
Access application with url: https://localhost:8888
Result: application get redirected to http://localhost:8888/login (http->https)

Expected behaviour: if application was accessed via https, all redirects should let user to stay in https.

kevinhughes27:
maybe? your setup is complicated. Can you find exactly where the app is re-directed and changing from https to http?

andrewromanenco:
I'll get back to you, I am working on an app
tnx

PR #166, opened 2015-10-13
Add callback_url to shopify_provider.rb generator

lulessa:
Ensures that `ShopifyApp.configuration.redirect_uri` is actually used in oauth flow for apps using install generator.
Resolves issues #165 and #160


kevinhughes27:
lgtm! Thanks

PR #167, opened 2015-10-14
fixed typos in README file on line 71 and 80

lemuelbarango:
Added a comma on line 71: After running the `install` generator, you can start...
Also changed 'create' to 'creates' on line 81: It also creates a model called...


kevinhughes27:
thanks!

PR #173, opened 2015-11-02
move SessionStorage from a generated class to a re-usable module

kevinhughes27:
The generated `SessionStorage` class was an un-needed extra bit of mis-direction surrounding the `SessionRepository` object which allows you to tell ShopifyApp how to retrieve your session. It used to look like this:

SessionRepository -> SessionStorage -> Shop

SessionStorage just looked up and returned a shop. With this done as a mixin users can include in whichever model includes the `shopify_domain` and `shopify_token` or then can write their own `store` and `retrieve` on a class of their choosing just like before. The flow now looks like this:

SessionRepository -> Shop

And shop includes the SessionStorage module.

Another bonus of this cleanup is it results in less generated code and more code that is part of the gem meaning this gem feels less invasive to the host rails application.

for review
@christinejordan @christhomson


lib/shopify_app/session_storage.rb
```diff
@@ -0,0 +1,23 @@
...
+        shop.shopify_token = session.token
+        shop.save!
+        shop.id
+      end
+
+      def retrieve(id)
+        return unless id
+        shop = self.find(id)
+        ShopifyAPI::Session.new(shop.shopify_domain, shop.shopify_token)
+      rescue ActiveRecord::RecordNotFound
```
christhomson: I might prefer something like:

```ruby
if shop = self.find_by(id: id) # won't raise if not found
  ShopifyAPI::Session.new(shop.shopify_domain, shop.shopify_token)
end
```


… just so we don't need to `rescue` here. But up to you.

christhomson:
The README could be updated here: https://github.com/Shopify/shopify_app/blob/82344669ab23c7509ebc37188e2717afb5ac52c6/README.md#shopifyappsessionrepository. I do think it'd be useful to leave some mention of being able to swap out the session storage, but we could remove the bit about how to implement session storage for AR, and just mention that it'll use the Shop model as its default storage.

LGTM otherwise.

kevinhughes27:
Thanks @christhomson I totally spaced and forgot to update the readme. I'll have a look and make some cope changes

lib/shopify_app/session_storage.rb
kevinhughes27: good point, might as well make this change while I am in here

kevinhughes27:
updated the readme!

christhomson:
:rowboat:

kevinhughes27:
@christinejordan is sick so I am pinging @stephenminded for a review here

CHANGELOG
```diff
@@ -1,3 +1,8 @@
+Master
+------
+
+* Move SessionStorage from a generated class to a re-usable module
```
stephenminded: Is there a migration path from the generated class to using the module? Should we detail it here?

README.md
```diff
@@ -111,47 +111,9 @@ end
...
-
-Changing the `ShopifyApp::SessionRepository.storage` can simply be done by editing
-`config/initializers/shopify_session_repository.rb` to use the correct model.
-
-```ruby
-ShopifyApp::SessionRepository.storage = 'Shop'
-```
-
-If you run the `shop_model` generator it will create the required code to use the generated Shop model as the SessionRepository and update the initializer.
+`ShopifyApp::SessionRepository` allows you as a developer to define how your sessions are retrieved and stored for a shop. The `SessionRepository` is configured using the `config/initializers/shopify_session_repository.rb` file and can be set to any object the implements `self.store(shopify_session)` and `self.retrieve(id)`.
```
stephenminded: We should probably state that the contract for `store` is that it returns a unique identifier for the shop (all our examples return `id`) and also that `retrieve` is expected to return a `ShopifyAPI::Session` object for the shop representing the id passed in.

CHANGELOG
kevinhughes27: good point - the migration path is pretty simple delete the old SessionStorage class and include the mixin

kevinhughes27:
updated!

CHANGELOG
```diff
@@ -1,3 +1,10 @@
+Master
+------
+
+* Move SessionStorage from a generated class to a re-usable module. To
+  migrate to this new class delete the old generated SessionStorage class
+  under models and include the SessionStorage mixin in your Shop model
```
stephenminded: This sentence is a bit awkward and refers to the new module as both a class and a mixin, maybe something like:

 - "To migrate to the new module, delete the old generated SessionStorage class in the models directory and include the SessionStorage module in your Shop model."

README.md
```diff
@@ -111,47 +111,9 @@ end
...
-
-Changing the `ShopifyApp::SessionRepository.storage` can simply be done by editing
-`config/initializers/shopify_session_repository.rb` to use the correct model.
-
-```ruby
-ShopifyApp::SessionRepository.storage = 'Shop'
-```
-
-If you run the `shop_model` generator it will create the required code to use the generated Shop model as the SessionRepository and update the initializer.
+`ShopifyApp::SessionRepository` allows you as a developer to define how your sessions are retrieved and stored for a shop. The `SessionRepository` is configured using the `config/initializers/shopify_session_repository.rb` file and can be set to any object the implements `self.store(shopify_session)` which stores the session and returns a unique identifier and `self.retrieve(id)` which returns a `ShopifyAPI::Session` for the passed id. See either the `InMemorySessionStore` or the `SessionStorage` module for examples.
```
stephenminded: :+1:

stephenminded:
Another small tweak to the changelog, otherwise LGTM

PR #175, opened 2015-11-02
include the flash messages partial in the embedded app generator

kevinhughes27:
This PR adds the esdk flash messages template during app generation if the app is embedded. I also made the script smart enough to detect turoblinks and use the appropriate event (tophatted both locally). I don't think we should bother including anything flash related for non-embedded apps since there is a lot of different ways to do that and it would require some sort of default styling to actually work out of the box.

for review:
@christhomson @kwestein

### DIFF
--- a/lib/generators/shopify_app/install/templates/_flash_messages.html.erb
+++ b/lib/generators/shopify_app/install/templates/_flash_messages.html.erb
@@ -1,0 +1,13 @@
+<script type="text/javascript">
+  var eventName = typeof(Turbolinks) !== 'undefined' ? 'page:change' : 'DOMContentLoaded';
+
+  document.addEventListener(eventName, function() {
+    <% if flash[:notice] %>
+      ShopifyApp.flashNotice("<%= j flash[:notice].html_safe %>");
+    <% end %>
+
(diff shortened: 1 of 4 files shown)


lib/generators/shopify_app/install/templates/_flash_messages.html.erb
```diff
@@ -0,0 +1,13 @@
+<script type="text/javascript">
+  var eventName = typeof(Turbolinks) !== 'undefined' ? 'page:change' : 'DOMContentLoaded';
+
+  $(document).on(eventName, function() {
```
christhomson: Do we depend on jQuery anywhere else in this gem?

kevinhughes27: this is also a valid point. I should make this pure JS

kevinhughes27:
jquery :fire: !

christhomson:
:ship: :it:

kevinhughes27:
@kwestein can I get a review on this first thing tomorrow! I want to get this stuff out the door before Helsinki

kwestein:
lgtm

PR #176, opened 2015-11-02
create shop fixtures as part of shop model generator

kevinhughes27:
While creating a new app with @christinejordan we realized that the shop model generator doesn't add any fixtures or even create the fixtures file. This PR adds that and it means one less thing for the user of this gem to do when setting up a new app.

for review:
@christinejordan @christhomson

### DIFF
--- a/lib/generators/shopify_app/shop_model/templates/shops.yml
+++ b/lib/generators/shopify_app/shop_model/templates/shops.yml
@@ -1,0 +1,3 @@
+regular_shop:
+  shopify_domain: 'regular-shop.myshopify.com'
+  shopify_token: 'token'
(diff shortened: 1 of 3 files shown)


lib/generators/shopify_app/shop_model/templates/shops.yml
```diff
@@ -0,0 +1,3 @@
+regular_shop:
+  shopify_domain: 'regular_shop.myshopify.com'
```
christhomson: Maybe make this `regular-shop.myshopify.com` since subdomains can't include underscores.

kevinhughes27: :+1:

christhomson:
:+1:

PR #177, opened 2015-11-03
add a webhooks manager to the framework

kevinhughes27:
This PR brings in the WebhooksManager that has made its way into most of our apps that require webhooks. It allows developers to say which webhooks their app needs in the ShopifyApp initializer and then the framework takes care of the rest. In the callback method when the shop is being installed or logged in we queue the webhooks manager to make sure all the required hooks are in place. This does require ActiveJob to be configured for the app to work properly since we wouldn't want to block waiting to talk to shopify again after the callback.

* This is totally optional and won't have any affect on existing apps unless they opt in and use this class
* If an app is using a background queue directly they can still use this. for example Sidekiq can live happily next to ActiveJob.

for review:
@t6d @christhomson

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,23 @@ ShopifyApp.configure do |config|
 end
 ```
 
+
+WebhooksManager
+---------------
+
+ShopifyApp can manage your app's webhooks for you (requires ActiveJob). Set which webhooks you require in the initializer:
+
+```ruby
+ShopifyApp.configure do |config|
+  config.webhooks = [
+    {topic: 'carts/update', address: 'example-app.com/webhooks', format: 'json'}
+  ]
+end
+```
+
+When the oauth callback is completed successfully ShopifyApp will queue a background job which will ensure all the specified webhooks exist for that shop. Because this runs on every oauth callback it means your app will always have the webhooks it needs even if the user uninstalls and re-installs the app.
+
+
 ShopifyApp::SessionRepository
 -----------------------------
 
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ ShopifyApp can manage your app's webhooks for you (requires ActiveJob). Set whic
 ```ruby
 ShopifyApp.configure do |config|
   config.webhooks = [
-    {topic: 'carts/update', address: 'example-app.com/webhooks', format: 'json'}
+    {topic: 'carts/update', address: 'example-app.com/webhooks'}
   ]
 end
 ```
--- a/README.md
+++ b/README.md
@@ -124,6 +124,21 @@ end
 
 When the oauth callback is completed successfully ShopifyApp will queue a background job which will ensure all the specified webhooks exist for that shop. Because this runs on every oauth callback it means your app will always have the webhooks it needs even if the user uninstalls and re-installs the app.
 
+There is also a WebhooksController module that you can include in a controller that receives Shopify webhooks. For example:
+
+```ruby
+class WebhooksController < ApplicationController
+  include ShopifyApp::WebhooksController
+
+  def carts_update
+    SomeJob.perform_later(shopify_domain: shop_domain)
+    head :ok
+  end
+end
+```
+
+The module skips the `verify_authenticity_token` before_action and adds an action to verify that the webhook came from Shopify.
+
 
 ShopifyApp::SessionRepository
 -----------------------------
--- a/lib/shopify_app/sessions_controller.rb
+++ b/lib/shopify_app/sessions_controller.rb
@@ -13,9 +13,14 @@ module ShopifyApp
     def callback
       if response = request.env['omniauth.auth']
         shop_name = response.uid
-        sess = ShopifyAPI::Session.new(shop_name, response['credentials']['token'])
+        token = response['credentials']['token']
+
+        sess = ShopifyAPI::Session.new(shop_name, token)
         session[:shopify] = ShopifyApp::SessionRepository.store(sess)
         session[:shopify_domain] = shop_name
+
+        WebhooksManager.queue(shop_name, token) if ShopifyApp.configuration.has_webhooks?
+
         flash[:notice] = "Logged in"
         redirect_to return_address
       else
--- a/lib/shopify_app/webhooks_manager_job.rb
+++ b/lib/shopify_app/webhooks_manager_job.rb
@@ -1,0 +1,11 @@
+module ShopifyApp
+  class WebhooksManagerJob < ActiveJob::Base
+    def perform(params = {})
+      shop_name = params.fetch(:shop_name)
+      token = params.fetch(:token)
+
+      manager = WebhooksManager.new(shop_name, token)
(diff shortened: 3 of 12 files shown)


README.md
```diff
@@ -108,6 +108,23 @@ ShopifyApp.configure do |config|
...
+
+WebhooksManager
+---------------
+
+ShopifyApp can manage your apps webhooks for you (requires ActiveJob). Set which webhooks you require in the initializer:
+
+```ruby
+ShopifyApp.configure do |config|
+  config.webhooks = [
+    {topic: 'carts/update', address: 'example-app.com/webhooks', format: 'json'}
```
kevinhughes27: there isn't really any validation on what you set here. I think this should be fine though - most config type things don't really validate. Let me know what you guys think.

kevinhughes27: I am also debating making `json` default do it doesn't need to be set here

README.md
```diff
@@ -108,6 +108,23 @@ ShopifyApp.configure do |config|
 end
 ```
 
+
+WebhooksManager
+---------------
+
+ShopifyApp can manage your apps webhooks for you (requires ActiveJob). Set which webhooks you require in the initializer:
```
christhomson: `app's webhooks`

lib/shopify_app/sessions_controller.rb
```diff
@@ -13,9 +13,14 @@ def create
...
       if response = request.env['omniauth.auth']
         shop_name = response.uid
-        sess = ShopifyAPI::Session.new(shop_name, response['credentials']['token'])
+        token = response['credentials']['token']
+
+        sess = ShopifyAPI::Session.new(shop_name, token)
         session[:shopify] = ShopifyApp::SessionRepository.store(sess)
         session[:shopify_domain] = shop_name
+
+        WebhooksManager.run(shop_name, token) if ShopifyApp.configuration.has_webhooks?
```
christhomson: Without looking at the next file, I thought this was going to be doing inline HTTP calls. Maybe rename this to `run_later` or something along those lines?

test/controllers/sessions_controller_test.rb
```diff
@@ -76,6 +76,28 @@ class SessionsControllerTest < ActionController::TestCase
...
+      config.webhooks = [{topic: 'carts/update', address: 'example-app.com/webhooks', format: 'json'}]
+    end
+
+    ShopifyApp::WebhooksManager.expects(:run)
+
+    mock_shopify_omniauth
+    get :callback, shop: 'shop'
+  end
+
+  test "#callback should doesn't run the WebhooksManager if no webhooks are configured" do
```
christhomson: `#callback doesn't ` …

test/shopify_app/webhooks_manager_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  test "#destroy_webhooks makes calls to destroy webhooks" do
+    ShopifyAPI::Webhook.stubs(:all).returns(Array.wrap(all_mock_webhooks.first))
+    ShopifyAPI::Webhook.expects(:delete).with(all_mock_webhooks.first.id)
+
+    @manager.destroy_webhooks
+  end
+
+  test "#destroy_webhooks does not destroy webhook that do not have a matching address" do
```
christhomson: `that does not`, or pluralize `webhooks` :trollface:

test/shopify_app/webhooks_manager_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  private
+
+  def expect_webhook_creation(topic, address)
+    stub_webhook = stub(persisted?: true)
+    ShopifyAPI::Webhook.expects(:create).with(topic: topic, address: address, format: 'json').returns(stub_webhook)
+  end
+
+  def all_webhooks
```
christhomson: Maybe name this `all_webhook_topics`?

README.md
christhomson: > there isn't really any validation on what you set here. I think this should be fine though - most config type things don't really validate.

I'm not sure about this. It wouldn't be great if you typo'd the `topic` or `address` key names, but I guess the dev would notice quickly enough when CreationFailed gets raised. There's nothing we can do to prevent people from typoing the topic, since we don't want to include a full list of webhook topics in this gem, so I'd say just letting the dev run into `CreationFailed` would be fine here.

> I am also debating making json default do it doesn't need to be set here

This is a good idea. Personally, I would default it to `json` and not even mention that the format is an option.

christhomson:
:+1:

README.md
kevinhughes27: yeah I made json the default and I am not mentioning it. I used a reverse_merge so someone could spec xml if they really wanted but then they are on their own from there.

lib/shopify_app/sessions_controller.rb
kevinhughes27: I renamed it to `queue`.

christhomson: Ah yes that's much nicer than `run_later` :+1:

kevinhughes27: yeah sort of helps imply that you need a job queue as well.

kevinhughes27:
I added another module to this PR - a WebhooksController module that does a couple of things - it disables the `verify_authenticity_token` filter and adds a before action to verify the shopify webhook. It also adds a shop_domain helper method that grabs this info from the header

I was originally waiting to add this because I wanted to looked into updating the SessionStorage mixin to use the shopify_domain rather than just any uid but then I realized that this wasn't neccessary for the webhooks controller mixin. This is because I don't actually want to add a `with_shopify_session` type method to a webhook receiving controller since this would encouage responding to webhooks inline (a bad idea when dealing with Shopify) rather than queueing something to process them and responding :ok: immediately.

can I get another look at this!

lib/shopify_app/webhooks_manager_job.rb
```diff
@@ -0,0 +1,11 @@
+module ShopifyApp
+  class WebhooksManagerJob < ActiveJob::Base
+    def perform(params = {})
+      shop_name = params[:shop_name]
```
christhomson: `params.fetch(:shop_name)` and `params.fetch(:token)` would provide slightly better error messages if one of those keys is not provided for some reason.

test/shopify_app/webhooks_controller_test.rb
```diff
@@ -0,0 +1,50 @@
...
+  test "#carts_update should verify request" do
+    with_application_test_routes do
+      data = {foo: :bar}.to_json
+      @controller.expects(:validate_hmac).with('secret', data.to_s).returns(true)
+      post :carts_update, data
+      assert_response :ok
+    end
+  end
+
+  test "un-verifed request returns unauthorized" do
```
christhomson: typo: `un-verified`

christhomson:
Looks good to me.

t6d:
:+1:

PR #183, opened 2015-11-15
Fix typo in QUICKSTART.md

christianblais:
@kevinhughes27

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -29,7 +29,7 @@ git remote set-url heroku git@heroku.com:<name>.git
 now we need to let git know where the remote server is so we'll be able to deploy later
 
 web:
-```  
+```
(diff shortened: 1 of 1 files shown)


kevinhughes27:
wooops, thanks!

christianblais:
Speaking of which, what would you think of replacing the current ` -redirect_uri=` option with a ` -application-uri=` option, which we could then leverage for the redirect url by appending said `auth/shopify/callback` automatically, but could then also be used for other things, such as, say, providing callback URLs via shopify API?

kevinhughes27:
I think I'd rather have an explicit redirect_uri param since there is already a lot of confusion around getting the oauth stuff setup for the first time. I think having the application uri as another param might be useful though.

PR #184, opened 2015-11-15
Create shop helper with shop model generation

christianblais:
In all the apps I've been creating, in all apps I've seen on Shopify, and I'm pretty sure in all apps relying on the Shop model for memory store, we always need to retrieve the shop model at one point or another, from the database. Since this is a snippet of code we've been copying over and over again, I figure it could make sense to have it as part of the generator?

The output would be appended to `application_controller.rb`, allowing any other controller to rely on the `shop` helper.

What do you think? Worth it or not?

@kevinhughes27

### DIFF
--- a/lib/generators/shopify_app/shop_model/templates/shop_helper.rb
+++ b/lib/generators/shopify_app/shop_model/templates/shop_helper.rb
@@ -1,0 +1,7 @@
+
+  protected
+
+  def shop
+    @shop ||= Shop.find_by(id: session[:shopify])
+  end
+  helper_method :shop
--- a/test/generators/shop_model_generator_test.rb
+++ b/test/generators/shop_model_generator_test.rb
@@ -4,7 +4,11 @@ require 'generators/shopify_app/shop_model/shop_model_generator'
 class ShopModelGeneratorTest < Rails::Generators::TestCase
   tests ShopifyApp::Generators::ShopModelGenerator
   destination File.expand_path("../tmp", File.dirname(__FILE__))
-  setup :prepare_destination
+
+  setup do
+    prepare_destination
+    provide_existing_authenticated_controller
+  end
 
   test "create the shop model" do
     run_generator
@@ -18,7 +22,14 @@ class ShopModelGeneratorTest < Rails::Generators::TestCase
   test "creates ShopModel migration" do
     run_generator
     assert_migration "db/migrate/create_shops.rb" do |migration|
-      assert_match "create_table :shops  do |t|", migration
+      assert_match "create_table :shops do |t|", migration
+    end
+  end
+
+  test "injects into authenticated controller" do
+    run_generator
+    assert_file "app/controllers/shopify_app/authenticated_controller.rb" do |controller|
+      assert_match "def shop", controller
     end
   end
 
--- a/lib/generators/shopify_app/shop_model/shop_model_generator.rb
+++ b/lib/generators/shopify_app/shop_model/shop_model_generator.rb
@@ -23,6 +23,14 @@ module ShopifyApp
         copy_file 'shops.yml', 'test/fixtures/shops.yml'
       end
 
+      def inject_into_authenticated_controller
+        inject_into_file(
+          'app/controllers/shopify_app/authenticated_controller.rb',
+          File.read(File.expand_path(find_in_source_paths('shop_helper.rb'))),
+          before: "\nend"
+        )
+      end
+
       private
 
       def copy_migration(migration_name, config = {})
--- a/lib/generators/shopify_app/shop_model/templates/db/migrate/create_shops.rb
+++ b/lib/generators/shopify_app/shop_model/templates/db/migrate/create_shops.rb
@@ -1,6 +1,6 @@
 class CreateShops < ActiveRecord::Migration
   def self.up
-    create_table :shops  do |t|
+    create_table :shops do |t|
       t.string :shopify_domain, null: false
       t.string :shopify_token, null: false
       t.timestamps
--- a/test/app_templates/app/controllers/shopify_app/authenticated_controller.rb
+++ b/test/app_templates/app/controllers/shopify_app/authenticated_controller.rb
@@ -1,0 +1,7 @@
+module ShopifyApp
+  class AuthenticatedController < ApplicationController
+    before_action :login_again_if_different_shop
+    around_filter :shopify_session
+    layout ShopifyApp.configuration.embedded_app? ? 'embedded_app' : 'application'
+  end
+end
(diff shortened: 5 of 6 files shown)


lib/generators/shopify_app/shop_model/templates/shop_helper.rb
```diff
@@ -0,0 +1,9 @@
+
+  protected
+
+  def shop
+    @shop ||= Shop.find(session[:shopify])
```
lulessa: How about this for a more concise approach? It'll return nil if not found and you can drop the rescue block.
```ruby
@shop ||= Shop.find_by(id: session[:shopify])
```

lulessa:
:+1: This forces a little more consistent structure when building an app. I used a different model name for memory store in my first app and that led to a bit of confusion.

lib/generators/shopify_app/shop_model/templates/shop_helper.rb
christianblais: :+1: , will change.

kevinhughes27:
yeah I actually have a branch doing this on my local machine that wasn't quite finished so I am in favor of the idea.

I would call it `current_shop` though as this is what our apps use.

test/generators/shop_model_generator_test.rb
```diff
@@ -29,6 +33,13 @@ class ShopModelGeneratorTest < Rails::Generators::TestCase
     end
   end
 
+  test "injects into application controller" do
+    run_generator
+    assert_file "app/controllers/application_controller.rb" do |controller|
+      assert_match "def shop", controller
+    end
```
kevinhughes27: might be nice to inject into the authenticated controller instead since then we know we have a shop

resistorsoftware: It is now April, 2016. What is the consensus on having access to a persisted Shop model in the Rails App code as it stands?

kevinhughes27: v6 shipped in April 2015 which was the first version of the more modern rails engine. This major version also modularized the generators and included a shop modal generator which was ran by default. So effectively unless someone on purpose removed the shop modal anyone who generated an app in the last year would have the standard shop model.

Does that answer you question? It would be simpler on this code base to force the shop model and it is something I have considered before.

resistorsoftware: Ok. I have no idea what that means, but I will persist in poking around with this neat-o app. I just want to know the shop logged in when creating an App.. seems I have to dig into this concept of the engine...

kevinhughes27: The engine provides a simple shop model - it will have created a migration for you and setup the SessionRepository to use your shop model. The SessionRepository is really just a connector object from the AuthController to whichever object stores your session (the shop model)

Simple answer - if you ran the recommended generator then any controller that inherits from `ShopifyApp::Authenticated` controller can call `shop` which will return the current shop

resistorsoftware: The Shop model I see, I assume is from the new App:

      class Shop < ActiveRecord::Base
        include ShopifyApp::Shop
        include ShopifyApp::SessionStorage
      end

Yet in my Home Controller, which inherits from AuthenticatedController, any call to shop is met with nil... 

     class HomeController < ShopifyApp::AuthenticatedController
         def index
             puts "Working on shop #{shop.shopify_domain}"
             @products = ShopifyAPI::Product.find(:all, params: { limit: 10 })
         end
     end

I guess that is why I am puzzled about where I went wrong.

kevinhughes27: What's in your `session[:shopify]`? 

This is the contents of the shop method:

```
@shop ||= Shop.find_by(id: session[:shopify])
```

Can you run that and let me know what happens? Its possible this method needs to be updated

resistorsoftware: session[:shopify] == the ID of the shop from the Shop model...

resistorsoftware: I get nothing for *shop* in any of the controllers while session[:shopify] is set. I ran the generators and am fiddling with a setup that comes pretty much straight from the current code at Github. I think I need to add the instantiation of shop in the application controller manually as per the point of this discussion...

kevinhughes27: oh shoot - this isn't merged. You'll need to add the shop method. Sorry for the confusion. The ID is stored in the session so you can write a method to fetch the model using the ID.

We should probably merge this but we need to figure out if we are making the Shop model part of this Engine as non optional or not.

resistorsoftware: Indeed. I just added the shop method and all works as advertised. Sorry for the intrusion on the old thread. Just wanted to clear the air a bit on this issue. It's critical for my Apps to have *shop* at my disposal throughout the Apps I build. I can understand though why there can be a circus around this for implementation.

*Thanks* for the App!

kevinhughes27:
@christianblais we should probably finish this as this PR can be merged without dealing with the larger issue of whether we should force the shop model or not. This is an enhancement to the shop model generator which adds a method that everyone using the shop model would want.

Comment about current_shop stands though and should be addressed. If you want me to finish this up I can.

christianblais:
If you have the time go for it, would probably require some rebase. Otherwise I may have some time end of week.

kevinhughes27:
I'll leave it for you then 😛

christianblais:
Switched to AuthenticatedController. Ready for review I guess?

lib/generators/shopify_app/shop_model/templates/shop_helper.rb
```diff
@@ -0,0 +1,7 @@
+
+  protected
+
+  def shop
+    @shop ||= Shop.find_by(id: session[:shopify])
+  end
+  helper_method :shop
```
kevinhughes27: what do you think about `current_shop` instead? This is how most the apps I work on have done it and it matches Devise which people may be familiar with.

kevinhughes27:
one comment LGTM

resistorsoftware:
I like just shop and not current_shop. I know Rails has the venerable
current_user concept, but current_shop is just too long-winded coming from
someone used to just using shop everywhere...

Dave Lazar
Resistor Software Inc.

Book an Appointment!
https://calendly.com/dlazar

lib/generators/shopify_app/shop_model/templates/shop_helper.rb
christianblais: I hate it 😄

kevinhughes27:
Okay looks like its 2 votes for `shop` and only me for `current_shop`. I don't have a strong opinion here so let :shipit:

lib/generators/shopify_app/shop_model/templates/shop_helper.rb
kevinhughes27: :trollface:

PR #185, opened 2015-11-17
Require omniauth-shopify-oauth2 ~> 1.1.11.

clayton-shopify:
@kevinhughes27 @EiNSTeiN- @christhomson

omniauth-shopify-oauth2 1.1.11 fixes an incompatibility with omniauth-oauth2 1.4.0 (https://github.com/Shopify/omniauth-shopify-oauth2/pull/32), so we should require that version.

### DIFF
--- a/shopify_app.gemspec
+++ b/shopify_app.gemspec
@@ -11,7 +11,7 @@ Gem::Specification.new do |s|
   s.add_dependency('rails', '>= 3.1', '< 5.0')
 
   s.add_runtime_dependency('shopify_api', '~> 4.0.2')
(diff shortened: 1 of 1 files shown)


EiNSTeiN-:
:+1:

christhomson:
:sheep: :it:

kevinhughes27:
@clayton-shopify If I understand the changes correctly then we should be able to remove this override here correct? https://github.com/Shopify/shopify_app/blob/master/lib/generators/shopify_app/install/templates/shopify_provider.rb#L7

PR #188, opened 2015-11-26
remove callback url override since its been fixed upstream

kevinhughes27:
thanks to #185 and the new omniauth gem we don't need to set the callback url any more.

@clayton-shopify @christhomson


christianblais:
While you're there, mind to remove the empty lines and move to the new hash syntax? :smile:

clayton-shopify:
@kevinhughes27 That's correct, this shouldn't be needed anymore. :+1:

kevinhughes27:
@christianblais sure

christianblais:
:heart:

PR #190, opened 2015-12-01
Use I18n to change app title

gmalette:
I don't want to have to copy this page to my app, so I'd rather be able to use I18n to change the page form `Shopify App - Installation` to `Script Editor`.

@christianblais @kevinhughes27

### DIFF
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <title>Shopify App - Installation</title>
+  <title><%= t("shopify_app.installation.title", default: "Shopify App - Installation") %></title>
 
   <style>
     html, body { padding: 0; margin: 0; }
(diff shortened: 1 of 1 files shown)


christianblais:
Should you be able to rely on I18n for every other strings of the page then, like the "Install" button?

gmalette:
I didn't see a need to change the others; they're fine as-is for most if not all use cases, whereas "Shopify App" is clearly just a placeholder

christianblais:
Except if you don't speak English.

gmalette:
Well, admin is english only

kevinhughes27:
Would it make more sense to add this to the engine config rather than
adding another thing to configure?

Really like the idea though

gmalette:
> Would it make more sense to add this to the engine config rather than
> adding another thing to configure?

I'm not sure; if we did want to add the label for the install button, or other labels elsewhere, would you also put them in the engine config? It seems to me like view strings do belong in i18n

christianblais:
As I said, I'd add i18n for everything while we're there, but up to you, not a blocker. I also think that I18n is the place for that kind of thing.

:+1:

kevinhughes27:
For more strings I agree I was more thinking if it is just the App title. Either way can you add some documentation for this please

PR #194, opened 2015-12-03
version bump after markup modernization

nwtn:
This PR updates some of the generated view markup and gets the output a little closer to how we use these apps. This is based on what we’re currently doing in channels, so forgive me if I'm making some faulty assumptions about what other apps might do.

Changes:

* more semantic/accessible markup
* remove the flash layout view, since we're no longer exposing errors in this way
* move render-blocking javascript to the bottom of the page to improve render time
* version bump, since some of this could be considered breaking changes (especially the flash message stuff)

closes https://github.com/Shopify/assets/issues/262

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,11 @@
+6.4.0
+-----
+
+* More semantic and accessible markup in the sessions/new, embedded_app, and product
+  index views.
+* Moved all JavaScript to load at the bottom of the page instead of the head, for
+  page loading better performance.
+
 6.3.0
 -----
 
--- a/README.md
+++ b/README.md
@@ -178,6 +178,14 @@ $ bundle exec spring stop
 
 Run shopify_app generator again.
 
+Testing an embedded app outside the Shopify admin
+-------------------------------------------------
+
+By default, loading your embedded app will redirect to the Shopify admin, with the app view loaded in an `iframe`. If you need to load your app outside of the Shopify admin (e.g., for performance testing), you can change `forceRedirect: false` to `true` in `ShopifyApp.init` block in the `embedded_app` view. To keep the redirect on in production but off in your `development` and `test` environments, you can use:
+
+```javascript
+forceRedirect: <%= Rails.env.development? || Rails.env.test? ? 'false' : 'true' %>
+```
 
 Questions or problems?
 ----------------------
--- a/lib/generators/shopify_app/install/templates/embedded_app.html.erb
+++ b/lib/generators/shopify_app/install/templates/embedded_app.html.erb
@@ -25,7 +25,7 @@
         apiKey: "<%= ShopifyApp.configuration.api_key %>",
         shopOrigin: "<%= "https://#{ @shop_session.url }" if @shop_session %>",
         debug: <%= Rails.env.development? ? 'true' : 'false' %>,
-        forceRedirect: <%= Rails.env.development? || Rails.env.test? ? 'false' : 'true' %>
+        forceRedirect: false
       });
     </script>
 
--- a/lib/shopify_app/version.rb
+++ b/lib/shopify_app/version.rb
@@ -1,3 +1,3 @@
 module ShopifyApp
-  VERSION = '6.3.0'
+  VERSION = '6.4.0'
 end


lib/generators/shopify_app/install/install_generator.rb
```diff
@@ -53,17 +53,21 @@ def inject_embedded_app_options_to_application
...
+          injection << "  layout \"embedded_app\"\n"
+        end
+
         inject_into_class(
           "app/controllers/application_controller.rb",
           'ApplicationController',
-          "  include ShopifyApp::Controller\n"
+          injection
         )
       end
```
christianblais: Since we're bumping major version, could we imagine the app not having to inject anything in an already existing file, but rather simply inject a module with `ActiveSupport.on_load(:action_controller)` in an initializer or something like that? I think @gmalette also had some ideas around that stuff.

nwtn: seems like a good idea. @christianblais @gmalette feel free to add commits to this pr.

christianblais:
Seems like this PR does a bit much. Mind to explode this in smaller chunks?

nwtn:
sure, that's fair. will throw up some new ones early next week.

kevinhughes27:
smaller chunks might be nice and can we also update the CHANGELOG? Love seeing more people contributing to this repo!

lib/generators/shopify_app/install/templates/_flash_messages.html.erb
```diff
@@ -1,13 +0,0 @@
...
-  document.addEventListener(eventName, function() {
-    <% if flash[:notice] %>
-      ShopifyApp.flashNotice("<%= j flash[:notice].html_safe %>");
-    <% end %>
-
-    <% if flash[:error] %>
-      ShopifyApp.flashError("<%= j flash[:error].html_safe %>");
-    <% end %>
-  });
-</script>
```
kevinhughes27: I'm not sure about removing this as its required for ESDK flash message to work. How are you showing your flash message errors?

lib/generators/shopify_app/install/install_generator.rb
gmalette: I think there's no good way to get around _this_ one specifically, but all the layouts should be optional and fall back by default on the app's.

kevinhughes27: I tried a bunch of different things try and make this more slick but couldn't come up with anything. Happy to brainstorm though

gmalette: @kevinhughes27 you mean about the views or the `include ShopifyApp::Controller` ?

kevinhughes27: I meant about the explicit layout set for esdk.

Including the concerns is nice imo because it gives an easy way to share code and for users to discover the code.

nwtn:
Broke everything up into smaller PRs except the flash message stuff and the version bump. 

* @kevinhughes27 you’re totally right, the flash message stuff shouldn’t be deleted, just updated to use banners.
* @kevinhughes27 I’ll write up a changelog message to go along with the version bump.

I’m going to leave this PR open until I have PRs for those last two bits ready (probably later today).

kevinhughes27:
sounds good! Thanks for shipping your updates here!

nwtn:
about to do a PR for the version bump. now that i've left the flash message stuff basically the same and haven't touched the default layout bit, any reason to consider this a major change? i'm thinking bumping to 6.4 is more appropriate.

kevinhughes27:
yeah I think that should be fine

nwtn:
I updated this PR to only have the version bump + CHANGELOG updates. Before this merges, I’m going to try :tophat:ing all the other changes that I’ve merged in. A couple other :tophat:s would be appreciated if possible @kevinhughes27 @tmlayton.

kevinhughes27:
I usually generate a new app and make sure it installs as a :tophat:. I will try one out later today.

kevinhughes27:
so I tophatted and I found one issue (not sure if it was your code though it may have been #196) After I install the app I am not redirected into the embedded app frame but the view still loads:

image
(not inside the admin)

If I reload the app from clicking in my admin I do end up inside the esdk

nwtn:
Haven't had a chance to take a look yet, but one of the changes I pushed was to disable the redirect in dev (but it should still redirect in prod environments). This has been helpful for me for testing some stuff in channels, but I understand it may not be a smart default. Happy to change that back if you think it's best.

kevinhughes27:
Can you give some more information about what removing this helped you to do? It seems like it would be quite confusing for a developer new to the platform to have the app not behave the same locally as it would in production. Especially since this would require diving pretty deep into this code or reading docs to figure out why it is behaving differently.

nwtn:
I was using it to look at page load performance. Doing it like this made it much easier to look at performance for the app rather than the admin as a whole. I think you’re right, though,  this might be too confusing if you’re not expecting it. I'm happy to change this behaviour back.

nwtn:
cc @djoume @bradsokol

kevinhughes27:
Makes sense. Lets put the expected behaviour back. It would be cool to make a readme note about doing perf testing though

nwtn:
updated to previous behaviour and added readme note

kevinhughes27:
:rocket:

PR #198, opened 2015-12-07
Markup for embedded app view

nwtn:
Breaking up #194 into smaller PRs. This one moves to more performant/semantic/accessible markup for the embedded app view that’s consistent with what we’re doing in channels apps.

### DIFF
--- a/lib/generators/shopify_app/install/templates/embedded_app.html.erb
+++ b/lib/generators/shopify_app/install/templates/embedded_app.html.erb
@@ -8,7 +8,14 @@
   </head>
 
   <body>
-    <%= yield %>
+    <div class="app-wrapper">
+      <div class="app-content">
+        <main role="main">
+          <%= yield %>
+        </main>
+      </div>
+    </div>
+
     <%= render 'layouts/flash_messages' %>
 
     <script src="//cdn.shopify.com/s/assets/external/app.js?<%= Time.now.strftime('%Y%m%d%H') %>"></script>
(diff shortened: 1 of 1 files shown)


kevinhughes27:
:+1:

lib/generators/shopify_app/install/templates/embedded_app.html.erb
```diff
@@ -1,24 +1,40 @@
...
-  <script type="text/javascript">
-    ShopifyApp.init({
-      apiKey: "<%= ShopifyApp.configuration.api_key %>",
-      shopOrigin: "<%= "https://#{ @shop_session.url }" if @shop_session %>",
-      debug: <%= Rails.env.development? ? 'true' : 'false' %>
-    });
-  </script>
+  <body>
+    <div class="app-wrapper">
+      <div class="app-content">
```
tmlayton: Are you using these classes for something?

tmlayton:
Just one comment about the classes and wrapping divs, otherwise LGTM

lib/generators/shopify_app/install/templates/embedded_app.html.erb
nwtn: For apps using the assets gem, yes: https://github.com/Shopify/assets/blob/master/app/assets/stylesheets/shopify_assets/embedded_app_styles/_layout.scss#L10-L23. For apps that aren’t I figured they might still be convenient styling hooks.

tmlayton: I personally wouldn't include this. A 3rd party using this won't be clear on why the wrapping div+classes exist since there are no styles packaged and the gem isn't available for 3rd parties yet. That being said I don't feel strongly so feel free to ship as-is

nwtn: I’m going to do this anyway because it’s not really harmful, it'll be useful to us internally, and 3rd parties without the gem can still use it if they want. :)

PR #201, opened 2015-12-07
Embedded app default layout

nwtn:
Breaking up #194 into smaller PRs. This makes `embedded_app` the default layout if `embedded_app?` is `true`.

There was discussion in the other thread about the best way to handle this. I’m no Ruby dev, so I’ll defer to you smart people on this — feel free to suggest changes or even add commits to this branch.

@christianblais @gmalette @kevinhughes27 @Shopify/channels-fed

### DIFF
--- a/test/generators/install_generator_test.rb
+++ b/test/generators/install_generator_test.rb
@@ -63,7 +63,15 @@ class InstallGeneratorTest < Rails::Generators::TestCase
     end
   end
 
-  test "injects into application controller" do
+  test "injects into application controller for embedded app" do
+    run_generator
+    assert_file "app/controllers/application_controller.rb" do |controller|
+      assert_match "  include ShopifyApp::Controller\n  layout \"embedded_app\"\n", controller
+    end
+  end
+
+  test "injects into application controller for non-embedded app" do
+    stub_embedded_false
     run_generator
     assert_file "app/controllers/application_controller.rb" do |controller|
       assert_match "  include ShopifyApp::Controller\n", controller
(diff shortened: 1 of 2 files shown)


kevinhughes27:
We actually already set this default here on the authenticated controller: https://github.com/Shopify/shopify_app/blob/master/app/controllers/authenticated_controller.rb#L4

Its kind of nice to have less generated code so I think its better the way it currently is - why did you want to propose this change? Did you miss the authenticated controller?

test/generators/install_generator_test.rb
```diff
@@ -73,7 +81,6 @@ class InstallGeneratorTest < Rails::Generators::TestCase
   test "creates the embedded_app layout" do
     run_generator
     assert_file "app/views/layouts/embedded_app.html.erb"
-    assert_file "app/views/layouts/_flash_messages.html.erb"
```
kevinhughes27: I don't think we want to remove this line in this PR

nwtn:
I totally missed it in the authenticated controller! I'm happy to leave this as-is. Closing.

kevinhughes27:
yeah I sort of worry that this controller isn't obvious enough but I think it is a good pattern to have a sub-controller for the Shopify area of the app and not assume that the entire app is just Shopify.

nwtn:
I think it’s ok, I was just being a bit dense when I missed it. I think this is a totally reasonable way to approach it.

kevinhughes27:
okay we'll leave it then.

PR #202, opened 2015-12-07
yield flash message js

nwtn:
Breaking up #194 into smaller PRs. Decided to leave the flash message stuff more or less alone since banners would require the assets gem. This PR just puts the flash JS into a content_for :javascript block.


kevinhughes27:
:+1:

tmlayton:
:+1:

PR #205, opened 2016-01-05
Change class monkeypatch to module monkeypatch

gmalette:
@kevinhughes27 @nwtn 

This will help me test generators in another gem


kevinhughes27:
:+1:

Those helpers pretty much come directly from thoughbot/clearance, I may have had to change them a bit. I feel like they should just be party of rails

gmalette:
I totally agree, but the gem I'm working on depends on how `ShopifyApp` does things, so I can't go half-measures. If the generators change, the gem's tests should break.

kevinhughes27:
:+1:

nwtn:
lgtm

PR #207, opened 2016-01-11
Extract HomeController generator

gmalette:
I don't know why someone would want `ShopifyApp` to create `HomeController` for them, but I didn't want to take product decisions. I just extracted it from `InstallGenerator`, so now it only runs when you do `rails g shopify_app` and not `rails g shopify_app:install`.

The reason behind this PR is that the generator breaks any existing app that have a `root` route, so basically every app.

@Krystosterone @kevinhughes27


Krystosterone:
This makes sense and will really help us for `shopify_app_platform` :+1:

gmalette:
@nwtn

kevinhughes27:
Yup this makes sense. The non-full generator isn't as commonly used so I'm not surprised to find usage issues with it (most people just run the full generator). This solution makes sense.

The main reason for the home_controller was to showcase how to setup the basic ESDK views and also to show the AuthenticatedController which should be inherited from.

nwtn:
lgtm

gmalette:
I'll bump the version

gmalette:
> The main reason for the home_controller was to showcase how to setup the basic ESDK

That makes sense

Krystosterone:
:tada:

PR #210, opened 2016-01-23
Adds ScripttagManager and job to auto create ScriptTag on shop creation

### LINKED ISSUE #209: Create ScriptTags automatically when shop is added to app
Similar to how the webhooks manager works, it would be nice to have ScriptTags created automatically once a shop is added to an app.

redronin:
resolves #209 

I've basically followed the same pattern of the WebhooksManager and WebhooksManagerJob to create a ScripttagsManager and ScripttagsManagerJob.

I've updated readme with instructions. I've tested this on my developing shopify app.

### DIFF
--- a/test/shopify_app/scripttags_manager_test.rb
+++ b/test/shopify_app/scripttags_manager_test.rb
@@ -31,16 +31,6 @@ class ShopifyApp::ScripttagsManagerTest < ActiveSupport::TestCase
     end
   end
 
-  test "#create_scripttags when creating a scripttag fails and the scripttag exists, do not raise an error" do
-    scripttag = stub(persisted?: false)
-    scripttags = all_scripttag_srcs.map{|t| stub(src: t)}
-    ShopifyAPI::ScriptTag.stubs(create: scripttag, all: scripttags)
-
-    assert_nothing_raised ShopifyApp::ScripttagsManager::CreationFailed do
-      @manager.create_scripttags
-    end
-  end
-
   test "#recreate_scripttags! destroys all scripttags and recreates" do
     @manager.expects(:destroy_scripttags)
     @manager.expects(:create_scripttags)
--- a/test/shopify_app/scripttags_manager_test.rb
+++ b/test/shopify_app/scripttags_manager_test.rb
@@ -1,0 +1,83 @@
+require 'test_helper'
+
+class ShopifyApp::ScripttagsManagerTest < ActiveSupport::TestCase
+
+  setup do
+    ShopifyApp.configure do |config|
+      config.scripttags = [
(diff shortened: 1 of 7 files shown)


kevinhughes27:
:clap: I'll review on monday!

lib/shopify_app/sessions_controller.rb
```diff
@@ -20,6 +20,7 @@ def callback
         session[:shopify_domain] = shop_name
 
         WebhooksManager.queue(shop_name, token) if ShopifyApp.configuration.has_webhooks?
+        ScripttagsManager.queue(shop_name, token) if ShopifyApp.configuration.has_scripttags?
```
kevinhughes27: we could actually break these into a method here and then that method could be overridden (with a super) to allow custom jobs to be queued from this point as well. Essentially the same thing I have on shopify-sinatra-app an after_auth_callback.

Lets not make that change in this PR, I'm just pointing it out for the discussion in #209

test/shopify_app/scripttags_manager_test.rb
```diff
@@ -0,0 +1,83 @@
...
+    scripttag = stub(persisted?: false, errors: stub(messages: ["Source needs to be https"]))
+    ShopifyAPI::ScriptTag.stubs(create: scripttag)
+
+    assert_raise ShopifyApp::ScripttagsManager::CreationFailed do
+      @manager.create_scripttags
+    end
+  end
+
+  test "#create_scripttags when creating a scripttag fails and the scripttag exists, do not raise an error" do
+    scripttag = stub(persisted?: false)
```
kevinhughes27: maybe I am confused but shouldn't this be true?

kevinhughes27:
one question then I'll merge!

ssuperczynski:
Hey @redronin any information when this feature might be done. Do you need some help?

redronin:
@kevinhughes27 oops...I'll take a look a this now.

test/shopify_app/scripttags_manager_test.rb
redronin: Sorry for replying long ago - I guess I didn't see the messages.
I've removed this test because it doesn't actually make any sense - you would never actually create a scripttag if it already exists - there's a check to make sure of that. And if it did fail, it should raise an error.

kevinhughes27:
Awesome thanks! I'll merge later this week! (you just missed being in v7 :cry:)

kevinhughes27:
Looks like it needs a rebase (Im thinking git isn't smart enough to auto merge the readme)

redronin:
@kevinhughes27 I rebased and fixed conflict

Issue #193, opened 2015-12-03
When installing app, the redirect_uri should contain https, but is using http instead on some computers


gduarte93:
After submitting the form on `https://myapp.herokuapp.com/login` on my some of my computers, the users will be (correctly) redirected to:

`https://mystore.myshopify.com/admin/oauth/authorize?client_id={api_key}&scope={scopes}&redirect_uri=https://mystore.herokuapp.com/auth/shopify/callback&state={nonce}`

There, they can login to their store and install the app.

However, on a couple of my computers, the user is redirected, instead, to:

`https://{store}.myshopify.com/admin/oauth/authorize?client_id={api_key}&scope={scopes}&redirect_uri=http://myapp.herokuapp.com/auth/shopify/callback&state={nonce}`

Where they will see the error **Oauth error invalid_request: The redirect_uri is missing or not whitelisted**. For some reason on those machines, it is redirecting using http instead of https. Manually adding the 's' into the address bar will allow them to install the app correctly, but how can I make it so that it always uses the https redirect_uri?

Shopify Partners Dashboard:

```
Callback URL:    https://myapp.herokuapp.com/
Application URL: https://myapp.herokuapp.com/
Redirection URL: https://myapp.herokuapp.com/auth/shopify/callback
```

Code:

**shopify_app.rb**

```
ShopifyApp.configure do |config|
  config.api_key = "my_key"
  config.secret = "my_secret"
  config.redirect_uri = "https://myapp.herokuapp.com/auth/shopify/callback"
  config.scope = "read_orders, read_products" # install assets/snippets?
  config.embedded_app = true
end
```

**omniauth.rb**

```
Rails.application.config.middleware.use OmniAuth::Builder do
  provider :shopify,
    ShopifyApp.configuration.api_key,
    ShopifyApp.configuration.secret,
    redirect_uri: ShopifyApp.configuration.redirect_uri,
    scope: ShopifyApp.configuration.scope
end
```

**routes.rb**

```
Rails.application.routes.draw do

  controller :sessions do
    get 'login' => :new, :as => :login
    post 'login' => :create, :as => :authenticate
    get 'auth/shopify/callback' => :callback
    get 'logout' => :destroy, :as => :logout
  end

  root :to => 'home#index'

end
```

kevinhughes27:
Nice issue write up - that is really quite bizarre.

[...]

PR #213, opened 2016-02-04
Adds login_url to login_protection to allow https redirection properly

jasonbuehler:
I was having a problem with my application being redirected to a path with http instead of https. This allows the full url to be passed along with the secure protocol on this redirection.

### DIFF
--- a/lib/shopify_app/login_protection.rb
+++ b/lib/shopify_app/login_protection.rb
@@ -39,14 +39,14 @@ module ShopifyApp
         head :unauthorized
       else
         session[:return_to] = request.fullpath if request.get?
-        redirect_to login_path(shop: params[:shop])
+        redirect_to login_url(shop: params[:shop])
       end
     end
 
     def close_session
       session[:shopify] = nil
       session[:shopify_domain] = nil
-      redirect_to login_path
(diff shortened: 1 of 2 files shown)


kevinhughes27:
interesting - kind of makes sense. Do you think this is the cause of #193?

Code lgtm, @gmalette can you take a look as well

jasonbuehler:
It sure looks like that could be the same issue. My problem was occurring on oauth redirection to the application, not during app authorization but when I hit actions directly that were wrapped in the shopify_session around filter. I hadn't actually tried authorizing my app from a shop for a while so it could have been there too. I think I had converted to https since the last time I authorized the app from a shop so the problem could have been present and I hadn't seen it yet.

gmalette:
@kevinhughes27 I don't know enough in this part to properly review, you know anyone that would?

kevinhughes27:
okay no big deal I am good with this change just thought you might want to take a look. I will :tophat: on monday and then merge

kevinhughes27:
I :tophat:ed and this works fine. I did notice that both this branch and master don't load the ESDK properly after install but thats an unrelated thing to look into.

Thanks for the PR

PR #214, opened 2016-02-08
use https for esdk include

kevinhughes27:
This include has been like this in our examples forever as far as I know but it recently started working (I did a bit of digging and it looks like chrome is getting more strict about mixed content now). This updates the generator to explicitly use https to grab the ESDK

@celsodantas @kmcphillips


kmcphillips:
:+1:

christianblais:
:+1:

PR #217, opened 2016-02-12
Add WebhooksController which delegates to Jobs

kevinhughes27:
This PR adds a WebhooksController to the ShopifyApp engine which further simplifies using Webhooks in an app. Previously we gave you all the utilities needed to handle Webhooks but you still had to do it yourself. With this new controller the developer doesn't even have to think about the controller anymore and just defines Webhook jobs which the controller automatically delegates too.

If the job is not found I raise a `ShopifyApp::MissingWebhookJobError`

Changes
------------
* Adds new webhooks controller to the engine
* updates the install generator to mount the engine and the home generator just adds the home route
* rename the old WebhooksController module to WebhookVerification to avoid confusion and more accurately describe the module
* move routing tests to their own directory

I updated this in my webhook demo app here: https://github.com/kevinhughes27/shopify-surge-pricing/compare/wip?expand=1 if you want to see the simplifications it makes

For review @Hammadk @gmalette @gavinballard (how does this stack up with your solution?)


gavinballard:
@kevinhughes27 Awesome, will take a look and let you know!

gavinballard:
@kevinhughes27, this looks really similar to my implementation. Key difference is that I only have a single route for all webhooks and dispatch based on the the `HTTP_X_SHOPIFY_TOPIC` header, rather than relying on the `:type` parameter in the URL as you've done. Your approach is probably better as it makes debugging webhook issues easier as you can see in the URL which topic something relates to.

CHANGELOG
```diff
@@ -1,3 +1,8 @@
+Unreleased
+----------
+* rename WebhooksController module to WebhookVerification
```
gmalette: this sounds like a breaking change

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def receive
+    job_args = {shop_domain: shop_domain, webhook: webhook}
+    webhook_job_klass.perform_later(job_args)
+    head :ok
+  end
+
+  private
+
+  def webhook
```
gmalette: I'd name this `webhook_params`

lib/generators/shopify_app/install/install_generator.rb
```diff
@@ -67,6 +67,9 @@ def create_embedded_app_layout
         end
       end
 
+      def mount_engine
+        route "mount ShopifyApp::Engine, at: '/'"
```
gmalette: We'll have to test this against platform, but it should work

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,27 @@
+class WebhooksController < ApplicationController
```
gmalette: Any reason this isn't in the `ShopifyApp` namespace?

gmalette: Including it in apps already defining a `WebhooksController` could lead to interesting results

CHANGELOG
kevinhughes27: it is, WebhooksController was only added in a recent version (6.3.0) so not sure how many people have adopted it yet. I'll update the note and when we release we'll bump to 6.5.0

app/controllers/webhooks_controller.rb
kevinhughes27: Its a good point, none of our controllers are in the ShopifyApp namespace but they probably should be. Should probably be a different PR though

CHANGELOG
gmalette: You mean when we bump to `7.0.0`? Unless we don't believe in SEMVER

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  private
+
+  def webhook
+    params.except(:controller, :action, :type)
+  end
+
+  def webhook_job_klass
+    "#{webhook_type.classify}Job".constantize
```
gmalette: `safe_constantize`

CHANGELOG
kevinhughes27: whoops I thought the middle digit allowed breaking changes. Yup 7.0 it is

lib/generators/shopify_app/routes/templates/routes.rb
```diff
@@ -5,3 +5,7 @@
     get 'auth/shopify/callback' => :callback
     get 'logout' => :destroy, :as => :logout
   end
+
+  namespace :webhooks do
+    post ':type' => :receive
+  end
```
gmalette: what's this template for, given the engine's routes are mounted?

kevinhughes27: if you run the routes generator then we unmount the engine and copy the routes that would have been mounted over directly, the same way we will also copy over the controllers and views. It lets you choose between having everything buried in the engine or copied into your app. Quite a few engines work this way (e.g. devise).

It might be worth discussing removing the generators for this though since you can still do this manually and its more of a power used thing.

gmalette:
:+1:

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,27 @@
+class WebhooksController < ApplicationController
+  include ShopifyApp::WebhookVerification
+
+  class ShopifyApp::MissingWebhookJobError < StandardError; end
+
+  def receive
+    job_args = {shop_domain: shop_domain, webhook: webhook_params}
+    webhook_job_klass.perform_later(job_args)
+    head :ok
```
Hammadk: You are returning `head :ok` from each webhook instead of `head :no_content`. I don't think there is a functional difference (200 response vs 204), but I’d check just in case.

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  private
+
+  def webhook_params
+    params.except(:controller, :action, :type)
+  end
+
+  def webhook_job_klass
+    "#{webhook_type.classify}Job".safe_constantize
+  rescue NameError => e
```
Hammadk: What's raising the NameError? `"test".classify.safe_constantize` and `"".classify.safe_constantize` both return nil.

gmalette: it's probably because it used to use `constantize` and wasn't updated

kevinhughes27: aaah woops this used to constantize. I should have looked into it before making that change suggested by @gmalette. This should be just constantize

app/controllers/webhooks_controller.rb
kevinhughes27: I want this to raise if the constant is not found so I can raise the specific error about the job not existing

app/controllers/webhooks_controller.rb
kevinhughes27: this is a good point. They both work but `no_content` is more correct I think. I'll update

app/controllers/webhooks_controller.rb
gmalette: > This should be just constantize

Careful with `constantize`, it respects Ruby's constant lookup which is generally not what you want. In this case it _probably_ doesn't change a thing, but it's better to keep sane habits. Also, testing against `nil` isn't very hard, and it's faster than exception-based flow control. 

Ref: http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-safe_constantize

gmalette: > I want this to raise if the constant is not found so I can raise the specific error about the job not existing

Why not raise if it's `nil`?

kevinhughes27: That works for me :+1:

kevinhughes27:
@gmalette do you want me to wait on merging until you test with platform? 

I'll put up a namespace PR in the near future (before 7.0.0) release. I think for v7 we should remove the controller, view and routes generator from the `all` generator since we really should be having people use the engine as much as possible

gmalette:
It's a breaking bump so go ahead and merge

gmalette:
Also good idea for the generators changes

Hammadk:
This is a great approach :+1:

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,25 @@
+class WebhooksController < ApplicationController
+  include ShopifyApp::WebhookVerification
+
+  class ShopifyApp::MissingWebhookJobError < StandardError; end
+
+  def receive
+    job_args = {shop_domain: shop_domain, webhook: webhook_params}
+    webhook_job_klass.perform_later(job_args)
```
gmalette: I thought I said this but looks like I didn't: using background jobs is a great way to control the pressure given by webhooks during a flash sale, so encouraging this is great!

kevinhughes27: yeah! The best part is that with active job you can run it without a proper queue and add the queue for production. Basically it doesn't complicate development at all and its done proper.

PR #219, opened 2016-02-18
remove routes, views and controllers generators from the all generator

kevinhughes27:
for the upcoming major version bump I am removing the controller, views and routes generator from the default all generator. Its unlikely that a developer needs to run these generators and it is preferred to let the engine own as much code as possible.

This is similar to how devise (probably the most popular engine) handles things.

I updated the readme to explain the process better.

@Krystosterone @gmalette

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,15 @@ Now we are ready to run any of the shopify_app generators. The following section
 Generators
 ----------
 
+### Default Generator
+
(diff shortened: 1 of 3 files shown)


gmalette:
Cool! We may have a few adjustments to do but nothing critical! Thanks for this

Krystosterone:
Nice! Thank you for the heads up! :+1:

PR #223, opened 2016-02-29
Add namespace isolation for the Engine

kevinhughes27:
As per a request by @gmalette this PR isolates the Engine into a namespace. With each new version ShopifyApp has transitioned closer to being a full rails engine and less of a generator. Namespaces is the next logical step.

Changes
------------
* isolates the namespace of the Engine from the host app. This is a better practise for the engine since it makes it less likely to conflict with the host app or other engines
* I fixed up the test routing to use the latest engine testing guidelines (thanks for the help @rafaelfranca)
* added rails rake tasks to the Rakefile to print the routes of the engine during development

Next
------
* Ship v7.0.0 of ShopifyApp
* update the example app directory

For review
@gmalette @jduff


kevinhughes27:
@maximebedard can you review instead of @gmalette please!

jduff:
I don't see any issues with this, looks like a nice cleanup and restructuring :+1:

lib/shopify_app/sessions.rb
```diff
@@ -1,5 +1,5 @@
 module ShopifyApp
-  module SessionsController
```
gmalette: `SessionsConcern` maybe?

gmalette: Do we have a standard way of naming concerns? I just saw `LoginProtection` so I guess not

test/routes/sessions_routes_test.rb
```diff
@@ -4,24 +4,24 @@ class SessionsRoutesTest < ActionController::TestCase
...
   setup do
     @routes = ShopifyApp::Engine.routes
+    @controller = 'shopify_app/sessions'
     ShopifyApp::SessionRepository.storage = InMemorySessionStore
     ShopifyApp.configuration = nil
   end
 
   test "login routes to sessions#new" do
-    assert_routing '/login', { controller: "sessions", action: "new" }
+    assert_routing '/login', { controller: @controller, action: "new" }
```
gmalette: I don't think this warrants a state ivar...

test/routes/webhooks_routes_test.rb
```diff
@@ -3,12 +3,13 @@
...
   setup do
     @routes = ShopifyApp::Engine.routes
+    @controller = 'shopify_app/webhooks'
   end
 
   test "webhooks routing" do
     assert_routing(
       { method: 'post', path: 'webhooks/test' },
-      { controller: 'webhooks', action: 'receive', type: 'test' }
+      { controller: @controller, action: 'receive', type: 'test' }
```
gmalette: same here

lib/shopify_app/sessions.rb
kevinhughes27: +1 to SessionConcern, didn't really like my name

kevinhughes27: I'll keep it out of this PR but given that v7 already has a bunch of name changes I'll put up another PR to rename LoginProtection to AuthConcern to keep things a bit consistent

gmalette:
:+1:

PR #224, opened 2016-03-01
Update quickstart.md

gregersen12:
Missing the part with changing URI for omniauth

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -59,10 +59,12 @@ use the keys from your app in the partners area
 rails generate shopify_app -api_key=a366cbafaccebd2f615aebdfc932fa1c -secret=8750306a895b3dbc7f4136c2ae2ea293 -redirect_uri=https://<name>.herokuapp.com/auth/shopify/callback
 git add .
 git commit -m 'generated shopify app'
-
-If you forget to set your keys or redirect uri above you will find them in the shopify_app initializer at: /config/initializers/.
-In light of this section we recommend adding a gem or utilizing ENV variables to handle your keys.
 ```
+
+If you forget to set your keys or redirect uri above you will find them in the shopify_app initializer at: `/config/initializers/.`
+In light of this section we recommend adding a gem or utilizing ENV variables to handle your keys.
+* Handle your keys before releasing your app!
+
 
 6. Deploy
 ---------


julionc:
:+1: maybe, we can use env variable too.

QUICKSTART.md
```diff
@@ -61,13 +61,20 @@ git add .
...
 
-6. Deploy
+6. Changing redirect URI in Omniauth
+------------------------------------
+```
+Go to your omniauth config file found at /config/initializers/omniauth and make sure to set the redirect URI.
+Set it to your APPNAME.herokuapp.com/auth/shopify/callback adress, as it will default to localhost:3000 on app creation.
+```
+
+7. Deploy
```
kevinhughes27: this shouldn't be required - the redirect_uri is set by the Shopify App Initializer using the parameter passed in the previous step.

kevinhughes27:
This example doesn't use the ENV since its an added step. Its pretty clear in the main README that we recommend using ENV for production apps. If you think its worth adding a note here too I am okay with that.

gregersen12:
That's true @kevinhughes27. I realized, however, that the shopify_app gem per default sets the URI to localhost:3000 instead of the url in which the app lives. Now this is rather obvious, but for someone undergoing first-time setup the app will not work untill this is set correctly.

I pointed at the wrong file by the way. I ment to edit the shopify_app initializer.

kevinhughes27:
How about we add a note to step 6 explaining it a bit rather than adding a new step?

gregersen12:
Seems reasonable Kevin :-)

kevinhughes27:
cool! let me know when you've updated and I'll take a look

gregersen12:
Forgive me for being unaware, but how do I edit my pull-request / update the 'code'?

kevinhughes27:
If you push new commits or amend your current commit (preferred) and push it back to github the pull request will update

gregersen12:
I have updated the code, and I realize I wanted to edit this because I was a toon and didn't realize that the redirect-uri is added on creation :P

QUICKSTART.md
```diff
@@ -59,6 +59,9 @@ use the keys from your app in the partners area
 rails generate shopify_app -api_key=a366cbafaccebd2f615aebdfc932fa1c -secret=8750306a895b3dbc7f4136c2ae2ea293 -redirect_uri=https://<name>.herokuapp.com/auth/shopify/callback
 git add .
 git commit -m 'generated shopify app'
+
+If you forget to set your keys or redirect uri above you will find them in the shopify_app initializer at: /config/initializers/.
+In light of this section we recommend adding a gem or utilizing ENV variables to handle your keys.
```
kevinhughes27: nitpick but can you just add `handle your keys before releasing your app.`

kevinhughes27:
small thing I think we should add. This is a nice improvement!

kevinhughes27:
woops this never got merged - can you rebase and I'll get it in!

PR #225, opened 2016-03-08
More robust redirects

nwtn:
Currently there are two types of redirects used in this app, standard Rails `redirect_to` and a custom `fullpage_redirect_to`. There are problems with both of these, which this PR aims to correct:

1. The HTTP response bodies do not contain valid HTML. (In our consuming app we’re validating HTML output and these redirects are getting caught by our validator.)
2. They have insufficient fallbacks:
  * `redirect_to` uses an anchor (`a href…`) as a fallback if the browser doesn’t honour the 302 status and `Location` headers, but does not automatically redirect using any means
 * `fullpage_redirect_to` uses JavaScript to redirect, with no fallback if JavaScript is disabled or broken (e.g. by a malfunctioning plugin, etc.)

This PR updates all redirects so that they fallback gracefully from HTTP 302 status w/ `Location` header (for non-embedded apps) to JavaScript to `<meta http-equiv="refresh"…>` to a link, all using valid HTML in the body.

@kevinhughes27 @djoume


CHANGELOG
```diff
@@ -1,7 +1,9 @@
-Unreleased
+6.5.0
```
kevinhughes27: next version is going to be 7.0.0 since some of the other things that have been merged are breaking

lib/shopify_app/controller.rb
```diff
@@ -3,16 +3,6 @@
...
     include ShopifyApp::LoginProtection
-
-    def fullpage_redirect_to(url)
-      if ShopifyApp.configuration.embedded_app?
-        render inline: %Q(<script type="text/javascript">window.top.location.href = #{url.to_json};</script>)
-      else
-        redirect_to url
-      end
-    end
-
```
kevinhughes27: moving this makes sense to me :+1: 

At this point do we even need this ShopifyApp:Controller module? all it does is include ShopifyApp::LoginProtection

lib/shopify_app/login_protection.rb
```diff
@@ -60,5 +60,52 @@ def login_url(params = {})
...
+
+    def redirect_to_with_fallback(url)
+      url_json = url.to_json
+
+      render inline: %Q(
+        <!DOCTYPE html>
+        <html lang="en">
+          <head>
+            <meta charset="utf-8" />
+            <meta http-equiv="refresh" content="0; url={url_json}>
```
nwtn: missing a `#` here, will add shortly (i'll also probably need to do wrap this in single quotes rather than double).

lib/shopify_app/login_protection.rb
```diff
@@ -60,5 +60,52 @@ def login_url(params = {})
...
+    def fullpage_redirect_to(url)
+      url_json = url.to_json
+
+      if ShopifyApp.configuration.embedded_app?
+        render inline: %Q(
+          <!DOCTYPE html>
+          <html lang="en">
+            <head>
+              <meta charset="utf-8" />
+              <meta http-equiv="refresh" content="0; url={url_json}>
```
nwtn: ditto

lib/shopify_app/controller.rb
nwtn: I have no idea. How good are the tests in here? If I delete this module and tests go green, can we assume everything is fine?

CHANGELOG
nwtn: :+1: will update

lib/shopify_app/controller.rb
kevinhughes27: I think we should remove this - this repo is moving more towards a proper rails engine from the old generator and it provides controllers now. 

What you'll need to do is replace any includes of ShopifyApp::Controller with ShopifyApp::LoginProtection. Also make a note of that in the CHANGELOG since downstream users will need to make this change when they update. Their is already module renames in this major version

nwtn: OK, sounds good.

lib/shopify_app/version.rb
```diff
@@ -1,3 +1,3 @@
 module ShopifyApp
-  VERSION = '6.4.2'
+  VERSION = '6.5.0'
```
kevinhughes27: bump this too!

kevinhughes27:
Thanks for the PR! lgtm

lib/shopify_app/controller.rb
kevinhughes27: I usually run through a tophat of starting a new app for changes to this repo. We need to go through it before v7 ships

nwtn: that makes sense. once the changes are in, i'll ping you and we can both run through a :tophat:.

nwtn:
feedback addressed, :tophat:ed in an existing local app. about to test creating a new one.

test/controllers/sessions_controller_test.rb
```diff
@@ -11,14 +11,14 @@ class SessionsControllerTest < ActionController::TestCase
     ShopifyApp.configuration.embedded_app = true
     auth_url = '/auth/shopify?shop=my-shop.myshopify.com'
     get :new, shop: 'my-shop'
-    assert response.body.match(/window\.top\.location\.href = "#{Regexp.escape(auth_url)}"/)
+    assert_match /window\.top\.location\.href = "#{Regexp.escape(auth_url)}"/, response.body
```
djoume: could you `assert_match` on the `meta http-equiv` as well? I'm worried about the url quoting not working properly ?

djoume:
LGTM

test/controllers/sessions_controller_test.rb
nwtn: sure!

nwtn:
Added a few more assertions per djoume’s request. :tophat:ed up to the install screen, at which point I got an error about not having a proper redirect url set in my config (which I think is expected).

@kevinhughes27 I’ll follow your lead on merge-readiness. Also are there release steps that will need to be followed once this is merged?

kevinhughes27:
can you continue through with setting up a proper redirect_uri and making sure that works please. Then we can merge. I have some more things that need to make it in before I ship v7 but plan to do so very soon

nwtn:
Ok, done. Was able to install and load the products view, and everything seems good.

kevinhughes27:
Okay I just merged a PR to add namespace isolation - you're going to need to rebase again. Sorry!

nwtn:
rebased

CHANGELOG
```diff
@@ -1,9 +1,13 @@
...
+7.0.0
 ----------
+
 * rename WebhooksController module to WebhookVerification
 * added a WebhooksController which automatically delegates to jobs
 * moved all engine controllers under the ShopifyApp namespace
 * rename SessionsController module to SessionsConcern
+* more robust redirects, with valid HTML in HTTP response bodies
+* `ShopifyApp::Controller` has been removed. You’ll need to replace all includes of `ShopifyApp::Controller` with `ShopifyApp::LoginProtection`
+>>>>>>> more robust full_page_redirect_to
```
kevinhughes27: rebase line is still here :trollface:

example/app/controllers/application_controller.rb
```diff
@@ -1,5 +1,5 @@
 class ApplicationController < ActionController::Base
-  include ShopifyApp::Controller
+  include ShopifyApp::LoginProtection
```
kevinhughes27: thanks for updating this too. I have another PR incoming since the example needs an update after the name-spacing.

kevinhughes27:
one comment then :ship:

CHANGELOG
nwtn: oh my god i'm the worst

nwtn:
rebase stupidity fixed, will :ship: on :green_apple:

PR #226, opened 2016-03-09
update example for v7

kevinhughes27:
This updates the example app for the v7 namespace changes (#223 moved all the controllers into a namespace)

for review:
@nwtn @christinejordan

### DIFF
--- a/example/app/controllers/authenticated_controller.rb
+++ b/example/app/controllers/authenticated_controller.rb
@@ -1,5 +1,0 @@
-class AuthenticatedController < ApplicationController
-  before_action :login_again_if_different_shop
-  around_filter :shopify_session
-  layout ShopifyApp.configuration.embedded_app? ? 'embedded_app' : 'application'
-end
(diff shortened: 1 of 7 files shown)


christinejordan:
:+1:

example/app/controllers/authenticated_controller.rb
```diff
@@ -1,5 +0,0 @@
-class AuthenticatedController < ApplicationController
```
nwtn: are we not providing any of this stuff for them anymore, or did it just get moved elsewhere?

nwtn:
other than my question, this looks :+1:

example/app/controllers/authenticated_controller.rb
kevinhughes27: Yeah this app is in the Engine so they can access it already. Developers just need to inherit from this controller for the most part anyways. In a real app you might copy it over if you need to add more methods but it doesn't serve a purpose in the example.

PR #227, opened 2016-03-09
add add_webhook generator

kevinhughes27:
This PR adds a generator to accompany the new code for handling webhooks added in #217.

#217 added a fairly specific and Rails style way of handling webhooks and given the convention it added I think it makes sense to add a generator to add new webhooks to your app. Its a simple generator and the code can be added manually if the docs are read and understood.

The generator:
- Adds a new webhook to your ShopifyApp config
- Adds a new webhook_job using the delegation name convention and adds the jobs perform method

For review:
@Hammadk @nwtn @gavinballard

### DIFF
--- a/lib/generators/shopify_app/add_webhook/add_webhook_generator.rb
+++ b/lib/generators/shopify_app/add_webhook/add_webhook_generator.rb
@@ -1,0 +1,61 @@
+require 'rails/generators/base'
+
+module ShopifyApp
+  module Generators
+    class AddWebhookGenerator < Rails::Generators::Base
+      source_root File.expand_path('../templates', __FILE__)
+
+      class_option :topic, type: :string, aliases: "-t", required: true
+      class_option :address, type: :string, aliases: "-a", required: true
+
+      def init_webhook_config
+        initializer = load_initializer
+        return if initializer.include?("config.webhooks")
+
+        inject_into_file(
+          'config/initializers/shopify_app.rb',
+          "  config.webhooks = [\n  ]\n",
+          before: 'end'
+        )
+      end
+
+      def inject_webhook_to_shopify_app_initializer
+        inject_into_file(
+          'config/initializers/shopify_app.rb',
+          webhook_config,
+          after: "config.webhooks = ["
+        )
+
+        initializer = load_initializer
+
+        unless initializer.include?(webhook_config)
+          shell.say "Error adding webhook to config. Add this line manually: #{webhook_config}", :red
+        end
+      end
+
+      def add_webhook_job
+        @job_file_name = address.split('/').last + '_job'
+        @job_class_name  = @job_file_name.classify
+        template 'webhook_job.rb', "app/jobs/#{@job_file_name}.rb"
+      end
+
+      private
+
+      def load_initializer
+        File.read(File.join(destination_root, 'config/initializers/shopify_app.rb'))
+      end
+
+      def webhook_config
+        "\n    {topic: '#{topic}', address: '#{address}', format: 'json'},"
+      end
+
+      def topic
+        options['topic']
+      end
+
+      def address
+        options['address']
+      end
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


test/generators/routes_generator_test.rb
```diff
@@ -8,7 +8,6 @@ class ControllerGeneratorTest < Rails::Generators::TestCase
   setup do
     prepare_destination
     provide_existing_routes_file
-    provide_existing_initializer_file
```
kevinhughes27: this should have been removed when the routes flag was dropped from the config as its not needed in this test anymore

test/support/generator_test_helpers.rb
```diff
@@ -17,13 +17,24 @@ def provide_existing_initializer_file
...
   end
 
+  def provide_existing_initializer_file_with_webhooks
+    copy_to_generator_root("config/initializers", "shopify_app_with_webhooks.rb", rename: 'shopify_app.rb')
+  end
+
   private
 
-  def copy_to_generator_root(destination, template)
+  def copy_to_generator_root(destination, template, rename: nil)
```
kevinhughes27: this rename flag lets us have a couple of templates for a file that will be modified by the generator.

lib/generators/shopify_app/add_webhook/add_webhook_generator.rb
```diff
@@ -0,0 +1,55 @@
...
+  module Generators
+    class AddWebhookGenerator < Rails::Generators::Base
+      source_root File.expand_path('../templates', __FILE__)
+
+      class_option :topic, type: :string, aliases: "-t", required: true
+      class_option :address, type: :string, aliases: "-a", required: true
+
+      def init_webhook_config
+        initializer = load_initializer
+        return if initializer.include?("config.webhooks = [")
```
Hammadk: I wonder if there is better way to check if we have already configured webhooks. Hardcoding `"config.webhooks = ["` feels really fragile to me. This could break if the config code doesn't following standard ruby guidelines: `config.webhook=[`, `config.webhook= [` or `config.webhook = Array.wrap()`.

lib/generators/shopify_app/add_webhook/add_webhook_generator.rb
```diff
@@ -0,0 +1,55 @@
...
+
+        inject_into_file(
+          'config/initializers/shopify_app.rb',
+          "config.webhooks = [\n]",
```
Hammadk: Do we need a new line in the `config.webhooks array`? Can we do `config.webhook = []` ?

nwtn:
looks ok to me, but this is pretty far outside my wheelhouse

lib/generators/shopify_app/add_webhook/add_webhook_generator.rb
kevinhughes27: I agree it is a bit fragile but I didn't want to go too far down the rabbit hole for edge cases in the way people define their config.

Maybe what we should do is make this first check a bit less specific i.e. only check for `config.webhooks` here and then warn if we can't find where to inject in the next function.

How does that sound?

lib/generators/shopify_app/add_webhook/add_webhook_generator.rb
kevinhughes27: makes it easier to do the next step. Plus it makes sense to break this into multiple lines or else its super hard to read your config

lib/generators/shopify_app/add_webhook/add_webhook_generator.rb
Hammadk: That's a good idea. Ideally `load_initializer` (or a sister method) could load the config as a hash, and check if there is anything present for `config.webhooks`. Your call :)

kevinhughes27:
Updated per @Hammadk's suggestion. Injecting a new webhook config is still a bit fragile but the generators intentions should be clear and the generator's output will indicate if we couldn't inject and how to respond:

image

Hammadk:
LGTM

PR #228, opened 2016-03-09
update install generator to use rails generator arguments properly

kevinhughes27:
This PR cleans up the way the install generator receives arguments to use the code built into Rails. The install generator will also match the new add_webhook (#227) generator in terms of args now. I also cleaned up the tests and they are now easier to read and better test the generator since the args are passed in correctly.

This is sort of a breaking change since it affects how to use the generators but most users only do this once and we are already about to ship a major version.

@Hammadk @nwtn

### DIFF
--- a/lib/generators/shopify_app/install/install_generator.rb
+++ b/lib/generators/shopify_app/install/install_generator.rb
@@ -6,26 +6,19 @@ module ShopifyApp
     class InstallGenerator < Rails::Generators::Base
       include Rails::Generators::Migration
       source_root File.expand_path('../templates', __FILE__)
-      attr_reader :opts
 
-      def initialize(args, *options)
-        @opts = Hash[options.first.join(' ').scan(/--?([^=\s]+)(?:=(\S+))?/)]
-        @opts = @opts.with_indifferent_access
-        @opts.reverse_merge!(defaults)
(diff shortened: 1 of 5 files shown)


nwtn:
lgtm

lib/generators/shopify_app/install/install_generator.rb
```diff
@@ -6,26 +6,19 @@ module Generators
...
-        }
-      end
+      class_option :api_key, type: :string, default: '<api_key>'
+      class_option :secret, type: :string, default: '<secret>'
+      class_option :redirect_uri, type: :string, default: 'http://localhost:3000/auth/shopify/callback'
+      class_option :scope, type: :string, default: 'read_orders, read_products'
+      class_option :embedded, type: :string, default: 'true'
 
       def create_shopify_app_initializer
+        @api_key = options['api_key']
```
Hammadk: Where is `options` defined?

Hammadk:
LGTM other than the question.

lib/generators/shopify_app/install/install_generator.rb
kevinhughes27: its part of the base generator class - the arguments defined in the class end up here

PR #233, opened 2016-03-14
remove the redirect uri param

### LINKED ISSUE #203: redirect_uri not used correctly
I'm "merging" shopify_app into our existing Rails project.

To avoid routing conflicts, I'm using the following routes:
```ruby
scope :modules do
    scope :shopify do
      controller :shopify_sessions do
        get 'login' => :new, :as => :login
        post 'login' => :create, :as => :authenticate
        get 'auth/shopify/callback' => :callback
        get 'logout' => :destroy, :as => :logout
      end
    end
  end
```

This means that the login pages is accessible via `localhost:3000/modules/shopify/login`, and the callback from `localhost:3000/modules/shopify/auth/shopify/callback`

On the parther dashboard, the URL's are changed accordingly.

The problem is that when installing the app, OmniAuth (at least I think this is the module doing it), will not add the correct direct_uri to the URL. 

When installing, it will redirect to:
`https://mytestshop.myshopify.com/admin/oauth/authorize?client_id=<trimmed>&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fauth%2Fshopify%2Fcallback&response_type=code&scope=read_customers%2C+read_orders%2C+write_orders%2C+write_fulfillments&state=<trimmed>`

(decoded this becomes `http://localhost:3000/auth/shopify/callback`)

Shopify detects that this doesn't match with the URI specified in the parther dashboard, so it will give an error: Oauth error invalid_request: The redirect_uri is missing or not whitelisted

**omniauth.rb**
```ruby
Rails.application.config.middleware.use OmniAuth::Builder do
  provider :shopify,

    :scope => ShopifyApp.configuration.scope,
    :redirect_uri => ShopifyApp.configuration.redirect_uri,

    :setup => lambda {|env|
       params = Rack::Utils.parse_query(env['QUERY_STRING'])
       site_url = "https://#{params['shop']}"
       env['omniauth.strategy'].options[:client_options][:site] = site_url
    }
end
```

**shopify_app.rb**
```ruby
  config.api_key = "mykey"
  config.secret = "mysecret"
  config.embedded_app = true
  config.scope = 'read_customers, read_orders, write_orders, write_fulfillments'
  config.redirect_uri = 'http://localhost:3000/modules/shopify/auth/shopify/callback'
end

```

I cannot tell if it is a bug, or a simple misconfiguration on my part.

kevinhughes27:
closes #203

The `redirect_uri` param was added in https://github.com/Shopify/shopify_app/pull/152 since Shopify required this param as of Shopify/shopify#46095. Omniauth was originally not sending this parameter correctly but it does now without us needing to set it (I didn't actually verify this addition but it does get sent now so I don't think finding the exact change in omniauth is required, also it might have been this Shopify/omniauth-shopify-oauth2/pull/32 that fixed it).

The inclusion of the `redirect_uri` parameter seemed to add to confusion for those mounting the engine at different locations (aka not `/`). I think not having it at all will be better since then anyone submounting will need to learn all the ins and outs of omniauth routing and won't be misslead by our parameter.

This was causing confusion for other developers and myself, for example this issue https://github.com/Shopify/shopify_app/issues/165

I :tophat:ed locally and my install worked fine with this new code. I would appreciate if someone else would tophat as well. I have a script that speeds up the process if you're into that https://github.com/kevinhughes27/dotfiles/blob/master/bash/aliases#L69

For review:
@dylanahsmith @clayton-shopify @EiNSTeiN- 
cc @ShayneP @joshubrown

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@
 * `ShopifyApp::Controller` has been removed. You’ll need to replace all includes of `ShopifyApp::Controller` with `ShopifyApp::LoginProtection`
 * adds add_webhook generator to make it easier to add new webhooks to your app
 * update the install generator to use standard rails generate arguments, usage has changed from `-api_key=your_key` to `--api_key your_key`
+* remove the redirect uri - this is done automatically inside omniauth now
 
 6.4.2
 -----
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -56,7 +56,7 @@ bundle install
 -------------------------------
 ```
 use the keys from your app in the partners area
-rails generate shopify_app -api_key=a366cbafaccebd2f615aebdfc932fa1c -secret=8750306a895b3dbc7f4136c2ae2ea293 -redirect_uri=https://<name>.herokuapp.com/auth/shopify/callback
+rails generate shopify_app --api_key a366cbafaccebd2f615aebdfc932fa1c --secret 8750306a895b3dbc7f4136c2ae2ea293
 git add .
 git commit -m 'generated shopify app'
 ```
--- a/README.md
+++ b/README.md
@@ -85,7 +85,6 @@ $ rails generate shopify_app:install --api_key <your_api_key> --secret <your_app
 Other options include:
 * `scope` - the Oauth access scope required for your app, eg 'read_products, write_orders'. For more information read the [docs](http://docs.shopify.com/api/tutorials/oauth)
 * `embedded` - the default is to generate an [embedded app](http://docs.shopify.com/embedded-app-sdk), if you want a legacy non-embedded app then set this to false, `--embedded false`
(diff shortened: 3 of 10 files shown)


clayton-shopify:
Is this meant to be an API-breaking change? I just tried upgrading an existing app and got:
```
config/initializers/shopify_app.rb:4:in `block in <top (required)>': undefined method `redirect_uri=' for #<ShopifyApp::Configuration:0x007f87933107f0> (NoMethodError)
```

clayton-shopify:
Tested with a new non-embedded app and it worked fine there.

dylanahsmith:
@clayton-shopify the change was added to the changelog for 7.0.0 which hasn't been released, so breaking changes should be fine.

kevinhughes27:
Yeah there is already a few breaking changes for v7.

clayton-shopify:
OK thanks, just wanted to be sure.

EiNSTeiN-:
LGTM

dylanahsmith:
LGTM

PR #239, opened 2016-03-22
Add translation for flash messages

albiere:
It adds translations for "Logged In" in English and Spanish.

Could you guys have a look at it? @sbfaulkner @mattvleming

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -1,4 +1,4 @@
-es: &shopify_app_es
+es:
   logged_in: 'Has iniciado sesión'
   logged_out: 'Cerrar sesión'
   could_not_log_in: 'No se pudo iniciar sesión en tu tienda de Shopify'
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -1,2 +1,4 @@
 es:
   logged_in: 'Has iniciado sesión'
+  logged_out: 'Cerrado sesión'
+  could_not_log_in: 'No se pudo iniciar sesión'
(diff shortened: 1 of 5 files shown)


mattvleming:
Beautiful. The version number should probs be updated. Also just checking, do we have any more translations to add?

sbfaulkner:
LGTM... but, we should probabaly get the other strings in the concern translated too

albiere:
Just added some extra keys to our locales files.

config/locales/es.yml
```diff
@@ -0,0 +1,4 @@
+es:
+  logged_in: 'Has iniciado sesión'
+  logged_out: 'Cerrado sesión'
```
julionc: Spanish native speaker here. I'd recommend to edit the following strings to:
* logged_out: `Su sesión ha finalizado correctamente` or `Cerrando sesión` (Although it sounds a bit weird)
* could_not_log_in: `No fue posible iniciar sesión en la tienda`

hcoleman:
@julionc 

How about:
could_not_log_in: "No se pudo iniciar sesión en tu tienda de Shopify"
logged_out: Cerrar sesión

cc: @albiere

kevinhughes27:
Code looks good to me and thanks for adding more tests - curious why the translation though since Shopify itself is not translated.

mattvleming:
We have a Rails app that uses shopify_app and is being translated for Spanish users. Feel free to DM Angelo or myself if you want to know which one (since this discussion is public).

kevinhughes27:
Okay awesome! That works for me I just wasn't sure if this was being done for a reason or not. :shipit:

mattvleming:
LGTM

Issue #230, opened 2016-03-12
Omniauth error


joshreeves:
I've set bundler to use the latest code from the repository, but now when I attempt to install or view the app I see a flash of a page with a single HTML link `Continue` which displays for a few seconds and then the entire browser refreshes and displays this Rails error: `OmniAuth::Strategies::OAuth2::CallbackError`.

I did not have this problem when using the latest gem. It seems to install properly and I have sometimes seen a flash of content coming from the API, but the error seems to always happen. Any help would be greatly appreciated.

Here is my `initializers/omniauth.rb` file:
```
Rails.application.config.middleware.use OmniAuth::Builder do
  provider :shopify,
    ShopifyApp.configuration.api_key,
    ShopifyApp.configuration.secret,
    redirect_uri: ShopifyApp.configuration.redirect_uri,
    scope: ShopifyApp.configuration.scope
end
```
And my `initializers/shopify_app.rb`:
```
ShopifyApp.configure do |config|
  config.api_key = ENV['SHOPIFY_CLIENT_API_KEY']
  config.secret = ENV['SHOPIFY_CLIENT_API_SECRET']
  config.redirect_uri = "[domain]/auth/shopify/callback"
  config.scope = "read_content, write_content, read_themes, write_themes, read_products, write_products, read_orders, write_orders, read_customers, write_customers, read_script_tags, write_script_tags"
  config.embedded_app = true
  config.webhooks = [
    {topic: 'products/delete', address: 'http://requestb.in/17y2ze91', format: 'json'},
  ]
end
```

kevinhughes27:
What is your system and browser info and is this an embedded app? @nwtn could be related to your changes.

@joshreeves we're going to be QAing the new version before shipping next week. I'd recommend using a released version until then.

nwtn:
Yup sounds like it's related to my change. What app is it? If possible I'd love to test it with the problematic app itself. +1 to using a released version.

joshreeves:
Embedded app in Latest Chrome on Mac OSX. It's an unpublished development app. If you need any more info let me know. I'll likely switch back to the gem. Thanks for the quick response!

kevinhughes27:
hmm okay I am also using latest chrome on osx.

I just ran through a tests creating and installing a new app using - localhost, forward and ngrok. All worked fine for me. I thought your scope might be an issue since there was some recent changes there but I tried with your scope and couldn't reproduce.

[...]

PR #243, opened 2016-03-30
Fixes CSRF OmniAuth issue

alanhill:
@nwtn @kevinhughes27 I ran into the same issue as mentioned here https://github.com/Shopify/shopify_app/issues/230?_pjax=%23js-repo-pjax-container. 

We're making 2 authentication requests and (I think) the last one is beating the first one, causing the token to be invalid. There's some mention of it in this issue here https://github.com/intridea/omniauth-oauth2/issues/58#issuecomment-55061098. 

The error I get through Heroku logs is `Authentication failure! csrf_detected: OmniAuth::Strategies::OAuth2::CallbackError, csrf_detected | CSRF detected`

I changed the refresh to 3 and had no issues, then changed it back to 0 and ran into the same issue. I then decreased from there and saw no issues with it at 1. 

I'm not certain this fixes the root cause, but it's seems to work. 

To replicate the broken version (same steps to :tophat: this branch):

1. Create a new app
2. Successfully install it on your shop
3. The app should be loaded and working
4. Close completely out of your browser
5. Re-open the browser and go back to your shop
6. Try to open the app

You should see one of a few rails error screens, all along the lines of OmniAuth. The logs are telling as well, you'll see the 2 requests fire at once. If it doesn't error the first time, try one or two more, but typically first time's the charm.

### DIFF
--- a/lib/shopify_app/login_protection.rb
+++ b/lib/shopify_app/login_protection.rb
@@ -64,7 +64,7 @@ module ShopifyApp
         <html lang="en">
           <head>
             <meta charset="utf-8" />
-            <meta http-equiv="refresh" content="0; url=#{url_json_no_quotes}">
+            <meta http-equiv="refresh" content="1; url=#{url_json_no_quotes}">
             <title>Redirecting…</title>
             <script type="text/javascript">
               window.location.href = #{url_json};
@@ -87,7 +87,7 @@ module ShopifyApp
           <html lang="en">
             <head>
               <meta charset="utf-8" />
-              <meta http-equiv="refresh" content="0; url=#{url_json_no_quotes}">
+              <meta http-equiv="refresh" content="1; url=#{url_json_no_quotes}">
               <base target="_top">
               <title>Redirecting…</title>
               <script type="text/javascript">
--- a/test/controllers/sessions_controller_test.rb
+++ b/test/controllers/sessions_controller_test.rb
@@ -16,7 +16,7 @@ module ShopifyApp
       auth_url = '/auth/shopify?shop=my-shop.myshopify.com'
       get :new, shop: 'my-shop'
       assert_match /window\.top\.location\.href = "#{Regexp.escape(auth_url)}"/, response.body
-      assert_match /meta http-equiv=\"refresh\" content=\"0; url=#{Regexp.escape(auth_url)}\"/, response.body
+      assert_match /meta http-equiv=\"refresh\" content=\"1; url=#{Regexp.escape(auth_url)}\"/, response.body
     end
 
     test "#new should authenticate the shop if the shop param exists non embedded" do
@@ -24,7 +24,7 @@ module ShopifyApp
       auth_url = '/auth/shopify?shop=my-shop.myshopify.com'
       get :new, shop: 'my-shop'
       assert_match /window\.location\.href = "#{Regexp.escape(auth_url)}"/, response.body
-      assert_match /meta http-equiv=\"refresh\" content=\"0; url=#{Regexp.escape(auth_url)}\"/, response.body
+      assert_match /meta http-equiv=\"refresh\" content=\"1; url=#{Regexp.escape(auth_url)}\"/, response.body
     end
 
     test "#new should trust the shop param over the current session" do
@@ -35,7 +35,7 @@ module ShopifyApp
       auth_url = "/auth/shopify?shop=#{new_shop_domain}"
       get :new, shop: new_shop_domain
       assert_match /window\.top\.location\.href = "#{Regexp.escape(auth_url)}"/, response.body
-      assert_match /meta http-equiv=\"refresh\" content=\"0; url=#{Regexp.escape(auth_url)}\"/, response.body
+      assert_match /meta http-equiv=\"refresh\" content=\"1; url=#{Regexp.escape(auth_url)}\"/, response.body
     end
 
     test "#new should render a full-page if the shop param doesn't exist" do
@@ -50,7 +50,7 @@ module ShopifyApp
         auth_url = '/auth/shopify?shop=my-shop.myshopify.com'
         post :create, shop: good_url
         assert_match /window\.top\.location\.href = "#{Regexp.escape(auth_url)}"/, response.body
-        assert_match /meta http-equiv=\"refresh\" content=\"0; url=#{Regexp.escape(auth_url)}\"/, response.body
+        assert_match /meta http-equiv=\"refresh\" content=\"1; url=#{Regexp.escape(auth_url)}\"/, response.body
       end
     end
 
@@ -61,7 +61,7 @@ module ShopifyApp
         auth_url = '/auth/shopify?shop=my-shop.myshopify.io'
         post :create, shop: good_url
         assert_match /window\.top\.location\.href = "#{Regexp.escape(auth_url)}"/, response.body
-        assert_match /meta http-equiv=\"refresh\" content=\"0; url=#{Regexp.escape(auth_url)}\"/, response.body
+        assert_match /meta http-equiv=\"refresh\" content=\"1; url=#{Regexp.escape(auth_url)}\"/, response.body
       end
     end
 


alanhill:
Oo that blew up tests :(

kevinhughes27:
I'm not really familiar with `http-equiv="refresh"` where did you get the idea to change the content number? and what is this number even for?

alanhill:
`content` is the interval for the document to refresh itself, so my theory is that being at 0 as soon as you open the app it's causing 2 authentication requests to fire, resulting in the error. Once you set it to 1, it gives the first authentication request a chance to complete before firing off any additional ones.

kevinhughes27:
That would certainly explain why we've only seen this intermittently

nwtn:
Great find! :+1:

kevinhughes27:
Okay I did some reading on `http-equiv` and this totally makes sense. 

My question now is why `content=1` other than the fact that it was determined empirically? Is this an upper threshold or a mandatory wait? If the browser will refresh as fast as it can and then maybe this should be higher otherwise couldn't this still affect slower computers? Maybe I don't fully understand this yet so please correct me if I am wrong.

Also @joshreeves can you check this out and see if it fixes it for you and report back? Thanks!

nwtn:
the timeout is in seconds, which means this change is probably making the page redirect w/ the javascript snippet instead of the `<meta>` tag.

if you leave the `meta` timeout at `1` and remove the `script` tag, does the problem come back? i'm wondering if it’s the manner of redirection that’s causing the conflict, rather than the timeout length.

kevinhughes27:
Thats makes sense - with the ESDK I think we want the refresh done by javascript

nwtn:
The idea was to have a the `meta` in there as a fallback in case javascript failed for some reason (in this case I was thinking of some user plugins/extensions that can block JS redirects). That said, if the `meta` doesn’t work at all, it’s not much of a fallback.

If this does’t work without the `script` tag, I’d rather remove the `meta` tag entirely rather than chance the timeout.

kevinhughes27:
My understanding is that it does work its just less preferred than the js redirect so I think this idea is a good one. We just might want to relax that number even more

joshreeves:
@kevinhughes27 Using this branch mostly fixes the issues for me. I still see the "Continue" text when reloading the app from an cold browser start, and if I click the "Continue" text it takes me back to the Apps page for the Shop with a flash error message that the app could not be loaded. But if I wait, the app will eventually load full-frame in the browser, and then redirect to the embedded view. Overall much better than what I was experiencing before.

kevinhughes27:
Excellent I think that makes sense. @joshreeves do you have some kind of browser extension that would be blocking javascript redirects?

I am going to merge this in then and if we get more reports we'll go deeper and maybe increase the value  of content but for now I think `1` is good.

joshreeves:
@kevinhughes27 No, running a basic Chrome installation.

kevinhughes27:
Okay thanks for the info.

wielinde:
I believe the issue is not fixed yet. I ran into the same issue with Chrome. It works nicely on Firefox and Safari (all browsers on Mac). I am using `shopify_app` in version `7.0.2`.

- I checked that the installed gem has the `content=1` set. And it is. So I believe I have the lastest version.
- I tried increasing to `content=10` to no effect.

I am not (yet) an expert on Omniauth but if I can help you debugging just let me know what you need.

kevinhughes27:
thanks for the information @wielinde - we've seen it happen a few times as well. We're looking into it.

kevinhughes27:
What is your version of `omniauth-oauth2` in your Gemfile.lock?

If you effectively revert this PR does it fix the problem? https://github.com/Shopify/shopify_app/pull/225

wielinde:
Excellent!

Kevin Hughes <notifications@github.com> schrieb am Do., 14. Apr. 2016,
19:29:

kevinhughes27:
Did either of those suggestions solve your problem?

wielinde:
@kevinhughes27 The Gemfile.lock says `omniauth-oauth2 (1.4.0)`

OK, I tried reverting pull request #225 ... and adopted our Rails app accordingly to no reveal. I get the  same CSRF error.

kevinhughes27:
Can you force omniauth to version 1.3.1 and see if that fixes it?

wielinde:
@kevinhughes27 `omniauth` actually is in version `1.3.1`. It is `omniauth-oauth2` that is in `1.4.0`. Here is the part in my `Gemfile.lock`:

```
    auth2 (1.1.0)
      faraday (>= 0.8, < 0.10)
      jwt (~> 1.0, < 1.5.2)
      multi_json (~> 1.3)
      multi_xml (~> 0.5)
      rack (>= 1.2, < 3)
    omniauth (1.3.1)
      hashie (>= 1.2, < 4)
      rack (>= 1.0, < 3)
    omniauth-facebook (3.0.0)
      omniauth-oauth2 (~> 1.2)
    omniauth-oauth2 (1.4.0)
      oauth2 (~> 1.0)
      omniauth (~> 1.2)
    omniauth-shopify-oauth2 (1.1.14)
      omniauth-oauth2 (~> 1.2)
```

kevinhughes27:
Oops sorry I meant omniauth-oauth2 to 1.3.1

wielinde:
@kevinhughes27 Same error :-/

kevinhughes27:
Can you clone https://github.com/Shopify/omniauth-shopify-oauth2 and set your app to use the local gem or use bundle open to edit the local gem and put a debugger here:

https://github.com/Shopify/omniauth-shopify-oauth2/blob/master/lib/omniauth/strategies/shopify.rb#L114

I am curious what the value of `callback_url` is vs what it calculates on the other side and where your request originated

wielinde:
`options[:callback_url]` is `"myngrokdomainhere/auth/shopify/callback"`

kevinhughes27:
hmm so its missing the protocol. If you patch it to have `https://` does that fix it?

wielinde:
Oops, sorry for not being clear enough. I replaced my domain with the placeholder before posting here and did not mention the protocol. So yes the URL has the the full protocol (https://).

kevinhughes27:
what if you change it to plain `http`

wielinde:
@kevinhughes27 I just changed everything from from `https` to `http`. First I ran into some redirect_uri param missing error. However, when I manually readded that missing param to the URL in the browser window and fired that again, I got the same CSRF error message.

kevinhughes27:
@wielinde any update from your side?

wielinde:
Nope, nothing new. Having trouble to get a minimal example app running,
that reproduces the error. However, I will give it another shot.

alanhill:
I have a 'working' version [here](https://github.com/alanhill/test_shopify_app) of the broken version using an older ref. It makes it stand out more with the `content=0` making it easier to replicate. 

Chrome is the biggest culprit - it's always replicable there. Clone the repo, `rake db:migrate`, go to localhost, install the app on your shop, close out of Chrome, re-open, go to your shop, open the app and you'll get the CSRF issue. 

I haven't been able to reliably replicate on Firefox, though it's happened a few times at least.

kevinhughes27:
Is it the csrf issue that you get still? I made some changes to the Shopify OAuth controller to try and return better error messages. On a production app I operate I am getting `The authorization code was invalid` which also seems to point to a race condition

agerber23:
I had to get rid of the http-equiv completely for the problems to go away. Is that the general consensus?

kevinhughes27:
@agerber23 we had just increased the value but perhaps removing entirely does make sense. @nwtn what do you think?

mhaylock:
@kevinhughes27 @nwtn: while I haven't encountered a CSRF error yet, I have noticed that on Chrome there is a second redirect performed thanks to the `http-equiv="refresh"` metatag.

It appears chrome does the top-level redirect thanks the javascript - our server receives the callback response and then subsequently receives a second authentication request. i.e. our server log looks like:
```
Started GET "/auth/shopify?shop=example.myshopify.com"

Started GET "/auth/shopify/callback?…"

Started GET "/auth/shopify?shop=example.myshopify.com"
```
It's clear though that despite the redirect of the top window, Chrome still honors the `http-equiv="refresh"` metatag before cleaning up the iframe. I also tried 2 and 3 second delays on the tag and only at 3 seconds did the second redirect stop occurring on my machine.

This is running our app server on my machine in development mode, so response times for it are very fast (< 30ms), but here in New Zealand latency to Shopify (or rather, TTFB - time to first byte, on the authorize request involved here) is about 350ms.

Based on my experience I would certainly encourage you to consider removing the meta tag. How many user agents are going to fail to execute the javascript? Surely it's a very small percentage - and in that case can we not have the more typical HTML fallback with a link ("If you are not redirected, please follow this link") sort of thing? We could possibly have the javascript strip the fallback HTML as soon as it executes, if there is a concern about the fallback HTML appearing to the user.

I'm happy to build a PR with such changes if you are interested in moving in that direction.

mhaylock:
@kevinhughes27 @nwtn: actually I tested a bit further and found this works to avoid the double-redirect, but will still fallback to the meta-tag if the javascript fails to execute:
```ruby
    def fullpage_redirect_to(url)
      url_json = url.to_json
      url_json_no_quotes = url_json.gsub(/\A"|"\Z/, '')

      if ShopifyApp.configuration.embedded_app?
        render inline: %(
          <!DOCTYPE html>
          <html lang="en">
            <head>
              <meta charset="utf-8">
              <base target="_top">
              <title>Redirecting…</title>
              <script type="text/javascript">
                window.top.location.href = #{url_json};
                while(document && document.lastChild) {
                  document.removeChild(document.lastChild);
                }
              </script>
              <meta http-equiv="refresh" content="1; url=#{url_json_no_quotes}">
            </head>
            <body>
            </body>
          </html>
        )
      else
        redirect_to_with_fallback url
      end
    end
```
I've only tested in Chrome so far, but here the javascript is executed first and strips the document of all it's contents before the meta tag is parsed and obeyed. This obviously also wouldn't be appropriate if the extensions you speak of will block the URL change but still allow the javascript to execute.

kevinhughes27:
Thats a pretty nifty solution. And very nice digging into the issue.

I actually think this whole thing is not required since we require javascript for the ESDK so why put all this effort into our redirect page. I like your solution but I think we should ultimately just remove the http-equiv tag all together.

Do you want to PR that up?

batasrki:
Further discussion with @kevinhughes27 and @Hammadk in the messenger-channel Slack seems to bring about the removal of `http-equiv`. If the majority of people on the issue are OK with that, I'll take on the PR.

mhaylock:
Sorry for not replying earlier @kevinhughes27, thanks @batasrki - sound good to me, go for it!

nwtn:
yes kill it, it’s brought way more trouble than it could ever possibly be worth

PR #246, opened 2016-04-10
Allow creation of independant webhooks through the WebhookManager

christianblais:
Often apps need to register webhooks after the installation process.
Whether the app adds a new feature, or customer requests access to
a restricted feature, not all required webhooks are known right from
the start.

With this patch, it is now possible to use the manager as follow;

```ruby
@manager.create_webhook(topic: 'topic', address: 'http://example.com')
```

@kevinhughes27 @alanhill what do you think?

### DIFF
--- a/lib/shopify_app/webhooks_manager.rb
+++ b/lib/shopify_app/webhooks_manager.rb
@@ -18,10 +18,18 @@ module ShopifyApp
     def create_webhooks
       return unless required_webhooks.present?
 
+      required_webhooks.each do |webhook|
+        create_webhook(webhook)
+      end
+    end
+
+    def create_webhook(attributes)
       with_shopify_session do
-        required_webhooks.each do |webhook|
-          create_webhook(webhook) unless webhook_exists?(webhook[:topic])
-        end
+        return webhook_for(attributes[:topic]) if webhook_exists?(attributes[:topic])
+        attributes.reverse_merge!(format: 'json')
+        webhook = ShopifyAPI::Webhook.create(attributes)
+        raise CreationFailed unless webhook.persisted?
+        webhook
       end
     end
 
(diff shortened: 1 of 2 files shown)


christianblais:
I was expecting Ruby 2.3, not 2.0, hence the failure. I'll refactor to be 2.0 compatible.

christianblais:
Now ruby 2.0 compatible.

kevinhughes27:
👍 

Should we update the script tags manager as well?

christianblais:
The more I think about this, the more I'm unsure about that solution. Since webhooks (or scripttags) both require HTTP calls, they'll most likely be executed in the context of a job. All shop specific jobs usually already wrap their entire `perform` method with a `with_shopify_session`, so I'm wondering if it makes sense to now rely on a WebhooksManager, which also needs to be instantiated with a shop domain and shop token because it too takes care of the session. So I guess my question is, does it make sense for the WebhooksManager to take care of the shopify session, or should it be left to the caller? Or maybe `WebhooksManager.create_webhook` should be a class method, which itself would need to be wrapped in a shopify session by either the job or, in the context of WebhooksManager, the call to `create_webhooks`. Thoughts?

kevinhughes27:
yeah that is always a tough call. The con of having the webhook manager setup the shopify session is that a user can make API calls without realizing it. The con of not having it do so is that the user is responsible for this. We could improve the dev experience by raising a specific and helpful error message  from inside the webhook manager if there is no Shopify session.

That might be the best solution what do you think?

alanhill:
It seems like for the most part this will be initiated by a user action (adding a new feature, upgrading a plan) then run in a background job. You already expect to wrap your `perform` method with a `with_shopify_session` so I think it makes sense to leave it to the caller in this case. 

It's a tough one because it makes the dev experience a bit easier by having the webhook manager setup the session, but it seems inconsistent with what you already expect to do.

christianblais:
I'll close this then, as I think we're better off leaving the session to the caller. I'll create another PR just to explore that idea.

PR #247, opened 2016-04-10
Bump ruby version to 2.3.0

christianblais:
As of February 24th, 2016;

> **About Ruby 2.0.0**
> As it has been announced before, all support for Ruby 2.0.0 has ended today. Bug and security fixes from more recent Ruby versions will no longer be backported to 2.0.0, and no further patch release of 2.0.0 will be released.
https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/

The current stable version is 2.3.0. While it introduces a lot of breaking changes (keyword arguments, frozen literals, the lonely operator, etc), I believe now is a good time to make the switch and avoid running on a deprecated version for too long. If you believe this is too much of a breaking change, I'd suggest to at least bump to 2.2.4. But I strongly believe a bump should happen.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: ruby
 
 rvm:
-  - 2.0.0
+  - 2.2.4
 
 gemfile:
   - Gemfile


kevinhughes27:
I am onboard for this.

christianblais:
I was thinking about this over the week-end and was waiting on your answer before replying, but I think that might be a little too aggressive. If we see that travis file as what ruby version we want to support, going to 2.2.4 might be a better move than right away going to 2.3.

kevinhughes27:
We can be a bit more conservative as well. Im not sure how common upgrading ShopifyApp is for most users which is why I am okay with pushing forward aggressively

christianblais:
I wasn't able to find any statistics regarding ruby 2.3 usage, but 2.2.X seems to have a fairly wide adoption, so I'd be comfortable bumping to 2.2.4.

I'm ready to merge whenever you are!

kevinhughes27:
:shipit:

PR #248, opened 2016-04-12
Managers are now shopify session agnostic

christianblais:
Here's the solution we discussed earlier. All managers would now be session agnostic, which gives to the caller the responsibility to instantiate it. With that being done, we'll be able to expose methods such as `create_webhook` and `create_scriptag` as independent methods, without the fear to interfere with an already existing session.  

@kevinhughes27 @alanhill

### DIFF
--- a/lib/shopify_app/scripttags_manager.rb
+++ b/lib/shopify_app/scripttags_manager.rb
@@ -2,12 +2,8 @@ module ShopifyApp
   class ScripttagsManager
     class CreationFailed < StandardError; end
 
-    def self.queue(shop_name, token)
-      ShopifyApp::ScripttagsManagerJob.perform_later(shop_name: shop_name, token: token)
-    end
-
-    def initialize(shop_name, token)
-      @shop_name, @token = shop_name, token
+    def self.queue(shop_domain, shop_token)
+      ShopifyApp::ScripttagsManagerJob.perform_later(shop_domain: shop_domain, shop_token: shop_token)
     end
 
     def recreate_scripttags!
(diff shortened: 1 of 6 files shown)


lib/shopify_app/scripttags_manager.rb
```diff
@@ -2,12 +2,8 @@ module ShopifyApp
...
 
-    def self.queue(shop_name, token)
-      ShopifyApp::ScripttagsManagerJob.perform_later(shop_name: shop_name, token: token)
-    end
-
-    def initialize(shop_name, token)
-      @shop_name, @token = shop_name, token
+    def self.queue(shop_domain)
+      ShopifyApp::ScripttagsManagerJob.perform_later(shop_domain: shop_domain)
     end
```
kevinhughes27: I like making this object not responsible for creating the shopify session if used stand alone but I think we should still pass the token in here. Otherwise we are assuming the existence of the Shop model which may not be true. We could also use the SessionRepository object inside the job but I find that less clear than just passing in the token and letting the job create an API session.

lib/shopify_app/scripttags_manager.rb
```diff
@@ -44,12 +37,6 @@ def is_required_scripttag?(scripttag)
...
 
-    def with_shopify_session
-      ShopifyAPI::Session.temp(@shop_name, @token) do
-        yield
-      end
-    end
-
     def create_scripttag(attributes)
       attributes.reverse_merge!(format: 'json')
       scripttag = ShopifyAPI::ScriptTag.create(attributes)
```
kevinhughes27: In the PR where we expose this method we should add specific errors that are raised when there is no ShopifyAPI session set.

lib/shopify_app/scripttags_manager.rb
christianblais: Fair enough, I was assuming a shop object, but you're right, it might not always be the case. Will fix.

lib/shopify_app/scripttags_manager.rb
christianblais: Absolutely.

christianblais:
Ready for re-review.

kevinhughes27:
LGTM

PR #253, opened 2016-05-04
Depend on shopify_api 4.2.x

EiNSTeiN-:
Bump dependency on shopify_api to 4.2.x. The current version 4.0.2 will be a year old tomorrow...

@kevinhughes27 @christianblais

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,8 @@
+7.0.4
+-----
+
+* Bump required shopify_api version to 4.x.
+
 7.0.3
 -----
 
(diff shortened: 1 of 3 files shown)


kevinhughes27:
👍 

The version was only added due to a security issue. I think we had a floating a dep before.

EiNSTeiN-:
Should I change to `~> 4.2` ?

CHANGELOG
```diff
@@ -1,3 +1,8 @@
+7.0.4
+-----
+
+* Bump required shopify_api version to >= 4.2.0.
```
alexcoco: `>=` is not correct here since the constraint is `~>`

EiNSTeiN-: I changed this to `4.x` so we don't end up stuck with 4.2.x after 4.3 comes out.

alexcoco:
👍

christianblais:
👍

PR #259, opened 2016-05-26
updating shop migration to use new rails pattern and adding limits and null falses

celsodantas:
@kevinhughes27  @pseudomuto 

Adding `limit` and `null: false` for the timestamp and updating to use Rails `change` method for migration.

---
This change is [Reviewable](https://reviewable.io/reviews/shopify/shopify_app/259)


kevinhughes27:
👍

pseudomuto:
Nice! :rocket:

PR #260, opened 2016-05-30
Make it possible to override shopify_token validation

cyprusad:
I have run into a situation where I wanted to create Shop records that didn't have `shopify_token` created  just yet. Turns out I cannot save a record without them, because the `validates presence: true` condition needs to be satisfied.

This change will check if token **needs** to exist by exposing a class method.

Please review: @kevinhughes27 @j-mutter

### DIFF
--- a/lib/shopify_app/shop.rb
+++ b/lib/shopify_app/shop.rb
@@ -4,11 +4,19 @@ module ShopifyApp
 
     included do
       validates :shopify_domain, presence: true, uniqueness: true
-      validates :shopify_token, presence: true
+      if shopify_token_needs_to_exist?
+        validates :shopify_token, presence: true
+      end
     end
 
     def with_shopify_session(&block)
       ShopifyAPI::Session.temp(shopify_domain, shopify_token, &block)
+    end
+
+    module ClassMethods
+      def shopify_token_needs_to_exist?
+        true
+      end
     end
 
   end


lib/shopify_app/shop.rb
```diff
@@ -4,12 +4,20 @@ module Shop
 
     included do
       validates :shopify_domain, presence: true, uniqueness: true
-      validates :shopify_token, presence: true
+      if shopify_token_needs_to_exist?
```
j-mutter: Any reason to not just use
```ruby
validates :shopify_token, presence: true, if: :shopify_token_needs_to_exist?
```

also, `shopify_token_required?` feels less awkward to me, but naming things is hard 😇

lib/shopify_app/shop.rb
```diff
@@ -4,12 +4,20 @@ module Shop
...
 
     def with_shopify_session(&block)
       ShopifyAPI::Session.temp(shopify_domain, shopify_token, &block)
     end
 
+    module ClassMethods
+      def shopify_token_needs_to_exist?
+        true
+      end
+    end
```
kevinhughes27: This code looks fine but my question for you guys is this: How common is this use case and it is even possible for non-internal apps? If this isn't useful to the app community at large then I would prefer to not accept this PR and for you to just have your own shop model concern in your app.

cyprusad:
Ehh.. This is a bit heavy-handed. I think I can work around it..

Issue #264, opened 2016-06-01
Webhook Manager Topic Whitelist


kevinhughes27:
It would be easy to have a constant of all the valid webhook topics and validate against them. Its pretty easy to type the topic or forget the tense of the topic.

adfaris:
Is this still an issue? It looks like #266 was a good attempt that was closed three years later. 

cc @tanema

PR #266, opened 2016-06-06
Add webhook topic whitelist

alanhill:
##### What's it do

Validates the topic of the webhook when generating a webhook via console or in the case of mis-typing directly into the config, when the Webhook attempts to install. (https://github.com/Shopify/shopify_app/issues/264)
##### How

There's a list of constants with all the valid webhook topics. In the case of the generator, it will raise an InvalidTopic error and list out the valid topics. If it fails upon creation on a shop, it raises an invalid topic error, it will log a warning with the valid topics, then fall back to the CreationFailed error like normal when webhook creation fails. 

@kevinhughes27 is this what you had in mind? I couldn't think of a good way to DRY the list of constants, all ears if you have ideas.


alanhill:
CI passed but I'm getting an error locally, seems to be unrelated though.

I pulled master and got the same error, just trying to run `bundle exec rake test`.

```
master_shopify_app/shopify_app/lib/shopify_app/webhooks_manager_job.rb:3: syntax error, unexpected ','
    def perform(shop_domain:, shop_token:)
```

kevinhughes27:
@alanhill that's because you're running tests locally with an old version of ruby. I'm not sure the best way to force a greater than ruby version

lib/shopify_app/webhooks_manager.rb
```diff
@@ -37,6 +77,12 @@ def is_required_webhook?(webhook)
       required_webhooks.map{ |w| w[:address] }.include? webhook.address
     end
 
+    def is_valid_webhook?(webhook)
+      unless VALID_WEBHOOK_TOPICS.any? { |valid| valid == webhook }
+        Rails.logger.warn "#{webhook} isn't a valid topic. Valid topics include: #{VALID_WEBHOOK_TOPICS}"
+      end
+    end
```
kevinhughes27: we could make this a concern that is included in both classes. Checkout ActiveSupport::Concern it should do the trick.

kevinhughes27:
lets dry this up using a concern otherwise this is exactly what I had in mind! Thanks!

alanhill:
Moved it into a concern.. can't seem to get `Rails.logger.warn` (or `shell.say`) to work though now that it's been moved. 👀  into it..

alanhill:
Also switched to using just CreationFailed in each case, instead of using InvalidTopic for the generator and CreationFailed for the manager.. KISS.

alanhill:
@kevinhughes27 everything up, moved it into a concern, rescues for each use case with it's own way of logging it. 

I was having a lot of trouble with the 1 test ensuring Rails.logger.warn runs - I can replace it if necessary..

lib/webhook_topic_validator.rb
```diff
@@ -0,0 +1,62 @@
+require 'active_support/concern'
+module ShopifyApp
+  module WebhookTopicValidator
+    class InvalidTopic < StandardError; end
+    extend ActiveSupport::Concern
```
kevinhughes27: nitpick but can you swap the order of the two lines and have a space in between them

alanhill:
@christianblais mind reviewing as well

lib/webhook_topic_validator.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    def invalid_topic_shell_message
+      shell.say "#{@topic} is an invalid webhook topic. "\
+      "Valid topics include: #{VALID_WEBHOOK_TOPICS}"
+    end
+
+    def invalid_topic_logger_message
+      Rails.logger.warn "#{@topic} is an invalid webhook topic. "\
+      "Valid topics include: #{VALID_WEBHOOK_TOPICS}"
+    end
```
kevinhughes27: I would just put this code in the rescue blocks above. That way this object knows if its valid and the caller knows what to do when this class returns invalid.

lib/generators/shopify_app/add_webhook/add_webhook_generator.rb
```diff
@@ -3,12 +3,20 @@
...
       class_option :topic, type: :string, aliases: "-t", required: true
       class_option :address, type: :string, aliases: "-a", required: true
 
       def init_webhook_config
+        begin
+          is_valid_topic?(topic)
+        rescue ShopifyApp::WebhookTopicValidator::InvalidTopic => e
+          invalid_topic_shell_message
+          raise
+        end
```
kevinhughes27: rather than using begin you could just use implicit rescue inside this whole function. Then it would look like this:

```ruby
def init_webhook_config
  is_valid_topic?(topic)
  ...
rescue ShopifyApp::WebhookTopicValidator::InvalidTopic => e
  ...
end
```

alanhill:
ping @kevinhughes27 comments fixed up!

lib/webhook_topic_validator.rb
```diff
@@ -0,0 +1,53 @@
...
+                            'products/update',
+                            'refunds/create',
+                            'shop/update',
+                            'themes/publish'
+                        ]
+    def is_valid_topic?(options = {})
+      @topic = options
+      unless VALID_WEBHOOK_TOPICS.any? { |valid| valid == @topic }
+        raise InvalidTopic
+      end
```
kevinhughes27: you could put this on one line

lib/shopify_app/webhooks_manager.rb
```diff
@@ -15,6 +18,13 @@ def create_webhooks
       return unless required_webhooks.present?
 
       required_webhooks.each do |webhook|
+        begin
+          is_valid_topic?(webhook[:topic])
+        rescue ShopifyApp::WebhookTopicValidator::InvalidTopic => e
+          Rails.logger.warn "#{@topic} is an invalid webhook topic. "\
+          "Valid topics include: #{VALID_WEBHOOK_TOPICS}"
+          raise
```
kevinhughes27: totally not required but we could probably do something really cool here with a suggestion. Find the closest matching topic using some string difference and Say did you mean ... rather than printing all the topics.

What do you think?

PR #270, opened 2016-06-10
Add FulfillmentService support

rjocoleman:
This needs a little work and some testing. I haven't looked at auth yet either.

I want to get some feedback please before moving further.

This PR adds support for `FulfillmentServices` in the same way as `Webhooks` and `ScriptTags`.

Key features:
- Support for multiple fulfilment services via namespacing (since the API supports it)
- `ShopifyApp.configuration.base_url` has been added for the `callback_url` base.
-  `ShopifyApp::FulfillmentServicesController` receives the GET request from Shopify and passes this down the the parent app in `./app/lib/custom_name_fulfillment_service.rb`. The `fetch_stock` and `fetch_tracking_numbers` methods here return a hash back for the response.
- `ShopifyApp::FulfillmentServicesManagerJob` to automatically register FulfillmentServices.
- Generator for initializer changes and fulfilment service stub file.

I note that the example at https://github.com/Shopify/shopify-fulfillment-integration automatically registers a `fulfillments/create` webhook. I've left this out since webhooks can be done separately easily if this use-case exists.

I'm happy for any feedback


app/controllers/shopify_app/fulfillment_services_controller.rb
```diff
@@ -0,0 +1,23 @@
+module ShopifyApp
+  class FulfillmentServicesController < ApplicationController
+    class ShopifyApp::MissingFulfillmentServiceError < StandardError; end
+
+    def action_missing(name)
+      render json: fulfillment_service_klass.send(name, fulfilment_service_params)
+    end
```
kevinhughes27: why is this method called action_missing? shouldn't there be a method for each action or a method missing that tries to send it to the right method?

lib/shopify_app/configuration.rb
```diff
@@ -26,6 +28,14 @@ def has_webhooks?
...
       scripttags.present?
     end
+
+    def has_fulfillment_services?
+      fulfillment_services.present?
+    end
+
+    def has_base_url?
+      base_url.present?
+    end
```
kevinhughes27: I have gone down this route of base_url before and I found it got kind of complicated. I kind of prefer to have people set the url explicitly in the initializer

lib/shopify_app/sessions_concern.rb
```diff
@@ -19,6 +19,7 @@ def callback
         session[:shopify] = ShopifyApp::SessionRepository.store(sess)
         session[:shopify_domain] = shop_name
 
+        FulfillmentServicesManager.queue(shop_name, token) if ShopifyApp.configuration.has_fulfillment_services? && ShopifyApp.configuration.has_base_url?
```
kevinhughes27: nitpick but can this one be last? (aka after webhooks and script tags)

kevinhughes27:
Thanks for this! Its been on my ToDo list for a while!

app/controllers/shopify_app/fulfillment_services_controller.rb
rjocoleman: `action_missing` is basically `method_missing` in controllers. I'll change it to be more readable: 

```ruby
    def fetch_stock
      render json: fulfillment_service_klass.send(:fetch_stock, fulfilment_service_params)
    end
```

lib/shopify_app/configuration.rb
rjocoleman: How do you want the interface to look then?

My use-case on heroku like hosting I want to set a `ENV['CURRENT_APP_URL']` and forget about it, so my webhooks look something like this:

```
config.webhooks = [
  {topic: 'orders/create', address: File.join(ENV['CURRENT_APP_URL'], 'webhooks/orders_create'), format: 'json'}
]
```

And I'm doing something similar here: https://github.com/Shopify/shopify_app/pull/270/files#diff-586b77bc2e3ea60fba8a578c84145c02R56

This means the initializer interface doesn't care about the `callback_url`, it's automatically namespaced and set correctly. What it doesn't cover is letting ShopifyApp's parent act only as a means to set the `callback_url` and not respond to it.

How do you want it to look?

lib/shopify_app/sessions_concern.rb
rjocoleman: sure, this is a hangover from me injecting webhooks automatically on fulfilment service creation (hint: it doesn't work because async so the code is gone except this one artefact)

rjocoleman:
@kevinhughes27 I've responded to all feedback.. I'm going to push on with this putting it into my app to see if there are any quirks when it's used in anger.

lib/shopify_app/configuration.rb
kevinhughes27: hmm thinking about this again perhaps we should have the base url option. Can you split that into a new PR though and have this one be explicit? 

If we go this route we should make all the services use this variable and add it to the generator and the readme

kevinhughes27:
Can you add some tests for the controller? should be similar to this: https://github.com/Shopify/shopify_app/blob/master/test/integration/webhooks_controller_test.rb

I am also interested in what the error looks like if you request an unavailable action since I am wondering if we should make the routes explicit

rjocoleman:
I've added the explicit routes, good call. I should have changed that in the last commit.

Testing is on my TODO list but I've got a tight deadline on two store builds that must come first 🎈 🎂

kevinhughes27:
sounds good. Ping me on the PR when its ready. Good luck with your builds

kevinhughes27:
I'd love to see this included for version 8 - do you think you'll have time to finish this up?

jonmchan:
@rjocoleman ping?

Issue #265, opened 2016-06-05
Soften dependency to allow support for Rails 5


andriytyurnikov:
Any chance to get rails 5 support?
```
shopify_app was resolved to 7.0.5, which depends on
      rails (< 5.0, >= 4.2.6) was resolved to 5.0.0.rc1, which depends on
        actionpack (= 4.2.6)
```

kevinhughes27:
I have not tested with rails 5 yet. I don't recall anything that affects engines or generators in the changelog when I read through it so it should be fine. I'd accept a PR that loosens the dep and adds a rails5 build to the test suite.

alanhill:
I just updated the title to reflect that this is to support Rails 5.

kevinhughes27:
@swalkinshaw this can be close eh?

PR #272, opened 2016-06-10
soften rails dependency

andriytyurnikov:


andriytyurnikov:
as discussed in #265

kevinhughes27:
Can we add to travis so we run tests for both rails 4 and 5

andriytyurnikov:
Like this? Both builds run against rails 4.2.x, as rails 5 is not released yet :)

kevinhughes27:
I was thinking more explicit - a gemfile for rails 4.2.6 and one for 5.0.0.rc1. But yeah I think thats how to do it with the travis.yml

andriytyurnikov:
and here we face the need for shopify_api softer rails dependency

andriytyurnikov:
@kevinhughes27 , if you have access to shopify_api, you might be interrested in checking out [https://github.com/Shopify/shopify_api/issues/274](https://github.com/Shopify/shopify_api/issues/274)

kevinhughes27:
I'll have a look next week.

swalkinshaw:
@andriytyurnikov can this be updated to reference Rails 5.0 and not the RC?

Actually I think the default version should just be `< 5.1` now. Then we only need a `Gemfile` for Rails `~> 4.2`. Additionally it would be nice to have a Gemfile for Rails master which is allowed to fail.

@kevinhughes27 thoughts on  `< 5.1` as the default?

kevinhughes27:
Works for me

andriytyurnikov:
@swalkinshaw updated for Rails 5.0.0 - still failing (as expected -- due to shopify_api issue)

shopify_app.gemspec
```diff
@@ -8,9 +8,7 @@ Gem::Specification.new do |s|
   s.author      = "Shopify"
   s.summary     = %q{This gem is used to get quickly started with the Shopify API}
 
-  s.add_dependency('rails', '>= 4.2.6', '< 5.0')
```
swalkinshaw: I think this should remain in the `gemspec` with `< 5.1`

Gemfile
```diff
@@ -1,4 +1,6 @@
 source "https://rubygems.org"
 
+gem 'rails', '< 5.1.0'
```
swalkinshaw: If we keep the default in the `gemspec` as I mentioned below I don't think this is needed in here.

swalkinshaw:
@andriytyurnikov thanks, made some comments. This PR will need to wait on https://github.com/Shopify/shopify_api/pull/240 so that's fine if it's failing for now.

shopify_app.gemspec
andriytyurnikov: honestly, it is hard for me to justify 2 dependecy records for same gem in both Gemfiles and gemspec

Gemfile.rails-5.0.0
```diff
@@ -0,0 +1,6 @@
+source "https://rubygems.org"
+
+gem 'rails', '5.0.0'
```
andriytyurnikov: it sounds like further improvement, initial goal was to have travis matrix for 4.2.6 and 5.0.0, sure many approaches possible like 4.2-stable, 5.0-stable, last-master etc -- feel free to add changes, or using branch as an example for improvements/alternatives

shopify_app.gemspec
swalkinshaw: well only the Gemfiles that override it would have it. My thought is: `gemspec` should always be the main source of dependencies (when possible). The only reason we have other separate Gemfiles is for testing purposes.

Gemfile.rails-5.0.0
swalkinshaw: Sure, that's fine. We can improve it after if you want.

It's just redundant right now because the default gemspec/Gemfile will resolve to `5.0.0` which covers `Gemfile.rails-5.0.0`.

shopify_app.gemspec
andriytyurnikov: @swalkinshaw to be honest I don't know for sure what are overriding rules for gemspec Gemfile combo (if there are any), and to speak freely I was happy to not know such thing, and to not force others to know such things, while reading files. Having same thing configured in 2 places == need for knowing overriding rules. Again -- feel free to change as it suits you

andriytyurnikov: @swalkinshaw , BTW, Gemfiles may be used during gem build and publishing procedure, which may or may not be a relevant factor for making decision here (and which, I believe, caused problem with currently published version of shopify_api gem)

swalkinshaw: On the shopify_api side, having `activeresource` in the Gemfile may have caused the issue but I'm not exactly sure. I realize the confusion between Gemfile and gemspec is annoying and not ideal.

http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

> When developing a gem, use the gemspec method in your Gemfile to avoid duplication. In general, a gem's Gemfile should contain the Rubygems source and a single gemspec line.

andriytyurnikov: I don't see this particular post relevant in our case, as it looks like Yehuda writes in the context of *duplication* in single gemspec and single Gemfile, which is obviously is not the case.
Ironically - duplication is the very thing I intended to avoid by explicitly putting differentiating versions of dependency in different Gemspecs.
With that being said, perhaps idea of not using Gemspec files for different purposes looks appealing:
```Gemspec
Gemspec.ci.rails-master
Gemspec.ci.rails-5-stable
Gemspec.ci.rails-4.2-stable

Comment (swalkinshaw) [inline shopify_app.gemspec]: The only way to vary the dependencies is through Gemfiles though as far as I know. Travis (or any CI) doesn't let you specify the gemspec file. Is what where you were recommending? That would be the ideal solution if it was possible.

Here's another example of how we have it setup: https://github.com/Shopify/active_fulfillment

* Dependencies in gemspec
* Default Gemfile only refers to gemspec
* Other Gemfiles only override what they need to

Comment (andriytyurnikov) [inline shopify_app.gemspec]: @swalkinshaw, I don't see this particular post very relevant in our case, as it looks like Yehuda writes in the context of duplication in single gemspec and single Gemfile, which is obviously is not our case.

Ironically - duplication is the very thing I intended to avoid by explicitly putting differentiating versions of dependency in different Gemspecs.

Another idea - not putting dependency version restrictions into gemspec at all.

And yet another idea-- not using same Gemfiles files for different purposes looks appealing:
```
Gemfile
Gemfile.ci-rails-master
Gemfile.ci-rails-5.0-stable
Gemfile.ci-rails-4.2-stable
```
Another crazy idea -- having both separate Gemfiles and gemspec files for CI matrix:
Sure -- overriding dependencies in Gemspec may be a valid approach, I just avoided it
Anyway, core of the issue I believe is having multiple Gemfiles, for CI matrix, which is a reflection of maintenance/support and versioning policy (of which I have no knowledge at all)

As I mentioned before, my hypothesis about source of shopify_api issue is a use of wrong Gemfile during gem publishing process

File: Gemfile.rails-5.0.0 (hunk 3 above)
Comment (andriytyurnikov) [inline Gemfile.rails-5.0.0]: And here I'll disagree, in example that you've mentioned same redundancy is present
https://github.com/Shopify/active_fulfillment

Using Gemspec for gem publishing, and other Gemspec.*.files with overrides for CI matrix, sounds like solid approach.

This very PR is a bug report about dependencies, which escalated to need for CI matrix, and to usage of different approaches of building CI matrices, and splitting responsibilities between gemspec and Gemfile. I just don't want to have same debate, when rails 5.1 would be announced, if you know what I mean ;)

File: shopify_app.gemspec (hunk 1 above)
Comment (swalkinshaw) [inline shopify_app.gemspec]: @andriytyurnikov how about this, change this PR to *only* address the direct problem. The fix:

```diff
- s.add_dependency('rails', '>= 4.2.6', '< 5.0')
+ s.add_dependency('rails', '>= 4.2.6')
```

After that we'll update the CI suite and matrix to our needs and coordinate new releases.

swalkinshaw:
Closing in favour of https://github.com/Shopify/shopify_app/pull/288

PR #274, opened 2016-06-16
Add hooks for tests in job

gmalette:
@karlhungus @kevinhughes27. I have _no_ idea how to test this because we don't have a `test_framework` generator. I manually 🎩 in script-editor where we do have the generator.


kevinhughes27:
generators are hard to test. I think this is acceptable though 👍

Issue #273, opened 2016-06-15
Support for App Proxies


mikeyhew:
We're going to need to use an app proxy for one of our new apps. I'm working on a fork of shopify_app that adds support for them, and will hopefully be sending a PR your way pretty soon.

Since this will be one of my first open source contributions, what do you need to see in the PR for it to be accepted? Is there any work already in progress on adding App Proxy support to this engine?

kevinhughes27:
No work in progress yet but I would love to see this feature added! 

As far as what you need to do it isn't too complicated - add your code to the engine and provide tests for the functionality. A Readme section is probably in order as well. That should be it, I'm excited to see it!

CgColorsnyc:
Add a proxy to your app

1. Click to open the Apps section of the Partners dashboard

2. Click on your App's name, then click Edit app settingsfind the "App Proxy" section near the bottom.

3. Find the "App Proxy" section near the bottom.

mikeyhew:
#275

PR #275, opened 2016-06-16
Add Support for App Proxies

mikeyhew:
#273 

Here's how we're using this right now:

```ruby
# config/routes.rb
namespace :app_proxy do
  get 'basic' => :basic
  resources :things
end

# app/controllers/app_proxy_controller.rb

class AppProxyController < ApplicationController
  include ShopifyApp::AppProxyVerification
  skip_before_action :verify_authenticity_token
  before_action :verify_proxy_request
 
 def basic
    render text: "Signature verification passed!"
  end
end

# app/controllers/things_controller.rb

class ThingsController < AppProxyController
  # CRUD methods here
end
```

One issue I've found is, if you use `force_ssl` on your ApplicationController, rails will respond to proxy requests with a redirect if the request sent from the browser to Shopify uses HTTP (even if the request from Shopify to your app is HTTPS). This is because Rack uses the first protocol listed in X-Forwarded-Proto to determine whether the request is http or https. For our use case, we added a monkey patch to Rack::Request#scheme that uses the last protocol listed instead.


mikeyhew:
We'll probably want to add a controller under the ShopifyApp namespace that people can inherit from

lib/shopify_app/app_proxy_verification.rb
```diff
@@ -0,0 +1,32 @@
+module ShopifyApp
+  module AppProxyVerification
+    extend ActiveSupport::Concern
+
+    # Usage:
+    #   include ShopifyApp::AppProxyVerification
+    #   before_action :verify_proxy_request
+    def verify_proxy_request
+      return head :unauthorized unless query_string_valid?(request.query_string)
+    end
```
mikeyhew: I didn't an included hook with `before_action :verify_proxy_request` because, well, I didn't know how to run the tests without getting error messages

kevinhughes27:
+1 for adding a controller to the engine. That will make testing easier as well since you can test that controller rather than the concern.

mikeyhew:
@kevinhughes27 I added AppProxyController and some basic tests. Once you're happy with it, I'll add some instructions to the README

lib/shopify_app/app_proxy_verification.rb
kevinhughes27: I'm sorry but I don't understand this comment - can you explain what you mean again?

Also our style would be to not have any comments in the code

test/shopify_app/app_proxy_verification_test.rb
```diff
@@ -0,0 +1,23 @@
+require 'test_helper'
+
+class AppProxyVerificationTest < ActiveSupport::TestCase
+
```
kevinhughes27: extra newline

app/controllers/shopify_app/app_proxy_controller.rb
```diff
@@ -0,0 +1,9 @@
+module ShopifyApp
+  class AppProxyController < ApplicationController
+    include ShopifyApp::AppProxyVerification
+
+    skip_before_action :verify_authenticity_token
+
+    before_action :verify_proxy_request
+  end
+end
```
kevinhughes27: hmm I don't really like that this controller doesn't do anything - the webhook controller provided delegates out to jobs and we were able to establish a pattern for doing so automatically which meant that uses of the gem didn't need to modify the controller at all meaning it could live inside the engine with no issues.

This controller will need to be edited so it doesn't make sense as an engine controller. We have 2 options here:
1. Add this controller to a new generator for app proxies and this controller lives in the host app and includes this module (the module should inject the before filter though)
2. I am not really familiar with App Proxies but perhaps we can do some kind of delegation to jobs like the webhook controller.

Let me know if that makes sense.

lib/shopify_app/app_proxy_verification.rb
kevinhughes27: I think I understand now - using an `included do` block to add the before_action caused the tests to fail? It really should be done like this so can you post the error from your tests or I can check out the branch and we'll figure out a fix here.

mikeyhew: Yes, that was the problem. The tests I wrote for this concern were simple - you just `include AppProxyVerification` directly in the test. But with an `included do` block that calls `before_action` and `skip_before_action`, the concern can only be included from a controller. But now that I've learned more about testing controllers, it will be simple to do. I'll add it in.

app/controllers/shopify_app/app_proxy_controller.rb
mikeyhew: The idea was that people would subclass the controller, not edit it directly. It makes sense to me to have all of your controllers that are behind an app proxy to inherit from `ShopifyApp::AppProxyController`. However, once I add the `included do` hook that we talked about, it will be simple to `include AppProxyVerification` from those controllers.

kevinhughes27: including the concern matches how we've done the other controller types. For webhooks I believe we offer the concern separately and then recommend overwriting the controller if required.

mikeyhew: The way an app proxy works, it doesn't make sense to delegate to jobs like is done for webhooks. Webhooks in their nature are handled asynchronously - the app responds very quickly with a 200 OK, and then it is processed in a worker - whereas app proxy requests are just like any other normal HTTP request that the app receives. The only difference being that they come from Shopify with some extra query parameters, and we need to verify them before we continue processing the request.

mikeyhew:
@kevinhughes27 I just pushed a new commit with the changes we discussed. Let me know what you think. I tried to follow the same conventions as WebHookVerification, particularly in the testing code, but may have missed something.

lib/shopify_app.rb
```diff
@@ -23,6 +23,7 @@
 require 'shopify_app/scripttags_manager'
 require 'shopify_app/webhook_verification'
 require 'shopify_app/utils'
+require 'shopify_app/app_proxy_verification'
```
kevinhughes27: suuuper picky of me but can this require be above `utils`

lib/shopify_app/app_proxy_verification.rb
```diff
@@ -0,0 +1,34 @@
...
+      signature = query_hash.delete('signature')
+      return false if signature.nil?
+
+      sorted_params = query_hash.collect{|k,v| "#{k}=#{Array(v).join(',')}"}.sort.join
+
+      calculated_signature = OpenSSL::HMAC.hexdigest(
+        OpenSSL::Digest.new('sha256'),
+        ShopifyApp.configuration.secret,
+        sorted_params
+      )
```
kevinhughes27: we could refactor out a method for `calculated_signature` to break this code up a bit

kevinhughes27:
Looks really good!

Can we also add a section to the readme to advertise this module to AppProxy users.

lib/shopify_app.rb
mikeyhew: haha sure

lib/shopify_app/app_proxy_verification.rb
```diff
@@ -0,0 +1,38 @@
...
+      return head :unauthorized unless query_string_valid?(request.query_string)
+    end
+
+    def query_string_valid?(query_string)
+      query_hash = Rack::Utils.parse_query(query_string)
+
+      signature = query_hash.delete('signature')
+      return false if signature.nil?
+
+      sorted_params = query_hash.collect{|k,v| "#{k}=#{Array(v).join(',')}"}.sort.join
```
kevinhughes27: might as well pull this into the method as well.

kevinhughes27:
Looks great - one last thing before merging, can you squash your commits please

mikeyhew:
Sure thing

mikeyhew:
Squashed

PR #276, opened 2016-06-17
Support rails 5 in webhooks controller

gmalette:
This PR adds basic support to rails 5 in webhooks controller. The problem is that `ActionController::Parameters` cannot be serialized, and they're empty if not whitelisted using `permit!`. 

I don't know how to test this without a series of potentially breaking PRs in `shopify_api`

@sgnr @kevinhughes27


sgnr:
:shipit:

kevinhughes27:
🚀 🌔

PR #281, opened 2016-06-24
add Shop.connect helper method

mikeyhew:
I use this method a lot when I'm playing around in the console. You call `Shop.first.connect` and then you're free to make api calls to that store for the rest of your console session. Whereas you should favour using `with_shopify_session` pretty much everywhere else in your app, the `connect` method is much more convenient to use when writing throwaway code because you don't have to wrap all of your api calls in a block.

### DIFF
--- a/lib/shopify_app/shop.rb
+++ b/lib/shopify_app/shop.rb
@@ -12,13 +12,8 @@ module ShopifyApp
     end
 
     def connect
-      ShopifyAPI::Base.activate_session(
-        ShopifyAPI::Session.new(
-          ShopifyApp::Utils.sanitize_shop_domain(shopify_domain),
-          shopify_token
-        )
-      )
-      nil
+      session = ShopifyAPI::Session.new(shopify_domain, shopify_token)
+      ShopifyAPI::Base.activate_session(session)
     end
 
   end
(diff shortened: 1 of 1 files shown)


lib/shopify_app/shop.rb
```diff
@@ -11,5 +11,15 @@ def with_shopify_session(&block)
...
+    def connect
+      ShopifyAPI::Base.activate_session(
+        ShopifyAPI::Session.new(
+          ShopifyApp::Utils.sanitize_shop_domain(shopify_domain),
+          shopify_token
+        )
+      )
+      nil
+    end
+
```
kevinhughes27: I would prefer it written like so to avoid extra nesting:

```ruby
def connect
  session = ShopifyAPI::Session.new(shopify_domain, shopify_token)
  ShopifyAPI::Base.activate_session(session)
end
```

I'd also drop the nil return.

kevinhughes27:
I usually add a method like this as well to my apps. I am not sure if we should include it in the framework though since its not really for building your app but for debugging etc and it might confuse users who don't understand its purpose.

I was also considering moving this function into a `.irbrc` file in my projects so it is only available in the console and not in application code.

mikeyhew:
I wouldn't worry too much about that. The way I see it, if we document it properly and people misuse it and run into trouble, that's their problem.

mikeyhew:
On the other hand, if people do get confused, you're the one who's going to be responding to the issues that get posted, so I can see why you'd want to minimize confusion ;)

mikeyhew:
@kevinhughes27 you should probably merge this or close it

Issue #280, opened 2016-06-21
Extract common patterns from ScripttagsManager and WebhooksManager


mikeyhew:
For one of our apps, I'm adding support for CarrierService to a fork of the shopify_api gem. In the process I've noticed that there's some commonalities between CarrierService and the existing code for webhooks and scripttags, and also probably with FullfillmentService which @rjocoleman is adding. All of these resources are created once for every shop in the callback in SessionsConcern, by queueing a job, and have *Manager and *ManagerJob classes with a lot of the same or similar method names.

It would make sense to extract the common patterns into a superclass or mixin. Let's talk about that here.

mikeyhew:
What would be a good name for the *Manager class? ResourceManager comes to mind but it seems a bit too broad...

mikeyhew:
I guess in a sense, they're all event listeners. We could call it ListenerResourceManager

kevinhughes27:
There is definitely some duplication there (and thanks for adding CarrierServices!)

I think the name `ResourceManager` is good.

PR #284, opened 2016-06-29
Add support for Carrier Services

mikeyhew:
Adds `CarrierServicesManager`, `CarrierServicesManagerJob`, `CarrierServiceVerification` and adds a call to `CarrierServicesManager.queue` from `SessionsConcern`. Also adds associated tests.

A couple of notes:
- There may be potential for adding routes and a controller, similar to what is done with webhooks. Delegating to jobs wouldn't work though because unlike webhooks, carrier_service requests need to be processed before responding. For now, we can just provide `CarrierServiceVerification`, until a useful pattern emerges.
- As mentioned in #280, there is lots of duplication with webhooks and scripttags. Adding CarrierServiceVerification was actually just a copy/paste + find/replace of WebhookVerification. Whether this code gets merged in its current state or not, we'll want to abstract away the common points soon before it gets to difficult to maintain. (Once this and #270 get merged, there will be the same code in 4 different places!)

### DIFF
--- a/lib/shopify_app/carrier_services_manager.rb
+++ b/lib/shopify_app/carrier_services_manager.rb
@@ -40,7 +40,7 @@ module ShopifyApp
     def create_carrier_service(attributes)
       attributes.reverse_merge!(format: 'json')
       carrier_service = ShopifyAPI::CarrierService.create(attributes)
-      binding.pry
+
       raise CreationFailed, carrier_service.errors.full_messages.to_sentence unless carrier_service.persisted?
       carrier_service
     end
--- a/lib/shopify_app/carrier_services_manager.rb
+++ b/lib/shopify_app/carrier_services_manager.rb
@@ -1,0 +1,56 @@
+module ShopifyApp
+  class CarrierServicesManager
+    class CreationFailed < StandardError; end
+
+    def self.queue(shop_domain, shop_token)
+      ShopifyApp::CarrierServicesManagerJob.perform_later(shop_domain: shop_domain, shop_token: shop_token)
+    end
+
+    def recreate_carrier_services!
+      destroy_carrier_services
+      create_carrier_services
+    end
+
+    def create_carrier_services
+      return unless required_carrier_services.present?
+
+      required_carrier_services.each do |carrier_service|
+        create_carrier_service(carrier_service) unless carrier_service_exists?(carrier_service[:name])
+      end
+    end
+
+    def destroy_carrier_services
+      ShopifyAPI::CarrierService.all.each do |carrier_service|
+        ShopifyAPI::CarrierService.delete(carrier_service.id) if is_required_carrier_service?(carrier_service)
+      end
+
+      @current_carrier_services = nil
+    end
+
+    private
+
+    def required_carrier_services
+      ShopifyApp.configuration.carrier_services
+    end
+
+    def is_required_carrier_service?(carrier_service)
+      required_carrier_services.map{ |w| w[:name] }.include? carrier_service.name
+    end
+
+    def create_carrier_service(attributes)
+      attributes.reverse_merge!(format: 'json')
+      carrier_service = ShopifyAPI::CarrierService.create(attributes)
+      binding.pry
+      raise CreationFailed, carrier_service.errors.full_messages.to_sentence unless carrier_service.persisted?
+      carrier_service
+    end
+
+    def carrier_service_exists?(name)
+      current_carrier_services[name]
+    end
+
+    def current_carrier_services
+      @current_carrier_services ||= ShopifyAPI::CarrierService.all.index_by(&:name)
+    end
+  end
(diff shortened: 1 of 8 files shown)


lib/shopify_app/carrier_services_manager.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    def recreate_carrier_services!
+      destroy_carrier_services
+      create_carrier_services
+    end
+
+    def create_carrier_services
+      return unless required_carrier_services.present?
+
+      required_carrier_services.each do |carrier_service|
```
kevinhughes27: if `required_carrier_services` defaulted to `[]` we wouldn't need to check presence and could just have the loop which would be a no-op

lib/shopify_app/carrier_services_manager.rb
```diff
@@ -0,0 +1,55 @@
...
+        create_carrier_service(carrier_service) unless carrier_service_exists?(carrier_service[:name])
+      end
+    end
+
+    def destroy_carrier_services
+      ShopifyAPI::CarrierService.all.each do |carrier_service|
```
kevinhughes27: why does this not call `current_carrier_services`?

lib/shopify_app/carrier_services_manager_job.rb
```diff
@@ -0,0 +1,10 @@
+module ShopifyApp
+  class CarrierServicesManagerJob < ActiveJob::Base
+    def perform(shop_domain:, shop_token:)
+      ShopifyAPI::Session.temp(shop_domain, shop_token) do
+        manager = CarrierServicesManager.new
+        manager.create_carrier_services
```
kevinhughes27: I think this should be `recreate_carrier_services!` so that on each auth we are making sure the shop has what we need. That being said destroying and creating sucks. We should look into making recreate smarter so that it checks to make sure what we want is there

test/shopify_app/carrier_service_verification_test.rb
```diff
@@ -0,0 +1,52 @@
+require 'test_helper'
+require 'action_controller'
+require 'action_controller/base'
+
+class CarrierServiceVerificationController < ActionController::Base
+  include ShopifyApp::CarrierServiceVerification
+  def carts_update
```
kevinhughes27: can you rename this fake endpoint to something more realistic for this controller

lib/shopify_app/carrier_services_manager.rb
mikeyhew: I don't know - I just copied `destroy_webhooks`:
```ruby
def destroy_webhooks
  ShopifyAPI::Webhook.all.each do |webhook|
    ShopifyAPI::Webhook.delete(webhook.id) if is_required_webhook?(webhook)
  end

  @current_webhooks = nil
end
```

test/shopify_app/carrier_service_verification_test.rb
mikeyhew: I can, but again, I was just copying from webhooks. Until we refactor, wouldn't it make sense to just keep them the same?

mikeyhew:
I like your idea about `verification_utils`. It seems to be pretty common that Shopify will sign `request.raw_post` and place the signature in the 'X-Shopify-Hmac' header. App proxies are the only exception as far as I know.

test/shopify_app/carrier_service_verification_test.rb
kevinhughes27: carts_update is an actual webhook name though which is why it was named so. Could you name this endpoint `callback` since that seems closest to how a carrier service would be used.

lib/shopify_app/carrier_services_manager.rb
kevinhughes27: I think they should both use the `current_*` method to be a bit more DRY

test/shopify_app/carrier_service_verification_test.rb
mikeyhew: Oh I see - sorry I thought you meant the name of the controller

lib/shopify_app/carrier_services_manager_job.rb
kevinhughes27: I was just working on the webhook manager and I noticed from reading the code that I think `create_*` already does what I describe (at least for webhooks). It tries to create all the required hooks - if we get an error we check if the webhook is persisted already and if it is we continue without raising and error. This might be unique to how webhooks because of topic validations so we should confirm still and this probably doesn't apply directly to carrier services

kevinhughes27:
Would be great to get this in for version 8 - do you have anytime to finish this up?

mikeyhew:
Closing because I doubt I will have time to work on this any time soon. People are welcome to look at this code as a starting point for adding the feature — if it's still relevant.

PR #285, opened 2016-07-01
Add note to README about scripttags and Rails javascript asset

cymen:
I'm not working in Rails all the time so it took a while to figure out the right place to put in the configuration that would use the proper asset file path (without this tweak, I'd get the non-fingerprinted path to the `application.js` asset which would 404).


kevinhughes27:
So I understand correctly this tweak is to allow serving a script tag that is processed through the asset pipeline?

cymen:
Yes, exactly.

README.md
```diff
@@ -190,6 +190,21 @@ end
 
 Scripttags are created in the same way as the Webhooks, with a background job which will create the required scripttags.
 
+If you want to use the application.js file as generated by the Rails asset pipeline, you can do so by moving the `ShopifyApp.configure`
```
kevinhughes27: `application.js` probably isn't the file you want to send as the script tag. Can you edit this line to just refer to bundles process by the rails asset pipeline

README.md
```diff
@@ -190,6 +190,21 @@ end
...
+to `config/application.rb` like so:
+
+```ruby
+config.after_initialize do
+	ShopifyApp.configure do |shopify_config|
+		shopify_config.scripttags = [
+			{event: 'onload', src: "https://example.com#{ActionController::Base.helpers.javascript_url('application')}"}
```
kevinhughes27: I wonder if we can lazy init this inside the class as is rather than having to move the config. If this method isn't called until rails is initialized fully won't we get the fingerprint?

README.md
cymen: I agree -- I just wanted to get a demo app working with integration as a proof of concept so I was using `application.js`.

README.md
cymen: I think that would solve it. If it is called too early, it just emits the same filename as in development (no fingerprint).

README.md
cymen: Ah, I reread. Hrm... Are you suggesting sending all the bundles generated by the asset pipeline? I think one would typically send one bundle, no?

kevinhughes27: Nope I meant I'd like this read that you can send one of your bundles of your choosing :)

README.md
```diff
@@ -190,6 +190,21 @@ end
...
+		]
+	end
+end
```
kevinhughes27: can you fix the indentation here to 2 space?

cymen:
I've left the employer I was working on this for -- if anyone else cares about this, please clean it up and get it in.

Issue #286, opened 2016-07-01
Current Master - 0554b47 does not load app_proxy/index.html by default


ashmaroli:
Current Master - 0554b47 does not load index.html
------------------------

I'm working on a test-app in rails, deployed to Heroku with 2 new controllers and routes.rb configured as per the Readme file.The app has also been set up properly at the Partner's Dashboard.

On trying to access the app from the storefront, I'm served a 404 Status with the URL:
```
https://appname.herokuapp.com/app_proxy/?shop=shopname.myshopify.com&path_prefix=%2Fapps%2Fappname&timestamp=
```
The page does not open the associated index.html inside _\app\views\app_proxy\index.html.erb_ by default.
but it loads if I were to *manually* insert 'index' or 'basic' into the url as 
```
https://appname.herokuapp.com/app_proxy/index?shop=shopname.myshopify.com&path_prefix=%2Fapps%2Fappname&timestamp=
```

kevinhughes27:
I believe that is intended. You need to specify what action should be used.
cc @mikeyhew

mikeyhew:
@ashmaroli why would you expect your index.html file to be rendered if there is no controller action for it?

mikeyhew:
Oh wait, I guess you do have a controller action for basic. OK, I'm assuming this is what your `config/routes.rb` file looks like:
```ruby
namespace :app_proxy do
  get :index
  get :basic
end
```
If that's the case, then you don't have a route for "/app_proxy/" - so of course you'll get a 404 at that url

ashmaroli:
@mikeyhew, I was testing the proxy verification as per the changes you made to the readme file.
Yes, my routes.rb does look like that.. I even added a get 'app_proxy/index' and get 'app_proxy/basic' in a later commit..but that didnt work obviously.

So, shouldn't the image you added to the readme file be showing instead:

**Proxy URL**
`https://your_app_website.com/app_proxy/index`

or did I fail to see something entirely?
Thanks in advance. 
:)

mikeyhew:
No, the proxy url in the image is correct. Assuming the proxy url you chose is "apps/your-app", a request to [some-store-name.myshopify.com/apps/your-app/some/sub/url]() will be routed to  [https://your_app_website.com/app_proxy/some/sub/url](). Does that make sense?

ashmaroli:
Yes, @mikeyhew it does make sense and it works correctly too.
[...]

PR #290, opened 2016-07-06
Update App Proxy usage in README.md

mikeyhew:
- Made the routes that get created more explicit
- added a root example, prompted by #286


kevinhughes27:
👍

PR #288, opened 2016-07-05
easier dependencies

andriytyurnikov:

PR #292, opened 2016-07-07
Travis CI - test multiple Rails versions

swalkinshaw:
Ref: https://github.com/Shopify/shopify_app/pull/288

Adds a Gemfile to ensure we're testing against Rails `~> 5.0`. Even though it's the latest version, this gem will still resolve to Rails `4.2.6` by default.

This also adds Ruby `2.3.1` to be tested in Travis.

/cc @kevinhughes27

### DIFF
--- a/lib/shopify_app/app_proxy_verification.rb
+++ b/lib/shopify_app/app_proxy_verification.rb
@@ -3,7 +3,7 @@ module ShopifyApp
     extend ActiveSupport::Concern
 
     included do
-      skip_before_action :verify_authenticity_token
+      skip_before_action :verify_authenticity_token, raise: false
       before_action :verify_proxy_request
     end
 
--- a/shopify_app.gemspec
+++ b/shopify_app.gemspec
@@ -8,8 +8,9 @@ Gem::Specification.new do |s|
   s.author      = "Shopify"
   s.summary     = %q{This gem is used to get quickly started with the Shopify API}
 
-  s.add_dependency('rails', '>= 4.2.6')
+  s.required_ruby_version = ">= 2.2.4"
 
+  s.add_runtime_dependency('rails', '>= 4.2.6')
   s.add_runtime_dependency('shopify_api', '~> 4.2.2')
   s.add_runtime_dependency('omniauth-shopify-oauth2', '~> 1.1.11')
 
--- a/test/controllers/sessions_controller_test.rb
+++ b/test/controllers/sessions_controller_test.rb
@@ -38,7 +38,7 @@ module ShopifyApp
     test "#new should render a full-page if the shop param doesn't exist" do
       get :new
       assert_response :ok
-      assert_template :new
+      assert_match %r{Shopify App — Installation}, response.body
     end
 
     ['my-shop', 'my-shop.myshopify.com', 'https://my-shop.myshopify.com', 'http://my-shop.myshopify.com'].each do |good_url|
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ sudo: false
 
 rvm:
   - 2.2.4
+  - 2.3.1
 
 gemfile:
   - Gemfile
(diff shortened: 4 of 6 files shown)


kevinhughes27:
👍

test/controllers/sessions_controller_test.rb
```diff
@@ -38,7 +38,6 @@ class SessionsControllerTest < ActionController::TestCase
     test "#new should render a full-page if the shop param doesn't exist" do
       get :new
       assert_response :ok
-      assert_template :new
```
swalkinshaw: In Rails 5.0, `assert_template` only exists in https://github.com/rails/rails-controller-testing now. Figured it's easier to remove the need for it since the test is still useful.

swalkinshaw:
@kevinhughes27 had to make 2 small fixes for Rails 5 compat if you want to give it another look.

lib/shopify_app/app_proxy_verification.rb
```diff
@@ -3,7 +3,7 @@ module AppProxyVerification
     extend ActiveSupport::Concern
 
     included do
-      skip_before_action :verify_authenticity_token
+      skip_before_action :verify_authenticity_token, raise: false
```
swalkinshaw: Rails 5 raises an exception if you try to act on an action that isn't defined. Easiest compatibility option is to skip raising the error so it acts like Rails 4.

kevinhughes27: to clarify Rails now raises if you try and skip and action that doesn't exist? In that case 👍

test/controllers/sessions_controller_test.rb
kevinhughes27: fine with me. Perhaps we can add an assertion on the contents of response.body to replace this

lib/shopify_app/app_proxy_verification.rb
swalkinshaw: correct 👍

test/controllers/sessions_controller_test.rb
swalkinshaw: Good idea, added an assertion

lib/shopify_app/app_proxy_verification.rb
ashmaroli: so to confirm:- current shopify_app passed the test using ruby >= v2.3 required by current version of shopify_api ?

swalkinshaw: Yeah, shopify_app passes the tests on 2.3.1 which is what shopify_api uses now. There were no changes necessary for Ruby 2.3.1. Just for Rails 5.0.

ashmaroli: Okay. Thanks for the clarification. :)

shopify_app.gemspec
```diff
@@ -8,8 +8,9 @@ Gem::Specification.new do |s|
   s.author      = "Shopify"
   s.summary     = %q{This gem is used to get quickly started with the Shopify API}
 
-  s.add_dependency('rails', '>= 4.2.6')
+  s.required_ruby_version = ">= 2.2.4"
```
sshaw: Why 2.2.4?

swalkinshaw: It's the lowest version we test on.

sshaw: It's a bit ridged. At least go with Rails 5 minimum required ruby version: 2.2.2.

sshaw: Also this is on 7.0.11 which prevents existing users using <= 2.2.4 from upgrading and taking advantage of the scripttag updates... 😿

swalkinshaw: Yeah `2.2.4` does seem a little arbitrary. @kevinhughes27 what do you think about moving this to either:

`>= 2.2` or `>= 2.2.2` in the next release?

kevinhughes27: +1 from me

PR #293, opened 2016-07-08
Pass Config to Webhook and Scripttag Managers

kevinhughes27:
Currently our resource managers read the configuration from ShopifyApp directly to determine what they need to do which works fine in environments where this config is static (this covers probably 95% of use cases). We have some applications that host multiple Shopify Apps where this is not the case and blocks us from using the managers (One use case for multiple apps is integrations where each integration warrants a spot in the app store and a separate experience for the user but a lot of code can be shared on the backend eg. accounting export).

The solution to this problem is to pass the configuration into the manager before creating the resources. This way if the ShopifyApp config is changed by the controller the most up to date information is used by the manager. As a bonus its also easier to test the managers now 😄 

For review:
@t6d @MarcLapierre 
cc @mikeyhew @redronin

### DIFF
--- a/lib/shopify_app/sessions_concern.rb
+++ b/lib/shopify_app/sessions_concern.rb
@@ -11,16 +11,10 @@ module ShopifyApp
     end
 
     def callback
-      if response = request.env['omniauth.auth']
-        shop_name = response.uid
-        token = response['credentials']['token']
-
-        sess = ShopifyAPI::Session.new(shop_name, token)
-        session[:shopify] = ShopifyApp::SessionRepository.store(sess)
-        session[:shopify_domain] = shop_name
-
-        WebhooksManager.queue(shop_name, token) if ShopifyApp.configuration.has_webhooks?
-        ScripttagsManager.queue(shop_name, token) if ShopifyApp.configuration.has_scripttags?
+      if auth_hash
+        login_shop
+        install_webhooks
+        install_scripttags
 
         flash[:notice] = I18n.t('.logged_in')
         redirect_to_with_fallback return_address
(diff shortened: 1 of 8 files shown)


test/shopify_app/webhooks_manager_test.rb
```diff
@@ -34,10 +32,9 @@ class ShopifyApp::WebhooksManagerTest < ActiveSupport::TestCase
     end
   end
 
-  test "#create_webhooks when creating a webhook fails and the webhook exists, do not raise an error" do
+  test "#create_webhooks doesn't create webhooks that are already created" do
     webhook = stub(persisted?: false)
-    webhooks = all_webhook_topics.map{|t| stub(topic: t)}
-    ShopifyAPI::Webhook.stubs(create: webhook, all: webhooks)
+    ShopifyAPI::Webhook.stubs(create: webhook, all: all_mock_webhooks)
```
kevinhughes27: this test was poorly named and the implementation was confusing. I re-wrote to better explain what we care about here

lib/shopify_app/sessions_concern.rb
```diff
@@ -11,16 +11,10 @@ def create
...
-        shop_name = response.uid
-        token = response['credentials']['token']
-
-        sess = ShopifyAPI::Session.new(shop_name, token)
-        session[:shopify] = ShopifyApp::SessionRepository.store(sess)
-        session[:shopify_domain] = shop_name
-
-        WebhooksManager.queue(shop_name, token) if ShopifyApp.configuration.has_webhooks?
-        ScripttagsManager.queue(shop_name, token) if ShopifyApp.configuration.has_scripttags?
+      if auth_hash
```
MarcLapierre: so much nicer.

MarcLapierre:
Looks great 🎉

t6d:
Nice! 👍

PR #295, opened 2016-07-13
App Proxy Generator

ashmaroli:
This branch deals with the addition of an optional App Proxy Controller Generator that is based on existing generators within shopify_app.

It creates:

- The AppProxyController which inherits from ApplicationController and conditionally includes the AppProxyVerification mixin alongwith sending "Content-Type='application/liquid'" header for only production environment.
- The associated view for index action that displays current shop info.

Additionally it injects the namespace route for the controller as currently defined into the routes.rb.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,13 @@ Table of Contents
 * [**Quickstart**](#quickstart)
 * [**Becoming a Shopify App Developer**](#becoming-a-shopify-app-developer)
 * [**Installation**](#installation)
+* [**Rails 5**](#rails-5)
 * [**Generators**](#generators)
  * [Default Generator](#default-generator)
  * [Install Generator](#install-generator)
  * [Shop Model Generator](#shop-model-generator)
  * [Home Controller Generator](#home-controller-generator)
+ * [App Proxy Controller Generator](#app-proxy-controller-generator)
  * [Controllers, Routes and Views](#controllers-routes-and-views)
 * [**Managing Api Keys**](#managing-api-keys)
 * [**WebhooksManager**](#webhooksmanager)
@@ -31,6 +33,7 @@ Table of Contents
 * [**Troubleshooting**](#troubleshooting)
  * [Generator shopify_app:install hangs](#generator-shopify_appinstall-hangs)
 * [**Testing an embedded app outside the Shopify admin**](#testing-an-embedded-app-outside-the-shopify-admin)
+* [**App Tunneling**](#app-tunneling)
 * [**Questions or problems?**](#questions-or-problems)
 
 
@@ -153,6 +156,15 @@ $ rails generate shopify_app:home_controller
 This generator creates an example home controller and view which fetches and displays products using the ShopifyAPI
 
 
+### App Proxy Controller Generator
+
+```sh
+$ rails generate shopify_app:app_proxy_controller
+```
+
+This optional generator, not included with the default generator, creates the app proxy controller to handle proxy requests to the app from your shop storefront, modifies 'config/routes.rb' with a namespace route, and an example view which displays current shop information using the LiquidAPI
+
+
 ### Controllers, Routes and Views
 
 The last group of generators are for your convenience if you want to start overriding code included as part of the Rails engine. For example by default the engine provides a simple SessionController, if you run the `rails generate shopify_app:controllers` generator then this code gets copied out into your app so you can start adding to it. Routes and views follow the exact same pattern.
@@ -249,43 +261,18 @@ The engine provides a mixin for verifying incoming HTTP requests sent via an App
 
 ### Recommended Usage
 
-1. Use the `namespace` method to create app proxy routes
-    ```ruby
-    # config/routes.rb
-    namespace :app_proxy do
-      # simple routes without a specified controller will go to AppProxyController
-      # GET '/app_proxy/basic' will be routed to AppProxyController#basic
-      get :basic
+The App Proxy Controller Generator automatically adds the mixin to the generated app_proxy_controller.rb
+Additional controllers for resources within the App_Proxy namespace, will need to include the mixin like so: 
 
-      # this will route GET /app_proxy to AppProxyController#main
-      root action: :main
+```ruby
+# app/controllers/app_proxy/reviews_controller.rb
+class ReviewsController < ApplicationController
+  include ShopifyApp::AppProxyVerification
+  # ...
+end
+```
 
-      # more complex routes will go to controllers in the AppProxy namespace
-      resources :reviews
-      # GET /app_proxy/reviews will now be routed to
-      # AppProxy::ReviewsController#index, for example
-    end
-    ```
-
-2. `include` the mixin in your app proxy controllers
-    ```ruby
-    # app/controllers/app_proxy_controller.rb
-    class AppProxyController < ApplicationController
-      include ShopifyApp::AppProxyVerification
-
-      def basic
-        render text: 'Signature verification passed!'
-      end
-    end
-
-    # app/controllers/app_proxy/reviews_controller.rb
-    class ReviewsController < ApplicationController
-      include ShopifyApp::AppProxyVerification
-      # ...
-    end
-    ```
-
-3. Create your app proxy url in the [Shopify Partners' Dashboard](https://app.shopify.com/services/partners/api_clients), making sure to point it to `https://your_app_website.com/app_proxy`.
+Create your app proxy url in the [Shopify Partners' Dashboard](https://app.shopify.com/services/partners/api_clients), making sure to point it to `https://your_app_website.com/app_proxy`.
 ![Creating an App Proxy](/images/app-proxy-screenshot.png)
 
 Troubleshooting
@@ -309,6 +296,12 @@ By default, loading your embedded app will redirect to the Shopify admin, with t
 ```javascript
 forceRedirect: <%= Rails.env.development? || Rails.env.test? ? 'false' : 'true' %>
 ```
+
+App Tunneling
+-------------
+
+For certain features like Application Proxy or Webhooks to receive requests from Shopify, your app needs to be on a publicly visible URL. This can be a hurdle during development or testing on a local machine. Fortunately, this can be overcome by employing a tunneling service like [Forward](https://forwardhq.com/), [RequestBin](requestb.in/), [ngrok](https://ngrok.com/) etc. These tools allow you to create a secure tunnel from the public Internet to your local machine.
+
 
 Questions or problems?
 ----------------------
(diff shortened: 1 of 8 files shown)


mikeyhew:
Hey @ashmaroli, thanks for getting the ball rolling on this. I have a few questions/suggestions:
- I suggest you change the generator's name from `app_proxy_controller` to `app_proxy`, since it's shorter and will avoid conflict later if we add a controller generator similar to `rails g controller`.
- Why wouldn't you send the "Content-Type: application/liquid" header during development? I'm assuming you'd want to see what the output of your liquid code is when you're developing...

lib/generators/shopify_app/app_proxy_controller/app_proxy_controller_generator.rb
```diff
@@ -0,0 +1,25 @@
...
+      def create_app_proxy_controller
+        template 'app_proxy_controller.rb', 'app/controllers/app_proxy_controller.rb'
+      end
+
+      def create_app_proxy_index_view
+        copy_file 'index.html.erb', 'app/views/app_proxy/index.html.erb'
+      end
+
+      def add_app_proxy_route
+        inject_into_file(
```
mikeyhew: use `route` instead of `inject_into_file`: http://api.rubyonrails.org/classes/Rails/Generators/Actions.html#method-i-route

lib/generators/shopify_app/app_proxy_controller/templates/app_proxy_controller.rb
```diff
@@ -0,0 +1,12 @@
+class AppProxyController < ApplicationController
+  if Rails.env.development? || Rails.env.test?
+   around_action :shopify_session
+  else
+   include ShopifyApp::AppProxyVerification
+  end
```
mikeyhew: Oh i see... you don't have access to Shopify during development? I always just use a dev shop and access the app proxy route through the shop.

lib/generators/shopify_app/app_proxy_controller/app_proxy_controller_generator.rb
ashmaroli: Thanks for the comments, Michael
I did consider using `route` initially over `inject_into_file` but I wasn't sure if route method could inject multiple lines ('\n') at the desired position ( `after: mount...` )

lib/generators/shopify_app/app_proxy_controller/templates/app_proxy_controller.rb
ashmaroli: The reason for this move was multi-fold:
- Since there's no way to access currently active shop's theme via localhost, there's actually no point in sending a special content-type header. 
- I dont have a Liquid Template/Parse-Render Engine defined for my local server. So during development, everytime I tried accessing the page with Content-type: 'application/liquid', the page would automatically get downloaded as a file, by the browser, instead.
- I assumed that the new app developers trying their hand at App Proxy may not have their local server to handle liquid headers either and hence would instead show them a default render.

Yes, I checked production status via a dev shop.

ashmaroli: the controller was named **app_proxy_controller** in sync with existing generator for home_controller. 
`rails generate controller shopify_app:app_proxy_controller`

The other shorter alternative to 'app_proxy_controller' can be made to be **proxy_controller**
`rails g controller shopify_app:proxy_controller`

Once again, thanks for the inputs. :)

lib/generators/shopify_app/app_proxy_controller/app_proxy_controller_generator.rb
ashmaroli: I had a closer look at the link you provided.
the route method itself uses `inject_into_file`

``` ruby
def route(routing_code)
  log :route, routing_code
  sentinel = /\.routes\.draw do\s*\n/m

  in_root do
    inject_into_file 'config/routes.rb', "  #{routing_code}\n", { after: sentinel, verbose: false, force: false }
  end
end
```

lib/generators/shopify_app/app_proxy_controller/templates/app_proxy_controller.rb
mikeyhew: About the liquid content-type stuff:

I don't really understand your first point. Why can't you use a dev shop in development?

ashmaroli: Because

- I simply never knew that you could use a dev shop locally. 
- I thought dev shop was like any other shop except it didnt have a validity nor billing fn. attached to it.

Please point me to the documentation to use a dev shop locally.
Thanks in advance.

mikeyhew: >I thought dev shop was like any other shop except it didnt have a validity nor billing fn. attached to it.

It is. You can't *run* Shopify on your own computer. But you can use a tunnelling service - we use [forward](https://forwardhq.com/) at the moment so we can use app proxies, webhooks, etc. in development.

ashmaroli: Ohh.. now, why didn't I think of this..? lol! Thanks. 
I'll make a note of it in the README changes as well.. It'll be helpful to somebody out there, if this PR does get merged.
:)

kevinhughes27: Yes please update to include docs and recommend using a tunnel. Thats how all the other things work (like testing webhooks locally).

ashmaroli:
@mikeyhew, If we decide to go ahead with the name **_proxy_controller.rb_** instead of **_app_proxy_controller.rb_**, I feel all instances of _app_proxy_ (excl. _app_proxy_verification.rb_) should be renamed too, to avoid complex routing.

This would mean:
On running
`rails g controller shopify_app:proxy` would give
```ruby
# config/routes.rb
  namespace :proxy do
    root action: 'index'
    # simple routes without a specified controller will go to ProxyController
    
    # more complex routes will go to controllers in the Proxy namespace
    # 	resources :reviews
    # GET /proxy/reviews will now be routed to
    # Proxy::ReviewsController#index, for example
  end

# controllers/proxy_controller.rb
class ProxyController < ApplicationController
   include ShopifyApp::AppProxyVerification

  def index
  end

end

# if any resource controllers to be defined later
# controllers/proxy/reviews_controller.rb
class Proxy::ReviewsController < ApplicationController
   include ShopifyApp::AppProxyVerification

  def index
  end

end
```
Also requires appropriate changes to `/images/app-proxy-screenshot.png` and related documentation.

ashmaroli:
Moved to PR #301

README.md
```diff
@@ -270,6 +254,12 @@ By default, loading your embedded app will redirect to the Shopify admin, with t
 forceRedirect: <%= Rails.env.development? || Rails.env.test? ? 'false' : 'true' %>
 ```
 
+App Tunneling
+-------------
+
+For certain features like Application Proxy or Webhooks to receive requests from Shopify, your app needs to be on a publicly visible URL. This can be a hurdle during development or testing on a local machine. Fortunately, this can be overcome by employing a tunneling service like [Forward](https://forwardhq.com/), [RequestBin](requestb.in/), [ngrok](https://ngrok.com/) etc. These tools allow you to create a secure tunnel from the public Internet to your local machine.
+
```
kevinhughes27: is ngrok mentioned elsewhere in the readme? We should have one section like this and reference it

ashmaroli: no kevin, this is the only mention.

kevinhughes27: 👍

kevinhughes27:
I am good with this. @mikeyhew you good to 🚢 ?

README.md
ashmaroli: @kevinhughes27 wont merging this commit as is remove the TOC from the README? Please wait for me to push a new commit to this branch.

kevinhughes27: Nope it will just merge it on top. You can see all the changes that will occur in the diff of this PR. Git is smart like that :)

ashmaroli: Ah, Thats cool! I've pushed a commit to update the ToC to reference the AppProxyController Generator documentation and other changes to README.

kevinhughes27:
Can you squash the commits please. This can all be one commit

mikeyhew:
My only nitpick is that I'd prefer `app_proxy` instead of `app_proxy_controller`, because to me it resembles the idea of "setting up the app proxy" rather than just creating a controller (it adds the root route and default liquid template, and the namespace). Although I realize that the `home_controller` generator does pretty much exactly that. But `app_proxy` is shorter :wink:

@ashmaroli I'll leave you with the final decision, since I'm just :bike:shedding. Other than that I'm good to go :+1:

kevinhughes27:
Woah commits are all over the place now. You need to rebase on upstream master and sqaush

ashmaroli:
I think Sebastian's PR may need to be re-merged  as well as #303 if thats even possible.

ashmaroli:
All conflicts resolved. Ready to :ship:

ashmaroli:
@mikeyhew, I decided to keep `app_proxy_controller` as the invoke name. I felt it to be semantically closer to the underlying action: `Generate a controller to handle proxy requests to the app.`

But we both can be happy with the inclusion of another .rb file with contents as:
```ruby
# lib\generators\shopify_app\app_proxy_generator.rb

module ShopifyApp
  module Generators
    class AppProxyGenerator < Rails::Generators::Base
      def app_proxy_controller_generators
        generate "shopify_app:app_proxy_controller"
      end
    end
  end
end
```
This file would be invoked by simply `rails g shopify_app:app_proxy`
```sh
$ rails g shopify_app:app_proxy

    generate shopify:app_proxy_controller

    create  app/controllers/app_proxy_controller.rb
    create  app/views/app_proxy/index.html.erb
    insert  config/routes.rb
```
another idea I toyed with, was to modify the .rb further to extract the generator from the shopify_app module to simply call `rails g app_proxy`. But, either ways, I haven't included the .rb in the PR though. Felt it to be an unnecessary step adding additional time to generate the controller.

mikeyhew:
Don't bother adding another generator if it does the same thing. Either name is OK, just a preference. I'm good to :ship:

ashmaroli:
**Newbie Question:**
Does this PR and #298 together count as added functionality, hence a minor version bump (7.0.11 => 7.1.0 ) as per semantic versioning? I'll delete this query once replied to.

kevinhughes27:
Yeah we should probably do a minor version bump now.

kevinhughes27:
Can you add one more commit adding to the CHANGELOG and updating the version.

ashmaroli:
okay. Will do.

kevinhughes27:
+1 looks like you need a rebase

PR #296, opened 2016-07-14
soften shopify_api dependency

mikeyhew:
Allows `shopify_api` versions above 4.2 while disallowing < 4.2.2, fixing the regression in 13d92a8415b1f08d260a529b762077b04f3dd619

### DIFF
--- a/shopify_app.gemspec
+++ b/shopify_app.gemspec
@@ -11,7 +11,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 2.2.4"
 
   s.add_runtime_dependency('rails', '>= 4.2.6')
-  s.add_runtime_dependency('shopify_api', '>= 4.2.2', '< 5.0')
+  s.add_runtime_dependency('shopify_api', '>= 4.2.2')
   s.add_runtime_dependency('omniauth-shopify-oauth2', '~> 1.1.11')
 
   s.add_development_dependency('rake')
--- a/shopify_app.gemspec
+++ b/shopify_app.gemspec
@@ -11,7 +11,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 2.2.4"
 
   s.add_runtime_dependency('rails', '>= 4.2.6')
-  s.add_runtime_dependency('shopify_api', '~> 4.2.2')
+  s.add_runtime_dependency('shopify_api', '>= 4.2.2', '< 5.0')
   s.add_runtime_dependency('omniauth-shopify-oauth2', '~> 1.1.11')
 
   s.add_development_dependency('rake')
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,15 @@ cache: bundler
 sudo: false
 
 rvm:
-  - 2.2.4
   - 2.3.1
 
 gemfile:
   - Gemfile
   - Gemfile.rails50
+
+matrix:
+  include:
+  - rvm: 2.2.4
+    gemfile: Gemfile.ruby22
+  - rvm: 2.2.4
+    gemfile: Gemfile.ruby22.rails50
--- a/Gemfile.ruby22
+++ b/Gemfile.ruby22
@@ -1,0 +1,6 @@
+source "https://rubygems.org"
+
+# Specify your gem's dependencies in shopify_app.gemspec
+gemspec
+
+gem 'shopify_api', '< 4.3'
--- a/Gemfile.ruby22.rails50
+++ b/Gemfile.ruby22.rails50
@@ -1,0 +1,9 @@
+source "https://rubygems.org"
+
+# Specify your gem's dependencies in shopify_app.gemspec
+gemspec
+
+gem 'rails', '~> 5.0'
+gem 'activeresource', github: 'rails/activeresource'
+
+gem 'shopify_api', '< 4.3'


SebastianSzturo:
This only works if you use the latest ActiveResource master branch. Same as here: https://github.com/Shopify/shopify_api/pull/283/files

We should make sure to mention that in the README.

mikeyhew:
It's weird the way the build is failing - isn't bundler sophisticated enough to figure out that it can use `shopify_api` version 4.2 without any problems?

mikeyhew:
@SebastianSzturo why do you say that? The latest version of `shopify_api` only requires `activeresource >= 3.0.0`: [https://rubygems.org/gems/shopify_api](https://rubygems.org/gems/shopify_api).

buggy:
I came across this problem last night while writing a tutorial on building Shopify apps with Rails 5.

@mikeyhew with a new Rails 5 installing including `shopify_app 7.0.11` results in a conflict because Rails 5 uses `activemodel 5.0.0` but `shopify_ap 7.0.11` results in

```sh
shopify_app (>= 7.0.11, ~> 7.0) was resolved to 7.0.11, which depends on
  shopify_api (~> 4.2.2) was resolved to 4.2.2, which depends on
    activeresource was resolved to 3.0.0, which depends on
      activemodel (= 3.0.0)
``` 
Adding `activeresource 4.1` to the Gemfile also doesn't work as it's locked to `activemodel 4.0`. The docs on https://github.com/Shopify/shopify_api also mention this and give the solution as using the unreleased master version which means adding this your `Gemfile`.

```
gem 'activeresource', github: 'rails/activeresource'
```

SebastianSzturo:
I will try to get this merged into master next week and investigate what we can do about `activeresource`. For the meantime you can use the [rails5 branch](https://github.com/Shopify/shopify_app/tree/rails5):

```ruby
gem 'shopify_api', github: 'shopify/shopify_app', branch: 'rails5'
gem 'activeresource', github: 'rails/activeresource'
```

mikeyhew:
@SebastianSzturo @buggy The point of this PR is to allow versions of `shopify_api` above 2.2. The reason the build for ruby 2.2.4 is failing is because bundler is resolving `shopify_api` to version 2.3.0, which requires a ruby version >= 2.3. It fails for both rails 4 and 5, so I don't see how this has anything to do with rails 5 or the fact that activeresource 5.0 hasn't been released yet.

mikeyhew:
At the risk of repeating myself, the point of this PR is meant to allow a higher version of `shopify_api`, not a higher version of `rails`. @buggy the problem you came across has to do with using rails 5 with the `shopify_app` and `shopify_api` gems, which is outside the scope of this PR.

SebastianSzturo:
You are right. This PR has nothing to do with rails 5 support, but it is a prerequisite for rails 5.

shopify_app.gemspec
```diff
@@ -11,7 +11,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 2.2.4"
 
   s.add_runtime_dependency('rails', '>= 4.2.6')
-  s.add_runtime_dependency('shopify_api', '~> 4.2.2')
+  s.add_runtime_dependency('shopify_api', '>= 4.2.2', '< 5.0')
```
kevinhughes27: why add the upper bound?

mikeyhew: Good question. I added it because before 13d92a8 it was `~> 4.2`, which means `>= 4.2.0 && < 5.0`. If you aren't too worried about breaking changes in the next major version of `shopify_api`, then I'd be happy to remove it.

kevinhughes27: I'm not overly worried about a change to ShopifyAPI breaking ShopifyApp so I would be ok with it

mikeyhew:
@SebastianSzturo is it though? I'm able to run these commands without any errors, using ruby 2.3.1:
```
rails _5.0.0_ new rails5test --skip-bundle
cd rails5test
echo "gem 'activeresource', github: 'rails/activeresource'" >> Gemfile
echo "gem 'shopify_app', github: 'Shopify/shopify_app'" >> Gemfile
bundle install
```

shopify_app.gemspec
ashmaroli: since its the Ruby Version that this PR fails CI, shouldn't the gemspec be:
```ruby
s.required_ruby_version = "~> 2.3"

s.add_runtime_dependency('rails', '>= 4.2.6')
s.add_runtime_dependency('shopify_api', '>= 4.2.2')
...
```

SebastianSzturo:
@mikeyhew You are right! This has nothing to do with rails 5. Sorry about the confusion.

mikeyhew:
Apparently the issue with bundler was supposed to be resolved in version 1.12, but that doesn't seem to be the case. There's a closed issue here: https://github.com/bundler/bundler-features/issues/95

mikeyhew:
@kevinhughes27 I got the build passing, but I had to add separate Gemfiles for ruby 2.2.4 because of the bundler bug referenced above. People on ruby 2.2.4 will have to add `gem 'shopify_api', '< 4.3'` to their Gemfiles - so probably a note in README.md is in order as well

ashmaroli:
@mikeyhew, I have a test-app without explicitly mentioning `gem 'shopify-api'` in the gemfile.
I have both Ruby2.2.4 & Ruby2.3.1 installed locally, and individual command prompts mapped to use R2.2.x and R2.3.x
The app runs smoothly in either console, on released gem versions & on the edge versions of both shopify_app and rails.

So, I feel this is not an issue concerning 'shopify_app'

ashmaroli:
The README note could be added under the Troubleshooting section and the ToC updated accordingly.

kevinhughes27:
I think its okay without the readme note. They can check the other gemfiles to see how we configured it for a green build.

Issue #297, opened 2016-07-18
Ability to change the queue name for ActiveJobs (Aws::SQS::Errors::NonExistentQueue in SessionsController#callback)


Skaelv:
I'd like to be able to customise the queue names for `ScriptTagManagerJob` and `WebhooksManagerJob`. We are using conventions for our queue names and I'd rather not create a new queue named `default`. 
I was thinking of adding  the possibility to add `ShopifyApp.config` vars called `scripttags_manager_queue_name` and `webhook_manager_queue_name` (default to `default`) and adding the following lines to each jobs file : 

```ruby
# webhooks_manager_job.rb 
 queue_as : ShopifyApp.config.webhook_manager_queue_name
```

```ruby
# scripttags_manager_job.rb 
 queue_as : ShopifyApp.config.scripttags_manager_queue_name
```

@kevinhughes27 what do you think of this approach ?

PR #298, opened 2016-07-18
add ability to customise queue names

Skaelv:
- will add spec when I get confirmation that we consider adding this feature

### DIFF
--- a/lib/shopify_app/scripttags_manager_job.rb
+++ b/lib/shopify_app/scripttags_manager_job.rb
@@ -1,5 +1,10 @@
 module ShopifyApp
   class ScripttagsManagerJob < ActiveJob::Base
+
+    queue_as do
+      ShopifyApp.configuration.webhooks_manager_queue_name
+    end
+
     def perform(shop_domain:, shop_token:, scripttags:)
       ShopifyAPI::Session.temp(shop_domain, shop_token) do
         manager = ScripttagsManager.new(scripttags)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+7.0.12
+------
+* Add ability to override the ActiveJob queue names in initializer file
+
 7.0.11
 ------
 * Pass configured resources (like webhooks or scripttags) into the job rather than reading the config again. This allows for dynamically setting ShopifyApp config in a web request and having the job handle it correctly. This change does not affect the usage of webhooks or scripttags
--- a/lib/shopify_app/configuration.rb
+++ b/lib/shopify_app/configuration.rb
@@ -12,6 +12,10 @@ module ShopifyApp
     attr_accessor :webhooks
     attr_accessor :scripttags
 
+    # customise ActiveJob queue names
+    attr_accessor :scripttags_manager_queue_name
+    attr_accessor :webhooks_manager_queue_name
+
     # configure myshopify domain for local shopify development
     attr_accessor :myshopify_domain
 
@@ -25,6 +29,14 @@ module ShopifyApp
 
     def has_scripttags?
       scripttags.present?
+    end
+
+    def scripttags_manager_queue_name
+      @scripttags_manager_queue_name ||= :default
+    end
+
+    def webhooks_manager_queue_name
+      @webhooks_manager_queue_name ||= :default
     end
   end
 
--- a/lib/shopify_app/webhooks_manager_job.rb
+++ b/lib/shopify_app/webhooks_manager_job.rb
@@ -1,5 +1,10 @@
 module ShopifyApp
   class WebhooksManagerJob < ActiveJob::Base
+
+    queue_as do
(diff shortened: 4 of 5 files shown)


Skaelv:
https://github.com/Shopify/shopify_app/issues/297

lib/shopify_app/scripttags_manager.rb
```diff
@@ -3,10 +3,12 @@ class ScripttagsManager
     class CreationFailed < StandardError; end
 
     def self.queue(shop_domain, shop_token, scripttags)
-      ShopifyApp::ScripttagsManagerJob.perform_later(
-        shop_domain: shop_domain,
-        shop_token: shop_token,
-        scripttags: scripttags
+      ShopifyApp::ScripttagsManagerJob
+        .set(queue: ShopifyApp.configuration.webhooks_manager_queue_name)
```
kevinhughes27: this is to allow dynamic config right? I would be okay with not having this part of the config be dynamic. I think the risk of someone being surprised by this would be low as in general I doubt many will use this setting.

kevinhughes27:
I am okay with this 👍

lib/shopify_app/scripttags_manager.rb
Skaelv: I don't actually need dynamic config but I had an issue trying to implement it like described in https://github.com/Shopify/shopify_app/issues/297.
 I'm setting the `config.webhooks_manager_queue_name` value in the `initializers/shopify_app.rb`. The issue is that it is still `nil` so it uses the default value when the app loads `webhooks_manager_job.rb` file.

kevinhughes27: Did you try `queue_as do` using the block?

Skaelv:
The `queue as do` block seems to do the trick. 
Ready to merge ?

README.md
```diff
@@ -190,6 +190,16 @@ end
...
+Customising Queue Names
+-----------------------------
+By default ActiveJob is using `:default` as queue names. You can override this value in the initializer :
+```ruby
+ShopifyApp.configure do |config|
+  config.scripttags_manager_queue_name = :'my-custom-name-worker'
+  config.webhooks_manager_queue_name = :'another-custom-name-worker'
+```
+
+
```
kevinhughes27: I'd prefer to not call this out in the readme as its pretty specific. The readme is already pretty big - maybe we need a wiki or something else on this repo now.

I think this feature is easy enough for someone to discover if they need it by reading the source

kevinhughes27:
Can you add a line to the Changelog

Skaelv:
done

lib/shopify_app/scripttags_manager_job.rb
```diff
@@ -1,5 +1,10 @@
 module ShopifyApp
   class ScripttagsManagerJob < ActiveJob::Base
+
+    queue_as do
+        ShopifyApp.configuration.webhooks_manager_queue_name
```
kevinhughes27: indented too far

kevinhughes27:
Thanks!

buggy:
@kevinhughes27 Won't this hard code the queue to `:default` unless it's overridden?

With Rails you can change the name of the default queue using the `config.active_job.queue_name` setting in `application.rb` (see section 5 of http://guides.rubyonrails.org/active_job_basics.html). From memory this setting only applies if you don't specify a queue name inside the job.

I'm probably not the only developer relying on this config setting so that each applications has it's own queue. If this patch does what I think it will, ignore `config.active_job.queue_name` and set the queue to `:default`, then there's going to be absolute chaos when all my applications and workers start using the same queue.

@Skaelv If you're only doing this to avoid using a queue named `default` and you don't need to have separate queues for the web hook/script tag managers from the rest of the application then can't you set `config.active_job.queue_name` in `application.rb`?

kevinhughes27:
@buggy good point. We should default to `config.active_job.queue_name` instead of queue name. Do you want to submit a PR?

Skaelv:
@buggy is right I can just change my `config.active_job.queue_name` to get the same result (I'm using only one SQS queue).
@kevinhughes27 we can revert this change or change the default to `config.active_job.queue_name`. This would allow people to have different queues (not sure if it's a common case).

kevinhughes27:
I think its still valid to put the Shopify init jobs into their own queue and I would like to enable that so I like the setting you added. We just need to update the default value to match the Rails config.

buggy:
@kevinhughes27 I'm in favor of reverting the patch instead of complicating shopify_app. The main use case for separate queues is to prevent the init jobs being stuck in a backlog. For most app developers that will never be a problem. If it does become a problem they can easily move their own jobs into a lower priority queue leaving the default queue for everything else.

kevinhughes27:
Its not really adding any complexity though. Most developers will never open that class and we haven't added any major documentation about it. All this really does is enable someone to tweak this without monkey patching

PR #299, opened 2016-07-19
do not override active_job queue_name in initializer

Skaelv:
https://github.com/Shopify/shopify_app/pull/298

### DIFF
--- a/lib/shopify_app/configuration.rb
+++ b/lib/shopify_app/configuration.rb
@@ -32,11 +32,11 @@ module ShopifyApp
     end
 
     def scripttags_manager_queue_name
-      @scripttags_manager_queue_name ||= :default
+      @scripttags_manager_queue_name ||= Rails.application.config.active_job.queue_name
     end
 
     def webhooks_manager_queue_name
-      @webhooks_manager_queue_name ||= :default
+      @webhooks_manager_queue_name ||= Rails.application.config.active_job.queue_name
     end
   end
 
--- a/test/shopify_app/configuration_test.rb
+++ b/test/shopify_app/configuration_test.rb
@@ -6,6 +6,10 @@ class ConfigurationTest < ActiveSupport::TestCase
     ShopifyApp.configuration = nil
   end
 
+  teardown do
+    Rails.application.config.active_job.queue_name = nil
+  end
+
   test "configure" do
     ShopifyApp.configure do |config|
       config.embedded_app = true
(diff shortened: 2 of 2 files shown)


lib/shopify_app/configuration.rb
```diff
@@ -32,11 +32,11 @@ def has_scripttags?
     end
 
     def scripttags_manager_queue_name
-      @scripttags_manager_queue_name ||= :default
+      @scripttags_manager_queue_name ||= Rails.application.config.active_job.queue_name || :default
     end
 
     def webhooks_manager_queue_name
-      @webhooks_manager_queue_name ||= :default
+      @webhooks_manager_queue_name ||= Rails.application.config.active_job.queue_name || :default
```
kevinhughes27: do we still need the || default? Can we just rely on Rails' default

Skaelv: Yep you are right. The `nil` check is done by activeJob.

buggy:
@Skaelv what happens if they haven't set `Rails.application.config.active_job.queue_name`? You probably need to use override queue -> Rails.application.config.active_job.queue_name -> :default.

kevinhughes27:
Missing a test update for the Configuration class

Skaelv:
It fails because `config.active_job.queue_name` is `nil` but we know that activeJob is going to set `queue_as` to `:default` if the config is not set.
Should I remove the test or test that it's `nil` ?

kevinhughes27:
Can you set it in the test

Skaelv:
Not sure I understand.
A normal behavior here is that `ShopifyApp.configuration.webhooks_manager_queue_name = nil` so that ActiveJob override with `:default`. 
Such a test wouldn't make sense : 
```ruby

  test "webhooks_manager_queue_name and scripttags_manager_queue_name are nil if not configured and ActiveJob queue_name is nil" do
    assert_equal nil, ShopifyApp.configuration.webhooks_manager_queue_name
    assert_equal nil, ShopifyApp.configuration.scripttags_manager_queue_name
  end
``` 
So I just removed it.

kevinhughes27:
Sorry I mean stub the Rails config so the test reveals how we intend this to work

Skaelv:
updated to : 
```ruby
test "webhooks_manager_queue_name and scripttags_manager_queue_name are nil if not configured and ActiveJob queue_name is nil" do
    Rails.application.config.active_job.stubs(:queue_name).returns(nil)

    assert_equal nil, ShopifyApp.configuration.webhooks_manager_queue_name
    assert_equal nil, ShopifyApp.configuration.scripttags_manager_queue_name
end
```

test/shopify_app/configuration_test.rb
```diff
@@ -46,12 +50,22 @@ class ConfigurationTest < ActiveSupport::TestCase
...
-    assert_equal :default, ShopifyApp.configuration.scripttags_manager_queue_name
+  test "webhooks_manager_queue_name and scripttags_manager_queue_name are equal to ActiveJob queue_name if not configured" do
+    Rails.application.config.active_job.queue_name = :'custom-queue-name'
+
+    assert_equal :'custom-queue-name', ShopifyApp.configuration.webhooks_manager_queue_name
+    assert_equal :'custom-queue-name', ShopifyApp.configuration.scripttags_manager_queue_name
+  end
+
+  test "webhooks_manager_queue_name and scripttags_manager_queue_name are nil if not configured and ActiveJob queue_name is nil" do
+    Rails.application.config.active_job.stubs(:queue_name).returns(nil)
```
kevinhughes27: I'd prefer you set a value like `default`

Skaelv: That's not going to describe the real behavior if we set the value to `:default` in the tests.
If we don't define `config.active_job.queue_name` it is going to be nil but it is going to be replaced by the `:default` value by activejob.

kevinhughes27: Thats my point though by the time this code runs Rails will have already done that and we will see the value of `default`

Skaelv: Ok it makes sense.

PR #305, opened 2016-07-20
only include login protection where required

kevinhughes27:
I recently made this change in some of my apps and think it belongs up stream. With this change ShopifyApp only modifies controllers it owns and it doesn't make anything any harder for the developer.

SessionsController and AuthenticatedController (both in the ShopifyApp namespace) require the module and therefore include it. Other controllers that inherit from ApplicationController don't have to worry about any name clash or added behaviour.

I ran an install flow 🎩  locally

For review:
@SebastianSzturo @alexcoco

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,7 @@
 7.0.12
 ------
+* Don't include ShopifyApp::LoginProtection in Application controller.
+  Include in Authenticate and Session Controller directly.
 * Add ability to override the ActiveJob queue names in initializer file
 
 7.0.11
--- a/app/controllers/shopify_app/authenticated_controller.rb
+++ b/app/controllers/shopify_app/authenticated_controller.rb
@@ -1,5 +1,6 @@
 module ShopifyApp
   class AuthenticatedController < ApplicationController
+    include ShopifyApp::LoginProtection
     before_action :login_again_if_different_shop
     around_action :shopify_session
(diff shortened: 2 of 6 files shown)


app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -1,5 +1,6 @@
 module ShopifyApp
   class SessionsController < ApplicationController
+    include ShopifyApp::LoginProtection
```
alexcoco: Does this need to go in the concern instead?

alexcoco:
Just one comment. Are there any tests that ensure the `LoginProtection` behaviour is still present in the `SessionsController`.

kevinhughes27:
Note - I should update the example app as well

app/controllers/shopify_app/sessions_controller.rb
kevinhughes27: It doesn't *need* to but it could. I think I like that idea. I will try it out

kevinhughes27:
@alexcoco yeah. Login protection includes a lot of the redirect logic required by the session concern so it is tested

alexcoco:
👍

SebastianSzturo:
I was wondering why this was on `ApplicationController`. Great change! 👍

mikeyhew:
👍  I always found it a bit weird that `shopify_app` needed to modify my ApplicationController

PR #306, opened 2016-07-21
add ability to delay webhook jobs

Skaelv:
Allows to do something like : 
```ruby 
   config.webhooks_queue_params: { wait_until: Date.tomorrow.noon }
end
```

or 

```ruby 
   config.webhooks_queue_params: { wait: 10.minutes }
end
```
See : https://github.com/rails/rails/tree/master/activejob

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 7.0.12
 ------
 * Add ability to override the ActiveJob queue names in initializer file
+* Add ability to delay webhook jobs
 
 7.0.11
 ------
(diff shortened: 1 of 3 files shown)


kevinhughes27:
what is the use case for this? why wouldn't you want to process as fast as possible?

Skaelv:
I need to associate a newly created cart with the intercom lead with an ajax call before I receive the webhook `cart/create`. 
Unfortunately your webhook delivery time is too fast :p 
It happens very often that I receive webhooks before the AJAX call is actually performed and I have a cart but no user to associate it with.

kevinhughes27:
There is a bunch of different ways you could solve this and you've mentioned 2 of them already. You could also look at more cleanly retrying the job (without an exception). It might also be possible to add some code directly to the job itself that always delays it rather than adding global config.

I feel pretty strongly that this doesn't belong in ShopifyApp. If you can't find another way to solve this issue then I would recommend coding this as a monkey patch in your application. If you think it would be useful to others you could release a gem that applies the patch but I don't think the number of people who would need this is very high which is why I don't want to include it in the gem.

PR #312, opened 2016-07-25
Add master-version badge to README

ashmaroli:
Add a **static** badge to denote current master-branch version and **requires: v7.1.0** to relevant areas in the README.
This is meant to prevent confusion till release of v7.1.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,12 @@
 Shopify App
 ===========
-[![Version][gem]][gem_url] [![Build Status](https://travis-ci.org/Shopify/shopify_app.png)](https://travis-ci.org/Shopify/shopify_app)
+[![Version][gem]][gem_url] [![Version][master]][repo] [![Build Status](https://travis-ci.org/Shopify/shopify_app.png)](https://travis-ci.org/Shopify/shopify_app)
 
 [gem]: https://img.shields.io/gem/v/shopify_app.svg
 [gem_url]: https://rubygems.org/gems/shopify_app
+[master]: https://img.shields.io/badge/master-v7.1.0-orange.svg
+[requires]: https://img.shields.io/badge/requires-v7.1.0-orange.svg
+[repo]: https://github.com/Shopify/shopify_app
 
 
 Shopify Application Rails engine and generator
@@ -21,7 +24,7 @@ Table of Contents
  * [Install Generator](#install-generator)
  * [Shop Model Generator](#shop-model-generator)
  * [Home Controller Generator](#home-controller-generator)
- * [App Proxy Controller Generator](#app-proxy-controller-generator)
+ * [App Proxy Controller Generator](#app-proxy-controller-generator--)  [![Requires v7.1.0][requires]](#)
  * [Controllers, Routes and Views](#controllers-routes-and-views)
 * [**Managing Api Keys**](#managing-api-keys)
 * [**WebhooksManager**](#webhooksmanager)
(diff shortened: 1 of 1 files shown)


kevinhughes27:
If I understand correctly this will have to be updated manually as master gets updated right? If thats the case I'd rather not introduce this. 

Not sure what the solution as it has happened a couple of times that users get confused by the latest documentation when the gem hasn't been released yet

ashmaroli:
> If I understand correctly this will have to be updated manually as master gets updated right?

Yes, you're correct. The master-version badge has to be updated manually. The badges that track Github repos currently in the wild work based on the Github release tags with semantic versioning. 

> Not sure what the solution as it has happened a couple of times that users get confused by the latest documentation when the gem hasn't been released yet

I hoped to minimize that using the `requires: v7.1.0` badge.
#313 brought in a cool change but did not include adding a note abut its usage in the readme. 
If included along with the **requires: badge**, it would inform users about both the presence of a new feature and the need to either use the latest master-branch or wait for a gem-release.

buggy:
@ashmaroli Won't telling people they need 7.1.0 be more confusing when they can't find it because it hasn't been released?

@kevinhughes27 Typically you would use a develop branch to avoid this problem. All PR's are merged to develop and develop is only merged to master when a release happens. Having separate master and develop branches also allows you to create bug fixes without being forced to include any new features that were merged since the last release (bugfix branches).

ashmaroli:
@buggy  You've got a valid point. I opted to add that badge instead of advising users to include the github repo to their gemfile. The gem badge should tell them that the latest release is only at 7.0.11 and the feature they're looking for is only available in '>= v7.1.0' . I was thinking of renaming that badge to `requires upcoming: v7.1.0` but dropped, as it was too long.   

By the way, I like that 'development' branch idea very much.

ashmaroli:
@kevinhughes27, is there a way to checkout a developer branch and revert _master_ back to the point of 7.0.11 release?

alexcoco:
> If I understand correctly this will have to be updated manually as master gets updated right? If thats the case I'd rather not introduce this.

Agreed. I'm not okay with adding something to the read my that needs to be updated every time a gem is cut. We try to automate as much as we can to avoid inconsistencies.

Additionally, the master branch of this gem does not reflect the code distributed in the gem, that's what we use tags for. Master is what @buggy is referring to as a develop branch and tags are being used as snapshots of cut gems.

buggy:
@alexcoco That's not what I'm saying

You should have both a develop and master branches. All pull requests are merged into develop which is your current development branch. When develop is ready for the next release then merge develop into master and tag the release on master.

It solves the README problem because master always contains the current release version. More importantly, it allows you to release bug fixes without being forced to also release any additional code that has been committed since the last release.

Read http://nvie.com/posts/a-successful-git-branching-model/ for more information.

alexcoco:
@buggy, yes I know what a develop branch is for but my point is that we don't need one because we are tagging releases. I'm simply saying that we use master in a way similar to what a develop branch would be used for. The result is that you should never depend on this gem's master branch. If you use a github gem dependency, you should be referencing the tag.

In any case, I don't think that we'll be switching to this model as we don't use it anywhere else internally.

kevinhughes27:
To echo @alexcoco this pattern of using master as develop and tagging releases is quiet common these days especially among modern languages like ruby and javascript. Adding a develop branch just adds complexity for very little gain. Everytime this happens I am surprised because this is how most other ruby gems work and I am pretty used to checking unreleased commits on other projects and then simply using the master branch until a release is cut.

We won't be switching to a different model. We can also just ship the gem quicker after a major feature is added to cut down on the likelihood of this happening.

PR #315, opened 2016-07-26
Fix #304 - Handle skip_before_action based on Rails version

### LINKED ISSUE #304: raise: false in WebhookVerification breaks in Rails 4
https://github.com/Shopify/shopify_app/blob/v7.0.11/lib/shopify_app/webhook_verification.rb#L6

`skip_before_filter :verify_authenticity_token` works but `skip_before_filter :verify_authenticity_token, raise: false` gives CSRF error

swalkinshaw:
Fixes a regression caused in 653942f which added `raise: false` for Rails 5 compatibility.

This does end up affecting Rails 4 apps as well so now we're explicitly handling this based on Rails version.

Another issue was that the tests weren't raising this error so the regression wasn't found. This also fixes the controller tests to ensure they will raise errors in the future.

@kevinhughes27

### DIFF
--- a/lib/shopify_app/app_proxy_verification.rb
+++ b/lib/shopify_app/app_proxy_verification.rb
@@ -3,7 +3,12 @@ module ShopifyApp
     extend ActiveSupport::Concern
 
     included do
-      skip_before_action :verify_authenticity_token, raise: false
+      if Rails.version >= '5.0'
+        skip_before_action :verify_authenticity_token, raise: false
+      else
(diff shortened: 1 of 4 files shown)


kevinhughes27:
👍

celsodantas:
using this gem with Rails 4.2.6 I'm getting "Can't verify CSRF token authenticity” when I get webhook requests. 

What is the regression that this is trying to fix?

kevinhughes27:
@celsodantas its fixed in `7.1.0`

PR #319, opened 2016-07-30
Updated Readme for engine use.

StephenKelly94:


README.md
```diff
@@ -170,6 +170,22 @@ This optional generator, not included with the default generator, creates the ap
...
+mount ShopifyApp::Engine, at: '/shopify'
+```
+
+To use named routes with the engine so that it can route between the custom app and the engine's routes it should be preficed with `main_app` or `shopify_app`.
+
+```ruby
+main_app.login_path # For a named login route on the rails app.
+
+shopify_app.login_path # For the shopify app store login route.
+```
```
ashmaroli: Hello @StephenKelly94, 
line 182:  it should be `prefixed` instead of `preficed`;
for lines 185 & 187, can you please explain how this is achieved by shopify_app gem and why including this in the readme would benefit new users of the gem?

StephenKelly94:
Yes, sorry, lack of checking my spelling and grammar when I was finished. Should be fine now, thanks.
I added line 185 & 187 because I think it's useful in some scenarios to know how to explicitly call named routes of the gem with the same name as the app. I ran into this kind of problem as a new user myself when creating a Shopify app that handles external users and thought it may benefit other new users likewise.

README.md
StephenKelly94: Hi @ashmaroli !
Yes, sorry, lack of checking my spelling and grammar when I was finished. Should be fine now, thanks.
I added line 185 & 187 because I think it's useful in some scenarios to know how to explicitly call named routes of the gem with the same name as the app. I ran into this kind of problem as a new user myself when creating a Shopify app that handles external users and thought it may benefit other new users likewise.

ashmaroli: While you could separate this entire section from the **Generators** section with a heading ( e.g. **Deploying to a Subdirectory** ), I still think `main_app.login_path` is not defined in the engine. There exists a `main_app.login_url` though, but it does not have an alias.

StephenKelly94: I agree it could possibly done in a separate section. However, it is part of what is generated and the engine was mentioned in this section so I felt it flowed naturally. You're correct there is no `main_app.login_path` in the engine. However, if you wanted to use a named route in a view rendered by the engine, you would need to prefix it with `main_app` and login_path was just my example of a conflicting named route.

ashmaroli: To me, this section is not related to routes _generated_ by the app. Its a custom scenario. So I'd prefer a separate section. And since, I do not have a specific use case in mind, it doesn't flow naturally, which resulted in all the confusion. 
So, my _suggestion_ ( for the first part of this PR ) is:

---
### Deploying to a Subdirectory
While the generators mount the the app engine at the root, by default, the `config/routes.rb` can be modified to mount the engine to any sub-directory (say `/app-name` ) as: 
```
mount ShopifyApp::Engine, at: "/app-name"
```
New consumers of your app will be automatically redirected to `/app-name/login` and `/app-name/<redirect_url>` after installing.

---

As for the named routes part, one of the admins will provide a better feedback.

StephenKelly94: I was hesitant to create a new section about this as it is not the biggest section. However, I have followed your advise to make it less confusing and have used some of your wording for the updated version. I have also put the named routes part under the `Deploying to Subdirectory` section, however, I left it as is.

kevinhughes27: +1 for a new section on `mounting`

README.md
```diff
@@ -170,6 +170,24 @@ This optional generator, not included with the default generator, creates the ap
 
 The last group of generators are for your convenience if you want to start overriding code included as part of the Rails engine. For example by default the engine provides a simple SessionController, if you run the `rails generate shopify_app:controllers` generator then this code gets copied out into your app so you can start adding to it. Routes and views follow the exact same pattern.
 
+Deploying to a Subdirectory
+-----------------
```
kevinhughes27: this line should be the same length as the text above it

ashmaroli:
Now that the _Deploying to..._ section has been green-lighted, @StephenKelly94, pls add to the  Table o' Contents as well.

Since, this PR is about adding to the README.......
@kevinhughes27 , The License section present in the gem's gh-page is missing from the README.

kevinhughes27:
There is a licence file in the repo itself isn't there

ashmaroli:
Ah.. right.. didnt see that..

StephenKelly94:
I have made the necessary changes to go in the Readme. Let me know if there is anything else! :)

ashmaroli:
To a rails-newbie, this section might raise the following questions:

- Do I include the named routes to `routes.rb` or `application_controller.rb` ?
- If `routes.rb`, then do I include the named routes before or after the `mount ...` statement
- How exactly is `main_app` route different from `shopify_app` route.

kevinhughes27:
I think this is fine. A Rails newbie probably wouldn't read that section nor be deploying a nested engine so I think this is fine.

README.md
```diff
@@ -170,6 +171,24 @@ This optional generator, not included with the default generator, creates the ap
 
 The last group of generators are for your convenience if you want to start overriding code included as part of the Rails engine. For example by default the engine provides a simple SessionController, if you run the `rails generate shopify_app:controllers` generator then this code gets copied out into your app so you can start adding to it. Routes and views follow the exact same pattern.
 
+Deploying to a Subdirectory
+---------------------------
```
kevinhughes27: can we call this section `Mounting the Engine` or something. I don't think deploying is the right word to use here

ashmaroli: Hows `Mounting the Engine at a Subdirectory` ? 
or the title of Issue 154 itself: `Running the app from a sub-directory` ?

@StephenKelly94, What may ever be your choice, please do update the ToC accordingly.

kevinhughes27: I prefer just `Mounting the Engine`. Its descriptive enough that users won't read the section unless they are trying to mount it in a different way and I think the title is descriptive enough to help users who are looking for sub directory mounting to find what they need

ashmaroli:
:+1:  for the new addition to the section

kevinhughes27:
This is great thanks!

PR #328, opened 2016-08-15
Require Ruby 2.2.2 instead of 2.2.4

swalkinshaw:
2.2.4 was a bit arbitrary and this matches up with Rails 5.0 better which also requires >= 2.2.2

We could tag this as 7.1.1

@kevinhughes27 

/cc @sshaw

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ gemfile:
 
 matrix:
   include:
-  - rvm: 2.2.4
+  - rvm: 2.2.2
     gemfile: Gemfile.ruby22
-  - rvm: 2.2.4
+  - rvm: 2.2.2
     gemfile: Gemfile.ruby22.rails50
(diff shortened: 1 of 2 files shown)


kevinhughes27:
👍

sshaw:
One important thing I forgot to mention. [This gem depends on Shopify API >= 4.2.2](https://github.com/Shopify/shopify_app/blob/c14049eb0ad4d7e8a775705b62d65d854dbd4edc/shopify_app.gemspec#L14). The Shopify API version after 4.2.2 (4.3.0) [requires Ruby >= 2.3](https://github.com/Shopify/shopify_api/blob/9a0fa26048ffe0779753e83a42372fadfb79693a/shopify_api.gemspec#L26).

So even with this change one may get an error when installing on a Ruby >= 2.2.2.

ashmaroli:
it was decided that users with Ruby < 2.3 should lock to shopify_api < v4.2.2 
ref: [ Issue Comment](https://github.com/Shopify/shopify_app/pull/296#issuecomment-234359716)

PR #331, opened 2016-08-19
Use more obvious method name

mariusbutuc:
For this generated example, we can use `#escape_javascript` instead of `#j`, as it's more obvious what the former does.

/review @kevinhughes27 @alexcoco please


kevinhughes27:
👍

alexcoco:
👌🏻

nickhoffman:
That was quick! 😃

PR #333, opened 2016-08-23
Bypass application layout rendering for SessionsController

mariusbutuc:
The SessionsController inherits rendering using the application layout.

By default, render with no layout at all.

/review @kevinhughes27 @alexcoco @swalkinshaw
/cc @nickhoffman

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+7.2.0
+-----
+* Disable application layout rendering for the `/login` page
+
 7.1.1
 -----
 * Lower required Ruby version to 2.2.2 to better match up with Rails 5.0
--- a/app/controllers/shopify_app/sessions_controller.rb
+++ b/app/controllers/shopify_app/sessions_controller.rb
@@ -1,6 +1,5 @@
 module ShopifyApp
   class SessionsController < ApplicationController
     include ShopifyApp::SessionsConcern
-    layout false
   end
 end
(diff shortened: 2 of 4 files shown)


alexcoco:
The new action should be rendered without a layout, why not update the session concern and set layout false for that action

nickhoffman:
@alexcoco, what's the benefit of specifying no layout for the `new` action only?

alexcoco:
The difference is that if someone inherits this controller then you're disabling it for all the actions even the ones that they add. I rather it be more explicit.

nickhoffman:
I usually err on the side of explicitness too. However, I figured that inheriting from this controller and adding actions is uncommon, and when it does occur, the application's default layout wouldn't be wanted. But if your experience says otherwise, :+1:

kevinhughes27:
I would prefer the explicitness as well

mariusbutuc:
Disabling the layout rendering moved into the concern, and it is now scoped to the `#new` action only, as of https://github.com/Shopify/shopify_app/commit/069d708e9f6e204d12b449d558cd4f6aef4d3434.

@kevinhughes27 @alexcoco @nickhoffman please have another look.

nickhoffman:
Code LGTM. :ship: once you bump the version and CI is green.

kevinhughes27:
👍

alexcoco:
👍

PR #341, opened 2016-09-14
Dynamic scripttags

sshaw:
Should probably add a test to make sure src returned by `#call` is not cached in the global list. 

Somewhat related question: under what circumstances [does this get called](https://github.com/Shopify/shopify_app/blob/076a2966823cfa2d74179cda675ff942de707c7a/lib/shopify_app/scripttags_manager.rb#L49) as if the API cannot create a resource an exception is raised.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,11 +268,14 @@ As with webhooks, ShopifyApp can manage your app's scripttags for you by setting
 ShopifyApp.configure do |config|
   config.scripttags = [
     {event:'onload', src: 'https://my-shopifyapp.herokuapp.com/fancy.js'}
+    {event:'onload', src: ->(domain) { dynamic_tag_url(domain) } }
   ]
 end
 ```
 
 Scripttags are created in the same way as the Webhooks, with a background job which will create the required scripttags.
+
+If `src` responds to `call` its return value will be used as the scripttag's source. It will be called on scripttag creation and deletion.
 
 ShopifyApp::SessionRepository
 -----------------------------
--- a/lib/shopify_app/scripttags_manager.rb
+++ b/lib/shopify_app/scripttags_manager.rb
@@ -6,14 +6,25 @@ module ShopifyApp
       ShopifyApp::ScripttagsManagerJob.perform_later(
         shop_domain: shop_domain,
         shop_token: shop_token,
-        scripttags: scripttags
+        # Procs cannot be serialized so we interpolate now, if necessary
+        scripttags: build_src(scripttags, shop_domain)
       )
     end
 
-    attr_reader :required_scripttags
+    def self.build_src(scripttags, domain)
+      scripttags.map do |tag|
+        next tag unless tag[:src].respond_to?(:call)
+        tag = tag.dup
+        tag[:src] = tag[:src].call(domain)
+        tag
+      end
+    end
 
-    def initialize(scripttags)
+    attr_reader :required_scripttags, :shop_domain
+
+    def initialize(scripttags, shop_domain)
       @required_scripttags = scripttags
+      @shop_domain = shop_domain
     end
 
     def recreate_scripttags!
@@ -24,14 +35,15 @@ module ShopifyApp
     def create_scripttags
       return unless required_scripttags.present?
 
-      required_scripttags.each do |scripttag|
+      expanded_scripttags.each do |scripttag|
         create_scripttag(scripttag) unless scripttag_exists?(scripttag[:src])
       end
     end
 
     def destroy_scripttags
-      ShopifyAPI::ScriptTag.all.each do |scripttag|
-        ShopifyAPI::ScriptTag.delete(scripttag.id) if is_required_scripttag?(scripttag)
+      scripttags = expanded_scripttags
+      ShopifyAPI::ScriptTag.all.each do |tag|
+        ShopifyAPI::ScriptTag.delete(tag.id) if is_required_scripttag?(scripttags, tag)
       end
 
       @current_scripttags = nil
@@ -39,8 +51,12 @@ module ShopifyApp
 
     private
 
-    def is_required_scripttag?(scripttag)
-      required_scripttags.map{ |w| w[:src] }.include? scripttag.src
+    def expanded_scripttags
+      self.class.build_src(required_scripttags, shop_domain)
+    end
+
+    def is_required_scripttag?(scripttags, tag)
+      scripttags.map{ |w| w[:src] }.include? tag.src
     end
 
     def create_scripttag(attributes)
--- a/test/shopify_app/scripttags_manager_test.rb
+++ b/test/shopify_app/scripttags_manager_test.rb
@@ -22,7 +22,7 @@ class ShopifyApp::ScripttagsManagerTest < ActiveSupport::TestCase
     @manager.create_scripttags
   end
 
-  test "#create_scripttags when creating a dynamic src, does overwrite the src with its result" do
+  test "#create_scripttags when creating a dynamic src, does not overwrite the src with its result" do
     ShopifyAPI::ScriptTag.stubs(all: [])
 
     stub_scripttag = stub(persisted?: true)
@@ -76,7 +76,7 @@ class ShopifyApp::ScripttagsManagerTest < ActiveSupport::TestCase
     @manager.destroy_scripttags
   end
 
-  test "#destroy_scripttags when deleting a dynamic src, does overwrite the src with its result" do
+  test "#destroy_scripttags when deleting a dynamic src, does not overwrite the src with its result" do
     ShopifyAPI::ScriptTag.stubs(all: Array.wrap(all_mock_scripttags.last))
     ShopifyAPI::ScriptTag.expects(:delete).with(all_mock_scripttags.last.id)
 
@@ -108,10 +108,6 @@ class ShopifyApp::ScripttagsManagerTest < ActiveSupport::TestCase
   def expect_scripttag_creation(event, src)
     stub_scripttag = stub(persisted?: true)
     ShopifyAPI::ScriptTag.expects(:create).with(event: event, src: src, format: 'json').returns(stub_scripttag)
-  end
-
-  def all_scripttag_srcs
-    @scripttags ||= ['https://example-app.com/fancy.js', 'https://example-app.com/foobar.js']
   end
 
   def all_mock_scripttags
--- a/test/shopify_app/scripttags_manager_test.rb
+++ b/test/shopify_app/scripttags_manager_test.rb
@@ -1,14 +1,16 @@
 require 'test_helper'
 
 class ShopifyApp::ScripttagsManagerTest < ActiveSupport::TestCase
+  include ActiveJob::TestHelper
 
   setup do
     @scripttags = [
       {event: 'onload', src: 'https://example-app.com/fancy.js'},
-      {event: 'onload', src: 'https://example-app.com/foobar.js'}
+      {event: 'onload', src: 'https://example-app.com/foobar.js'},
+      {event: 'onload', src: ->(domain) { "https://example-app.com/#{domain}-123.js" } }
     ]
 
-    @manager = ShopifyApp::ScripttagsManager.new(@scripttags)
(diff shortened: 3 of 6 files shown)


sshaw:
I also realized that `ScripttagsManager` only has access to the shop domain and token via `ShopifyAPI::Base.site.host`  and `ShopifyAPI::Base.headers`. While I do think it's a good idea to pass these to the method it felt a bit icky to get them from there. Thoughts?

test/shopify_app/scripttags_manager_test.rb
```diff
@@ -5,7 +5,8 @@ class ShopifyApp::ScripttagsManagerTest < ActiveSupport::TestCase
   setup do
     @scripttags = [
       {event: 'onload', src: 'https://example-app.com/fancy.js'},
-      {event: 'onload', src: 'https://example-app.com/foobar.js'}
+      {event: 'onload', src: 'https://example-app.com/foobar.js'},
+      {event: 'onload', src: ->() { 'https://example-app.com/dynamic.js' } }
```
kevinhughes27: What do you think about expanding this test case to be a bit more dynamic? I think it might help reveal the purpose of making these callable

lib/shopify_app/scripttags_manager.rb
```diff
@@ -25,6 +25,8 @@ def create_scripttags
       return unless required_scripttags.present?
 
       required_scripttags.each do |scripttag|
+        # src may be call()able and we don't want to overwrite it
+        scripttag = scripttag.merge(src: build_src(scripttag[:src]))
```
kevinhughes27: would this work instead?

```ruby
scripttag[:src] = build_src(scripttag[:src])
```

test/shopify_app/scripttags_manager_test.rb
sshaw: Agreeded.

lib/shopify_app/scripttags_manager.rb
sshaw: This will modify the callers Hash and, depending on how the manager is called, may overwrite the callable object in the global scripttag config.

sshaw:
@kevinhughes27, what are your thoughts on [this question](https://github.com/Shopify/shopify_app/pull/341#issuecomment-246891584).

kevinhughes27:
Have a look here: https://github.com/Shopify/shopify_app/blob/master/lib/shopify_app/sessions_concern.rb#L77

This information is already passed into the job which initializes the ScriptTagManager. You could add the shop name to the constructor

sshaw:
Checkout my latest updates.

lib/shopify_app/scripttags_manager.rb
```diff
@@ -12,8 +12,9 @@ def self.queue(shop_domain, shop_token, scripttags)
 
     attr_reader :required_scripttags
 
-    def initialize(scripttags)
+    def initialize(scripttags, shop_domain)
       @required_scripttags = scripttags
+      @shop_domain = shop_domain
```
kevinhughes27: can you add an `attr_reader` for this

test/shopify_app/scripttags_manager_test.rb
```diff
@@ -5,20 +5,32 @@ class ShopifyApp::ScripttagsManagerTest < ActiveSupport::TestCase
...
-    @manager = ShopifyApp::ScripttagsManager.new(@scripttags)
+    @manager = ShopifyApp::ScripttagsManager.new(@scripttags, 'example-app.com')
   end
 
   test "#create_scripttags makes calls to create scripttags" do
     ShopifyAPI::ScriptTag.stubs(all: [])
 
     expect_scripttag_creation('onload', 'https://example-app.com/fancy.js')
     expect_scripttag_creation('onload', 'https://example-app.com/foobar.js')
+    expect_scripttag_creation('onload', "https://example-app.com/example-app.com-123.js")
```
kevinhughes27: 👍

kevinhughes27:
@alexcoco can you have a look at this too when you have a chance!

alexcoco:
I'm away until next week, don't wait on me if this is urgent.

kevinhughes27:
No worries we'll check in next week or find another set of 👀

lib/shopify_app/scripttags_manager.rb
sshaw: Added.

lib/shopify_app/scripttags_manager.rb
```diff
@@ -57,5 +60,11 @@ def scripttag_exists?(src)
     def current_scripttags
       @current_scripttags ||= ShopifyAPI::ScriptTag.all.index_by(&:src)
     end
+
+    def build_src(src)
+      src.respond_to?(:call) ? src.call(@shop_domain) : src
```
kevinhughes27: this should use the attr_reader

sshaw: Oopsy, fixed now 🔧

kevinhughes27:
@gmalette can you also take a look here

sshaw:
Yeah I just remembered that `Proc`s cannot be serialized 🙀. Soooooo, what do you think about interpolating `src` beforehand? Something like this:

```ruby
    def self.queue(shop_domain, shop_token, scripttags)     
      ShopifyApp::ScripttagsManagerJob.perform_later(
        shop_domain: shop_domain,
        shop_token: shop_token,
        scripttags: build_src(scripttags)
      )
    end

    def self.build_src(scripttags, domain)
      scripttags.map do |tag|
        next tag unless tag[:src].respond_to?(:call)
        # We don't want to overwrite the original call()able src
        tag.merge(src: tag[:src].call(domain))
      end
    rescue => e
      raise CreationFailed, "cannot create script src: #{e.message}"
    end

    def initialize(scripttags, shop_domain)
      @required_scripttags = self.class.build_src(scripttags, shop_domain)
      @shop_domain = shop_domain
    end
```

Actually I just remembered there's a `required_scripttags` accessor, so one can always just add `Proc` after initialization which will still need to be interpolated.  I guess we'll just have to interpolate in `.queue` **and** `#create_scripttags`/`#destroy_scripttags`.

---

As an aside, Is there's a use case for having the accessor? (I'm not necessarily suggesting it be removed, just curious why it's there to begin with). I've found that most of the time adding accessors for instance variables needlessly exposes internals to callers, and prevents one from doing trivial things like the above, which also forces us to interpolate the src [a multitude of times](https://github.com/sshaw/shopify_app/blob/1f71de114a0ae1d5f9333f1642f6ace501917dae/lib/shopify_app/scripttags_manager.rb#L46).

kevinhughes27:
I tend to use the accessors to make it more clear that the instance variable is defined once and used multiple places - I see your point though this does expose it externally.

I am okay if you `build` the script tags before queuing the job. Can you change the exception to a new one though? Like `ScriptTagBuildFailed`

+1 for tests of the script tag job

lib/shopify_app/scripttags_manager.rb
alexcoco: > This will modify the callers Hash

Use `dup` instead of merging into itself since it shows your intention more clearly.

lib/shopify_app/scripttags_manager.rb
```diff
@@ -57,5 +60,11 @@ def scripttag_exists?(src)
     def current_scripttags
       @current_scripttags ||= ShopifyAPI::ScriptTag.all.index_by(&:src)
     end
+
+    def build_src(src)
+      src.respond_to?(:call) ? src.call(shop_domain) : src
+    rescue => e
+      raise CreationFailed, "cannot create script src: #{e.message}"
```
alexcoco: I think that we should let this method raise the original exception. It's not creation that failed here, it's some logic in the app that sent incorrect types or data. Thinly wrapping the _real_ exception in `CreationFailed` doesn't add much value and might make it more difficult to figure out what _actually_ happened.

sshaw: If we do it right then figuring out what happened shouldn't be a problem. All that's missing is the original backtrace, which can be solved via:

```ruby
raise CreationFailed, "cannot create script src: #{e.message}", e.backtrace
```

Providing the caller with consistent exception(s) has merit.

I made the change you've requested here, but this brings me back to an original exception related concern I had:

> under what circumstances [does this get called](https://github.com/Shopify/shopify_app/blob/076a2966823cfa2d74179cda675ff942de707c7a/lib/shopify_app/scripttags_manager.rb#L49)? If the API cannot create a resource an exception is raised.

lib/shopify_app/scripttags_manager.rb
```diff
@@ -24,23 +35,28 @@ def recreate_scripttags!
...
       end
     end
 
     def destroy_scripttags
-      ShopifyAPI::ScriptTag.all.each do |scripttag|
-        ShopifyAPI::ScriptTag.delete(scripttag.id) if is_required_scripttag?(scripttag)
+      scripttags = expanded_scripttags
+      ShopifyAPI::ScriptTag.all.each do |tag|
+        ShopifyAPI::ScriptTag.delete(tag.id) if is_required_scripttag?(scripttags, tag)
       end
```
sshaw: This was changed so that we don't build scripttag sources m*n times

alexcoco:
Wrapping it in this exception doesn't add much value, IMO. Ultimately I'll let @kevinhughes27 decide but I just don't see the value here.

lib/shopify_app/scripttags_manager.rb
kevinhughes27: I agree with @alexcoco here let's just raise the original. If this involved network or something I could see wrapping it since a lot of failures could come through here with no actionable changes but this is likely to be syntax error stuff imo.

sshaw: My last commit does just that. May be hard to tell with this new :octocat: code review stuff but it's there. 

Also I've asked this 2 time already but here goes \#3: Under what circumstances [does this get called](https://github.com/Shopify/shopify_app/blob/076a2966823cfa2d74179cda675ff942de707c7a/lib/shopify_app/scripttags_manager.rb#L49)? If the API cannot create a resource an exception is raised.

kevinhughes27: This API is called after the app is authenticated (which is why it has to be idempotent since it will run more than once so it's not an install hook). It is triggered in the omniauth callback here: https://github.com/Shopify/shopify_app/blob/master/lib/shopify_app/sessions_concern.rb#L18

README.md
```diff
@@ -268,12 +268,15 @@ As with webhooks, ShopifyApp can manage your app's scripttags for you by setting
...
     {event:'onload', src: 'https://my-shopifyapp.herokuapp.com/fancy.js'}
+    {event:'onload', src: ->(domain) { dynamic_tag_url(domain) } }
   ]
 end
 ```
 
 Scripttags are created in the same way as the Webhooks, with a background job which will create the required scripttags.
 
+If `src` responds to `call` its return value will be used as the scripttag's source. It will be called on scripttag creation and deletion.
+
```
kevinhughes27: is it relevant to specify that it also gets called on deletion or is that an internal detail (we need to know this in order to delete the right one). I prefer keeping the readme as simple as possible.

lib/shopify_app/scripttags_manager.rb
kevinhughes27: Nice!

kevinhughes27:
I am good with this now 👍

kevinhughes27:
ping @gmalette

README.md
sshaw: The reason I added this is so implementors know the circumstances in which their code will be executed and take heed when performing operations that may impact performance, be non-idempotent, etc...

kevinhughes27: Okay I am okay to leave it then.

kevinhughes27:
Can you add to the CHANGELOG then I will merge

sshaw:
> Can you add to the CHANGELOG then I will merge

Me?

kevinhughes27:
Yes please!

sshaw:
Okay I updated the CHANGELOG -and fixed a typo in 2 test names. I usually wouldn't add the version (I'd leave that to you guys) but...

kevinhughes27:
Thanks!

PR #343, opened 2016-09-16
Upgrading Gems for compatibility

xmjw:
There seems to be a couple of issues with Rails and Shopify_App combatilbility in the example app. This seems to resolve the issue.

Prior to this I was getting a lot of weird errors about Hashie, and for some reason `application_name=` being undefined.

### DIFF
--- a/example/Gemfile
+++ b/example/Gemfile
@@ -1,10 +1,12 @@
 source 'https://rubygems.org'
 
+ruby '2.2.2'
+
 gem 'rails', '~> 4.2'
 gem 'sqlite3'
 
 gem 'dotenv-rails'
-gem 'shopify_app', '~> 7.2.0'
+gem 'shopify_app'
 
 gem 'sass-rails', '~> 5.0'
 gem 'uglifier', '>= 1.3.0'
--- a/example/Gemfile
+++ b/example/Gemfile
@@ -1,10 +1,10 @@
 source 'https://rubygems.org'
 
-gem 'rails', '4.2.5.1'
+gem 'rails', '~> 4.2'
 gem 'sqlite3'
 
 gem 'dotenv-rails'
-gem 'shopify_app', '~> 7.0.0'
+gem 'shopify_app', '~> 7.2.0'
 
 gem 'sass-rails', '~> 5.0'
 gem 'uglifier', '>= 1.3.0'
--- a/example/Gemfile
+++ b/example/Gemfile
@@ -1,7 +1,5 @@
 source 'https://rubygems.org'
 
-ruby '2.2.2'
-
 gem 'rails', '~> 4.2'
 gem 'sqlite3'
 


example/Gemfile
```diff
@@ -1,10 +1,10 @@
 source 'https://rubygems.org'
 
-gem 'rails', '4.2.5.1'
+gem 'rails', '~> 4.2'
 gem 'sqlite3'
 
 gem 'dotenv-rails'
-gem 'shopify_app', '~> 7.0.0'
+gem 'shopify_app', '~> 7.2.0'
```
ashmaroli: I think its best to not specify the ShopifyApp version at all.. let the example run with the latest release

xmjw: Yeah, I think that's a good point. I've removed this. But I've also added the minimum ruby version from the enclosing gem...

ashmaroli: :+1:  for removing the ShopifyApp version.
But I'm on the fence regarding the addition of Ruby version. When a Ruby version is specified, *Heroku* for example, uses only that version. Also should a user on a newer version, download the example app and start the local server, as is, there will be conflicts raised along the lines:
```
 You are using Ruby 2.3.1 but the Gemfile specified 2.2.2 
```

( I've nt tested this myself though )

xmjw: Yeah, I think this is fair. I've been playing with this all morning, and I think it can go. I tried out a few different versions and I think it's safe. (One aside though: because of SQLite, this can't be run on Heroku.)

ashmaroli: > because of SQLite, this can't be run on Heroku.

oh right... never looked at the example's `database.yml` before.. always assumed it to be PostgreSQL as well.. :+1:

kevinhughes27: Keeping the example up to date with this repo isn't super well thought out. We kind of just update it as we remember. I had some sort of issue trying to specify the latest version but I don't recall exactly what it was.

It might be better to create a generator for the ESDK example code and remove this folder. It is kind of nice to have the example in the repo though but maybe if we did this then we could just re-generate this folder periodically.

Just some ideas.

kevinhughes27:
Can I ask why you are updating the example - is it to play with some of the ESDK code in it?

xmjw:
@kevinhughes27 Sure, I was helping out a developer who wanted to build on the Shopify API, but was having trouble knowing where to start. So they'd pulled down this example hoping to learn from it. Some things were broken so I thought I'd fix them. I think this is a great thing to have, as the ability to just pull a repo, and stand up a sample app that just works is a really good developer experience.

Auto generating it would be nice, but I've seen this get gnarly in the past. And remembering to update is also a bit of a pain. Perhaps drop this example from this repo, and create a new one that can be set against a given version? That way it will always work out of the box. Slowly upgrade it over time to match the latest code... Perhaps even provide with a Heroku deploy button, something like that?

The easier it is to get started, the more apps people are going to build...

kevinhughes27:
For some background, the example used to be a separate repo and I moved it into this one. A lot of people were cloning the example rather than running the generator which meant they often got old code and then would open issues on the wrong repo which went un-noticed.

I don't think generating this code is too hard. The example is really just a vanilla run of the shopify app generators plus a handful of views with ESDK examples. We probably just need to copy over the views and add a controller to make this work.

If this works now then I am happy to merge this.

xmjw:
I've checked this with a clean checkout on a different machine and it seems fine to me.

kevinhughes27:
I am updating the example for the version 8 release.

PR #344, opened 2016-09-21
add troubleshooting hint about reinstallation problems when session isn't deleted

exterm:
When using the shopify_app gem naively, but according to documentation, the session in the app's session store will not be deleted when the app is uninstalled.

That leads to a situation where a re-installation of the app is not possible because the app still has an outdated session.

You can use the app/uninstalled hook though to delete the session and fix this problem. This PR documents that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,10 @@ $ bundle exec spring stop
 
 Run shopify_app generator again.
 
+### App installation fails with 'The page you’re looking for could not be found' if the app was installed before
+
+This issue can occur when the session (the model you set as `ShopifyApp::SessionRepository.storage`) isn't deleted when the user uninstalls your app. A possible fix for this is listening to the `app/uninstalled` webhook and deleting the corresponding session in the webhook handler.
+
 Testing an embedded app outside the Shopify admin
 -------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


exterm:
@kevinhughes27 @damianthe

kevinhughes27:
👍

jamiemtdwyer:
LGTM :+1:

orenmazor:
@kevinhughes27 why isn't this default behaviour? :(

kevinhughes27:
@orenmazor why don't you just add it to the generator :trollface:

orenmazor:
fuck this is is as bad as @camilo 's sleep jokes

camilo:
banners_and_alerts_and_webscale_vocabulary

buggy:
If the app was previously installed doesn't the new install simply find the old record in the database and update the session token? I can't see how deleting the session storage during the app/uninstall webhook helps with the reported error. It just makes the next install clean because you're not picking up settings from the previous install.

kevinhughes27:
Yes that is what happens with the database - technically the troubleshooting should reference the browser session which is not cleared which is the problem so the wording it a little off now that I look at it again

PR #346, opened 2016-09-29
 Update to correct url in webhook example

wnm:
Without specifying protocol (http/https) the webhook won't be created, and ShopifyAPI::Webhook will return an error


kevinhughes27:
nice catch!

PR #349, opened 2016-10-10
Use fixed-size `secure_compare` in `WebhookVerification#hmac_valid?`

mikeyhew:
`variable_size_secure_compare` exists to protect against timing attacks that could reveal the length of the inputs. Since the inputs are SHA256 hmacs, their length is already known and nothing is gained from using `variable_size_secure_compare`. Switching to `secure_compare`, which short circuits if the inputs' lengths are not the equal.


kevinhughes27:
👍 

cc @Hammadk (I think you fixed this originally)

Hammadk:
LGTM

PR #353, opened 2016-11-12
Update example app to use rails 5

eljojo:
Hi everyone,
I tried using the example app but couldn't get it to work. I thought I might as well just update it to Rails 5.

I simply created a new rails 5 app, installed the gem and copied controllers, views, migrations and everything that might be needed.

I had issues with API keys, so I couldn't try this out. Please 🎩  before merging.


kevinhughes27:
@eljojo did you copy the controllers or did you run the generator? Ideally someone goes through making an app with rails 5 and using the generator and then updates the example. 

The example is intended for people to look at but is not the source of correctness. The source of correctness is what the generator produces

eljojo:
@kevinhughes27 I copied the controllers.

I don't think I'll have time soon to fix this, but feel free to take over this PR and correct it :)

kevinhughes27:
@ShayneP someone from your team should try and look into ShopifyApp + Rails 5 in the new year

eljojo:
As a short term solution you might want to get rid of ``application_name`` in the `shopify_app` initializer, like [here](https://github.com/Shopify/shopify_app/pull/353/files#diff-497d9a4888bb7b5a9a5bab0f2ec43bf8L2).

I don't know if the method has been removed, but it prevents the example app from booting.
I was representing Shopify in a hackathon over the weekend, and it was highly frustrating for attendees that the example app wouldn't even boot.

ShayneP:
@kevinhughes27 I'll add it to our Todo :+1:

eljojo:
@kevinhughes27 feel free to close this PR if you (or someone else) can't take over it :)

ashmaroli:
IMHO, we should either host two separate examples: one for rails 4.2.x, another for rails 5, (or) drop support for rails 4 altogether..

Issue #356, opened 2016-11-27
App generates incorrect configuration if options include whitespace


ashmaroli:
### Environment:
**Rails Version**: 5.0.0.1
**Ruby**: ruby 2.3.1p112 (2016-04-26 revision 54768) [i386-mingw32]
**ShopifyApp Version**: 7.2.0

### Steps to reproduce bug:
```sh
rails new test-app
cd test-app

# ensure gem 'rails', '~> 5.0.0', '>= 5.0.0.1' in Gemfile
# add following to Gemfile:
#   gem 'shopify_app', '7.2.0'
#   gem 'activeresource', github: 'rails/activeresource'

bundle install
rails g shopify_app --api_key 123456 --secret 654321 --scope read_orders, read_products, write_products --application_name Test App
```

### Expected Result:
```ruby
ShopifyApp.configure do |config|
  config.application_name = "Test App"
  config.api_key = "123456"
  config.secret = "654321"
  config.scope = "read_orders, read_products, write_products"
  config.embedded_app = true
end
```

### Actual Result:
```ruby
ShopifyApp.configure do |config|
  config.application_name = "Test"
  config.api_key = "123456"
  config.secret = "654321"
  config.scope = "read_orders,"
  config.embedded_app = true
end
```

ashmaroli:
Confirmed by Travis in #357

buggy:
I don't think this is really a bug. It's pretty standard to enclose parameters in quotes when you need spaces in them.

`rails g shopify_app --api_key 123456 --secret 654321 --scope "read_orders, read_products, write_products" --application_name "Test App"`

ashmaroli:
@buggy It IS a bug and I have evidence to prove it. I opened the #357 just for this reason.
It tests against a non-quoted string containing whitespace and a double-quoted string containing whitespace.

If you check the Travis builds in that pull, you'll see that irrespective of the framework version, both tests fail on every count.
```
1) Failure:
InstallGeneratorTest#test_creates_the_ShopifyApp_initializer_with_args [/home/travis/build/Shopify/shopify_app/test/generators/install_generator_test.rb:27]:
Expected /config\.application_name\ =\ "Test\ Name"/ to match "ShopifyApp.configure do |config|\n\n  config.application_name = \"Test\"\n\n  config.api_key = \"key\"\n  config.secret = \"shhhhh\"\n  config.scope = \"read_orders,\"\n  config.embedded_app = true\nend\n".

2) Failure:
[...]

PR #357, opened 2016-11-27
test arguments containing whitespace

ashmaroli:
test  for #356
label: `Experimental`, `not-for-merge`


ashmaroli:
no longer required to be open..

PR #358, opened 2016-11-27
Patch install_generator to allow whitespace

ashmaroli:
A patch to address #356.

- [x] Update documentation informing users that they are able to optionally pass double-quoted options for `--application_name` but `--scope` takes a comma-space-delimitted array..

### DIFF
--- a/lib/generators/shopify_app/install/install_generator.rb
+++ b/lib/generators/shopify_app/install/install_generator.rb
@@ -7,17 +7,20 @@ module ShopifyApp
       include Rails::Generators::Migration
       source_root File.expand_path('../templates', __FILE__)
 
-      class_option :application_name, type: :string
+      class_option :application_name, type: :array
       class_option :api_key, type: :string, default: '<api_key>'
       class_option :secret, type: :string, default: '<secret>'
-      class_option :scope, type: :string, default: 'read_orders, read_products'
+      class_option :scope, type: :array, default: ['read_orders,', 'read_products']
       class_option :embedded, type: :string, default: 'true'
 
       def create_shopify_app_initializer
-        @application_name = options['application_name']
+        #
+        # enable users to provide options containing whitespace and strip them off 'double-quotes', if present.
+        #
+        @application_name = options['application_name'].join(' ').tr('"', '') if options['application_name']
         @api_key = options['api_key']
         @secret = options['secret']
-        @scope = options['scope']
+        @scope = options['scope'].join(' ').tr('"', '')
 
         template 'shopify_app.rb', 'config/initializers/shopify_app.rb'
       end
--- a/lib/generators/shopify_app/install/install_generator.rb
+++ b/lib/generators/shopify_app/install/install_generator.rb
@@ -7,17 +7,14 @@ module ShopifyApp
       include Rails::Generators::Migration
       source_root File.expand_path('../templates', __FILE__)
 
-      class_option :application_name, type: :array
+      class_option :application_name, type: :array, default: ['My', 'Shopify', 'App']
       class_option :api_key, type: :string, default: '<api_key>'
       class_option :secret, type: :string, default: '<secret>'
       class_option :scope, type: :array, default: ['read_orders,', 'read_products']
       class_option :embedded, type: :string, default: 'true'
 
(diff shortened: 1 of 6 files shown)


kevinhughes27:
NVM apparently array argument fixes it. LGTM. Ping when you've updated the documentation and I will merge.

ashmaroli:
@kevinhughes27 Ready for re-review..

ashmaroli:
Not sure if this is the right place to state this, but I think the CHANGELOG should have 7.3.0 acc. to semver as the change by #341 was a non-breaking new-feature ( ? ). Ignore, if otherwise. :smiley:

lib/generators/shopify_app/install/install_generator.rb
```diff
@@ -7,17 +7,20 @@ class InstallGenerator < Rails::Generators::Base
...
-      class_option :scope, type: :string, default: 'read_orders, read_products'
+      class_option :scope, type: :array, default: ['read_orders,', 'read_products']
       class_option :embedded, type: :string, default: 'true'
 
       def create_shopify_app_initializer
-        @application_name = options['application_name']
+        #
+        # enable users to provide options containing whitespace and strip them off 'double-quotes', if present.
+        #
+        @application_name = options['application_name'].join(' ').tr('"', '') if options['application_name']
```
kevinhughes27: I don't think the comment is necessary especially since a test covers this. 

Also lets set a default for application name instead of adding this if check. Could just be `MyApp`

ashmaroli: > I don't think the comment is necessary especially since a test covers this.

okay.. will remove it..


> Also lets set a default for application name instead of adding this if check. Could just be MyApp

Thats probably out of current pull's scope as it will also warrant a change in the assoc `.erb` template.. what say?

kevinhughes27: I think its a small enough change to add to this PR

ashmaroli: okay then..

ashmaroli: Done.. I opted for `My Shopify App`..
There is a slight loss of DRY now.. will add a private method if you think so too..

kevinhughes27: Might as well, Thanks!

ashmaroli: updated..

lib/generators/shopify_app/install/install_generator.rb
```diff
@@ -61,6 +61,10 @@ def mount_engine
       def embedded_app?
         options['embedded'] == 'true'
       end
+
+      def refined(array)
+        array.join(' ').tr('"', '')
+      end
```
kevinhughes27: I find this name a bit unclear - can you be explicit and call it `join_array_argument` or `format_array_argument`

lib/generators/shopify_app/install/install_generator.rb
```diff
@@ -7,17 +7,17 @@ class InstallGenerator < Rails::Generators::Base
...
+      class_option :scope, type: :array, default: ['read_orders,', 'read_products']
       class_option :embedded, type: :string, default: 'true'
 
       def create_shopify_app_initializer
-        @application_name = options['application_name']
+        @application_name = refined options['application_name']
         @api_key = options['api_key']
         @secret = options['secret']
-        @scope = options['scope']
+        @scope = refined options['scope']
```
kevinhughes27: can you please use the braces here

ashmaroli: braces? don't you mean brackets?
```ruby
@scope = format_array_argument(options['scope'])
```

kevinhughes27: woops sorry, yes brackets, that was not clear :p

ashmaroli: updated..

kevinhughes27:
Can you also remove the application name logic in this file: https://github.com/Shopify/shopify_app/blob/master/lib/generators/shopify_app/install/templates/embedded_app.html.erb#L6

PR #359, opened 2016-11-29
Handle empty webhook result sets

orenmazor:
sometimes the shopify api can return a nil to us. so lets make sure we dont try to call #each on nil.

@kevinhughes27


orenmazor:
This could/should probably be fixed on the shopify_api gem but… I had this one already opened ¯\_(ツ)_/¯

PR #362, opened 2016-12-08
Updated readme to include webhook filtering

jamiemtdwyer:
Added a section describing how to filter webhook data when configuring webhooks.

@kevinhughes27

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,9 +127,9 @@ $ rails generate shopify_app:install --api_key <your_api_key> --secret <your_app
 ```
 
 Other options include:
-* `application_name` - the name of your app, it can be supplied with or without double-quotes if a whitespace is present. (e.g. `--application_name Example App` or `--application_name "Example App"`)
-* `scope` - the Oauth access scope required for your app, eg **read_products, write_orders**. *Multiple options* need to be delimited by a comma-space, and can be supplied with or without double-quotes
-(e.g. `--scope read_products, write_orders, write_products` or `--scope "read_products, write_orders, write_products"`)
+* `application_name` - the name of your app, it can be supplied with or without double-quotes if a whitespace is present. (e.g. `--application_name Example App` or `--application_name "Example App"`)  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -127,9 +127,9 @@ $ rails generate shopify_app:install --api_key <your_api_key> --secret <your_app
 ```
 
 Other options include:
-* `application_name` - the name of your app, it can be supplied with or without double-quotes if a whitespace is present. (e.g. `--application_name Example App` or `--application_name "Example App"`)  
-* `scope` - the Oauth access scope required for your app, eg **read_products, write_orders**. *Multiple options* need to be delimited by a comma-space, and can be supplied with or without double-quotes  
-(e.g. `--scope read_products, write_orders, write_products` or `--scope "read_products, write_orders, write_products"`)
```
ashmaroli: The whitespaces at the end of ln#131 and #132 were deliberate. They were included to *force* the `(e.g. ` and `For more..` onto the following line when the markdown document is rendered. 
I'm okay with the deletion on ln#130.

README.md
```diff
@@ -236,6 +236,16 @@ When the oauth callback is completed successfully ShopifyApp will queue a backgr
 
 ShopifyApp also provides a WebhooksController that receives webhooks and queues a job based on the webhook url. For example if you register the webhook from above then all you need to do is create a job called `CartsUpdateJob`. The job will be queued with 2 params `shop_domain` and `webhook` which is the webhook body.
 
+If you are only interested in particular fields, you can optionally filter the data sent by Shopify by specifying the `fields` parameter in `config/webhooks`. Note that you will still receive a webhook request from Shopify every time the resource is updated, but only the specified fields will be sent.
+
```
kevinhughes27: can you confirm if we filter the webhook send or not? I think we might.

jamiemtdwyer: Yup, I tested this and the behaviour is as written

kevinhughes27:
:shipit:

jamiemtdwyer:
I'll fix the whitespace and then :shipit:. Need to figure out how to prevent Atom from removing trailing whitespace...

PR #366, opened 2016-12-15
Use postMessage to redirect to Shopify during authentication

Hammadk:
A recent change was proposed in [Chrome Canary 57.0.2933.0](https://github.com/WICG/interventions/issues/16) that will break embedded app authentication. 

### Proposed solution:
* Instead of using `window.location.origin` to redirect to the authentication URL, we should use ` window.parent.postMessage` (this logic is in `#fullpage_redirect_to`).

This PR won't be merged until extensively tested

Related refactor:
* This PR also replaces `#redirect_to_with_fallback` with the Rails method: `redirect_to`. `#redirect_to_with_fallback` doesn't work like we think it does. The status and location headers take priority over the inline response. The fallback only happens if the browser ignores the status and location headers -- which doesn't happen on any modern browser.
 
* This PR moves `#sanitized_shop_name` and `#sanitize_shop_param(params)` from `SessionsConcern` to `LoginProtection`. LoginProtection is already included `SessionController`.

Please see the following issue for reference: https://github.com/Shopify/shopify/issues/93680

@nwtn @peterjm @kevinhughes27 Please review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+7.2.2
+-----
+* Use postMessage to redirect parent iframe during authentication
+  https://github.com/Shopify/shopify_app/pull/366
+
 7.2.1
 -----
 * Add support for dynamically generating scripttag URLs
--- a/lib/shopify_app/login_protection.rb
+++ b/lib/shopify_app/login_protection.rb
@@ -55,47 +55,33 @@ module ShopifyApp
       shopify_app.login_url(params)
     end
 
-    def redirect_to_with_fallback(url)
-      url_json = url.to_json
-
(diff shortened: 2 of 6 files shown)


peterjm:
Requires the beta flag to be removed from this feature, but I'm sure you know that 👍

Hammadk:
@peterjm Done in https://github.com/Shopify/shopify/pull/95550 :)

Hammadk:
Updated and tested in the following browsers:
- [x] Canary Chrome  57.0.2973.0
- [x] Firefox 43.0.4
- [x] Safari 10.0.2
- [x] Chrome 55
- [x] IE11 on Windows 7

PR #373, opened 2017-01-20
Sanitizes the shop param to only call authenticate on valid input

dreamsInDigital:
This changes sessions#new to check if the shop parameter received is valid before calling authenticate. Currently we just check if `params[:shop].present?`, but this causes some strange redirect behaviour on unexpected shop values. 

If we call an app with an invalid shop value (ex. `shop=example.com`), it will pass that to authenticate. Since `sanitized_shop_name.present?` would be false in this case, it redirects to `return_address`. If the `return_address` calls sessions#new again, this results in circular redirects.

@kevinhughes27 @gauravmc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+7.2.4
+-----
+* Fix redirect issue by sanitizing shop name on sessions#new
+
 7.2.3
 -----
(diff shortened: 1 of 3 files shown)


gauravmc:
👍 nice fix

kevinhughes27:
Can you add a line to the changelog? Then I'll merge

PR #383, opened 2017-02-09
Add french locales for i18n

samirElm:
Added french translation used by i18n in locales. 
Let me know if there is anything else I can do.


kevinhughes27:
Thanks!

PR #385, opened 2017-02-13
Add support for online mode authentication (per user)

humancopy:
Adding support to the new online mode authentication. I didn't do tests cause I wanted to consult with you guys first as to how you'd like to do it for testing both online and offline mode apps.

### DIFF
--- a/lib/generators/shopify_app/shop_model/templates/db/migrate/create_shops.rb
+++ b/lib/generators/shopify_app/shop_model/templates/db/migrate/create_shops.rb
@@ -1,12 +1,14 @@
 class CreateShops < ActiveRecord::Migration
   def self.up
     create_table :shops  do |t|
-      t.string :shopify_domain, null: false
-      t.string :shopify_token, null: false
+      t.string  :shopify_domain, null: false
+      t.string  :shopify_token, null: false
+      t.integer :associated_user_id
+      t.text    :extra_json
       t.timestamps
     end
 
-    add_index :shops, :shopify_domain, unique: true
+    add_index :shops, [:shopify_domain, :associated_user_id], unique: true, name: 'shopify_user'
   end
 
   def self.down
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,5 @@ source "https://rubygems.org"
 
 # Specify your gem's dependencies in shopify_app.gemspec
 gemspec
+
+gem 'shopify_api', '>= 4.3.5'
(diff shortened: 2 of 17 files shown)


humancopy:
It's failing in Ruby 2.2 because it's using an older version of ShopifyAPI and the initialize method there expects 2 arguments, so passing the `extra` argument makes it crash.

lib/generators/shopify_app/shop_model/templates/db/migrate/create_shops.rb
```diff
@@ -1,12 +1,14 @@
 class CreateShops < ActiveRecord::Migration
   def self.up
     create_table :shops  do |t|
-      t.string :shopify_domain, null: false
-      t.string :shopify_token, null: false
+      t.string  :shopify_domain, null: false
+      t.string  :shopify_token, null: false
+      t.integer :associated_user_id
+      t.text    :extra_json
```
EiNSTeiN-: These 2 fields do not belong to `shop` because many users from the same shop may belong to a single shop. I think this should either be in a separate table, or simply kept in the user's session.

Is there a reason you needed to keep this in the db? If not, I would recommend to simply use the session storage.

EiNSTeiN-:
I'm open to merge this PR, but I think the duplication of shops in the `shops` table is an issue. Let me know if you're open to fixing this, otherwise I can take the PR from here (although it will probably take some weeks before I get around to this).

humancopy:
The reason I added it to the `shops` table is so more than 1 staff member can be connected at the same time to the app. Otherwise they will disconnect each other because the `shops` table as it is currently can only hold one shop entry at a time. I'm happy to change the code if you have another idea how to resolve this issue ;)

EiNSTeiN-:
> Otherwise they will disconnect each other because the shops table as it is currently can only hold one shop entry at a time.

I may be missing something but why would this happen? What part of the code would disconnect them? As far as I know they should be able to be connected at the same time and reference the same shop?

humancopy:
The problem is mainly [here](https://github.com/Shopify/shopify_app/blob/e38a0e2892269f804a3768a3dc9181d34ec3eb48/lib/shopify_app/session_storage.rb#L7-L9):

```ruby
shop = self.find_or_initialize_by(shopify_domain: session.url)
```

This will fetch a shop by it's name and will populate the `shopify_token` column. This will cause the row in the table to change, so can't have one for each staff member.

kevinhughes27:
Closing because we have minimal support for this in version 8. Some of this code might still be worth adding but it should be extracted into smaller PRs

Issue #398, opened 2017-03-29
Cannot redirect to charge confirmation from embedded app


nsweeting:
Hi there,

First time using embedded app stuff. So bare with me. When attempting to create a charge and do the redirect to the confirmation url, I end up with a blank page within the embedded app.

My controller code looks like this:

```ruby
  def create
    charge = current_shop.recurring_application_charges.new
    if charge.save
      fullpage_redirect_to(charge.confirmation_url)
    else
      #handle errors
    end
  end
```

I then end up with a blank page in the app. The iframe source looks like this (link details modified):

```html
        <!DOCTYPE html>
        <html lang="en">
          <head>
            <meta charset="utf-8" />
            <base target="_top">
            <title>Redirecting…</title>
            <script type="text/javascript">

              // If the current window is the 'parent', change the URL by setting location.href
              if (window.top == window.self) {
                window.top.location.href = "https://shop.myshopify.com/admin/charges/1/confirm_recurring_application_charge?signature=code_is_here";

              // If the current window is the 'child', change the parent's URL with postMessage
              } else {
                normalizedLink = document.createElement('a');
                normalizedLink.href = "https://shop.myshopify.com/admin/charges/1/confirm_recurring_application_charge?signature=code_is_here";

                data = JSON.stringify({
                  message: 'Shopify.API.remoteRedirect',
                  data: { location: normalizedLink.href }
                });
                window.parent.postMessage(data, "https://");
              }

            </script>
          </head>
          <body>
          </body>
        </html>
```

I also end up with the following error in the console:

```
Uncaught DOMException: Failed to execute 'postMessage' on 'Window': Invalid target origin 'https://' in a call to 'postMessage'.
    at https://my-app.herokuapp.com/subscriptions:23:31
```

Any help would be appreciated.

Thanks

PresleyWong:
I experience same issue here.  I tried use older fullpage_redirect_to definition as below and it works but thats not a solution...

`	def fullpage_redirect_to(url)
	    url_json = url.to_json
	    url_json_no_quotes = url_json.gsub(/\A"|"\Z/, '')

	    if ShopifyApp.configuration.embedded_app?
	      render inline: %Q(
	        <!DOCTYPE html>
	        <html lang="en">
	          <head>
	            <meta charset="utf-8" />
	            <base target="_top">
	            <title>Redirecting…</title>
	            <script type="text/javascript">
	              window.top.location.href = #{url_json};
	            </script>
	          </head>
	          <body>
[...]

PR #401, opened 2017-04-10
Update LoginProtection#fullpage_redirect_to to get shopify domain from session

Hammadk:
In https://github.com/Shopify/shopify_app/issues/398 we found that `fullpage_redirect_to` cannot be used when the `:shop` param is not present and the shop is already logged in. This PR solves this problem by grabbing the domain from the session.

At this point I am not sure if there is a better method of grabbing the domain. Another option is to`ShopifyApp::SessionRepository.retrieve(session[:shopify]).shopify_domain`

Sidenote: Our setup for `LoginProtectionTests` is pretty weird. I'd like to unit test `fullpage_redirect_to` but that's not straight forward.

### DIFF
--- a/lib/shopify_app/login_protection.rb
+++ b/lib/shopify_app/login_protection.rb
@@ -86,7 +86,7 @@ module ShopifyApp
                   message: 'Shopify.API.remoteRedirect',
                   data: { location: normalizedLink.href }
                 });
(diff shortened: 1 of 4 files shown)


Hammadk:
@kevinhughes27 can you please give this another look. The last two commits fixed leaky tests that were setting `ShopifyApp.configuration.embedded_app = false`.

PR #402, opened 2017-04-10
Package for release 7.2.6

Hammadk:
This PR updates just packages this gem for a [minor version bump](https://github.com/Shopify/shopify_app/blob/master/RELEASING). Closes https://github.com/Shopify/shopify_app/issues/398


Hammadk:
I found a couple of more flakey tests :/ I'll fix this before releasing this version

PR #404, opened 2017-04-13
Rails 5: ActionDispatch::Reloader#to_prepare is deprecated

djones:
When Rails 5.1 comes out, the use of `ActionDispatch::Reloader#to_prepare` here will cause the session initializer to break.

This fixes the following deprecation warning, while maintaining support for Rails 4 and below:

> DEPRECATION WARNING: to_prepare is deprecated and will be removed from Rails 5.1 (use ActiveSupport::Reloader.to_prepare instead)


kevinhughes27:
LGTM Thanks!

Issue #407, opened 2017-05-01
Rails 5.1 with latest gem using undefined method 'to_prepare' for ActionDispatch


mingersoll:
The gem package for shopify_app (7.2.6) doesn't include the latest merge request (https://github.com/Shopify/shopify_app/pull/404) which fixes an issue with the session initializer.

I'm trying to generate the shopify app locally using the instuctions in the quickstart.mdb and am getting the error below. I'm using Rails 5.1.0 on Linux Mint 18.1

I'll try going back to Rails 5.0 but the gem should probably be updated to include the latest fixes.

Thanks!

rails generate shopify_app --api_key=MYKEY --secret=MYSECRET
/home/mike/test-app/config/initializers/shopify_session_repository.rb:4:in `<top (required)>': **undefined method `to_prepare' for ActionDispatch::Reloader:Class**
Did you mean?  to_param (NoMethodError)
	from /home/mike/.rvm/gems/ruby-2.4.1/gems/activesupport-5.1.0/lib/active_support/dependencies.rb:286:in `load'
	from /home/mike/.rvm/gems/ruby-2.4.1/gems/activesupport-5.1.0/lib/active_support/dependencies.rb:286:in `block in load'
	from /home/mike/.rvm/gems/ruby-2.4.1/gems/activesupport-5.1.0/lib/active_support/dependencies.rb:258:in `load_dependency'
	from /home/mike/.rvm/gems/ruby-2.4.1/gems/activesupport-5.1.0/lib/active_support/dependencies.rb:286:in `load'
	from /home/mike/.rvm/gems/ruby-2.4.1/gems/railties-5.1.0/lib/rails/engine.rb:655:in `block in load_config_initializer'
	from /home/mike/.rvm/gems/ruby-2.4.1/gems/activesupport-5.1.0/lib/active_support/notifications.rb:168:in `instrument'
	from /home/mike/.rvm/gems/ruby-2.4.1/gems/railties-5.1.0/lib/rails/engine.rb:654:in `load_config_initializer'
	from /home/mike/.rvm/gems/ruby-2.4.1/gems/railties-5.1.0/lib/rails/engine.rb:612:in `block (2 levels) in <class:Engine>'
	from /home/mike/.rvm/gems/ruby-2.4.1/gems/railties-5.1.0/lib/rails/engine.rb:611:in `each'
[...]

PR #408, opened 2017-05-01
Packaging for release 7.2.7

Hammadk:
Closes https://github.com/Shopify/shopify_app/issues/407

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,5 @@ rvm:
   - 2.3.1
 
 gemfile:
-  - Gemfile
+  - Gemfile.rails40
   - Gemfile.rails50
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,3 @@ rvm:
 gemfile:
   - Gemfile
   - Gemfile.rails50
-
-matrix:
-  include:
-  - rvm: 2.2.2
-    gemfile: Gemfile.ruby22
-  - rvm: 2.2.2
-    gemfile: Gemfile.ruby22.rails50
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 7.2.7
 -----
+* Require `shopify_app` `>= 4.3.2`. This version relaxes the ruby version requirement from `>= 2.3.0` to `>= 2.0`
 * Rails 5: ActionDispatch::Reloader#to_prepare is deprecated
   [[#404]](https://github.com/Shopify/shopify_app/pull/404)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+7.2.7
+-----
+* Rails 5: ActionDispatch::Reloader#to_prepare is deprecated
+  [[#404]](https://github.com/Shopify/shopify_app/pull/404)
+
 7.2.6
 -----
 * Update LoginProtection#fullpage_redirect_to to get shopify domain from session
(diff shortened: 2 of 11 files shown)


Hammadk:
Rails 5 builds are failing with the following errors:
* `Gem Load Error is: undefined method 'alias_method_chain' for ActiveResource::Connection:Class`
* `NoMethodError: undefined method 'raise_in_transactional_callbacks=' for ActiveRecord::Base:Class`

kevinhughes27:
I think we can just remove the calls to `raise_in_transactional_callbacks` as its just in our example and test dummy. Not sure about the other issue - maybe bump the required version of the api gem?

kevinhughes27:
we might need to update the tests to use the new rails controller testing style of `params: { #the params }` instead of just keyword args.

Hammadk:
Yep, I'll think about how we can make those changes so they are backward compatible with Rails 4 tests

kevinhughes27:
We could also drop support for rails 4

SebastianSzturo:
@Hammadk Can I slip in #409 before we release 7.2.7? 😅

Hammadk:
@SebastianSzturo Sure! Your tests in your PR are failing at the moment. Some of the commits in this PR are trying to fix those failures -- so you'll have to duplicate some of this work if you want the tests to pass.

I recommend that we get this PR out, and then we can release your's right after

Hammadk:
@kevinhughes27 @SebastianSzturo: We had to choose between dropping support for Rails 4, or coming up with an alternate solution. I decided against dropping support for Rails 4 because that would require us to add a runtime dependency for `gem 'activeresource', github: 'rails/activeresource'`. That points to the master branch of `activeresource` on github and I feel that's a little heavy-handed.

Instead I made the following changes:
* Update our Rails 5 tests to work with Rails `5.0.2`. This version of Rails 5 still supports Rails 4 test syntax
* Change the `'shopify_api` version to `'>= 4.3.2'`. This [version relaxes the ruby requirement](https://github.com/Shopify/shopify_api/blob/master/CHANGELOG#L51-L53) to ruby `> 2.0`. That means we don't need extra tests for ruby `2.2.2` so I have removed those.

What do you think about this approach? When a new version of active resource is released, we should drop support for Rails 4.

kevinhughes27:
Sounds good to me! Thanks @Hammadk

kevinhughes27:
@rafaelfranca is there a reason the newer version of active resource hasn't shipped yet that you are aware of?

rafaelfranca:
Mostly because I was trying to come up with a version of Active Resource that supports both Rails 4 and Rails 5. It is ready since a few months ago but now it requires more work. I can work on it in the next days.

kevinhughes27:
Sweet! We get a fair number of people posting issues here who miss the detail in the readme about using the master branch of active resource for rails 5 so it would definitely help to get a release!

rafaelfranca:
Can you test the master branch? It is ready for release, I just need to test Shopify on it.

kevinhughes27:
In Shopify or here? I think our rails 5 gemfile already uses it https://github.com/Shopify/shopify_app/blob/master/Gemfile.rails50#L5

rafaelfranca:
Cool, so we just need to trigger a new build. Let me do that

rafaelfranca:
5.0.0 released

Hammadk:
Thanks, Rafael! I spoke with Kevin and we are going to:
* Release this PR as `v7.2.7`
* Add Sebastian's PR and release that as `v7.2.8`

After that, we'll create another PR
* Drop support for Rails 4
* Require ActiveResource 5.0
* Update the test syntax 
* Release as `v7.3.0`

PR #417, opened 2017-05-28
Remove support for rails 4

Hammadk:
We need to remove support for Rails 4 because changes to the testing syntax In Rails 5 are not compatible with this version.

This PR makes the following changes:
* Update Travis CI to only test Rails 5
* Update test syntax to match Rails 5
* Clean up tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,3 @@ sudo: false
 
 rvm:
   - 2.3.1
-
-gemfile:
-  - Gemfile
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,4 @@ rvm:
   - 2.3.1
 
 gemfile:
-  - Gemfile.rails40
-  - Gemfile.rails50
+  - Gemfile
--- a/Gemfile.rails40
+++ b/Gemfile.rails40
@@ -1,4 +1,0 @@
-source 'https://rubygems.org'
-gemspec
-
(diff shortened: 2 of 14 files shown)


Hammadk:
Sub @swalkinshaw for @kevinhughes27 Since Kevin is away. Scott please see the comments in [this](https://github.com/Shopify/shopify_app/pull/408) thread for context on why we are removing support for Rails 5).

.travis.yml
```diff
@@ -6,5 +6,4 @@ rvm:
   - 2.3.1
 
 gemfile:
-  - Gemfile.rails40
-  - Gemfile.rails50
+  - Gemfile
```
swalkinshaw: Is this even needed anymore?

swalkinshaw:
Oh, can you update https://github.com/Shopify/shopify_app#rails-5

.travis.yml
Hammadk: We didn't need that. Removed 👍

kevinhughes27:
Awesome! Thanks for doing this :)

ashmaroli:
Shouldn't this have been released as `v7.3.0` to avoid breaking apps with dependency on minor?
```ruby
gem 'shopify_app', '~> 7.2.0'
```

Hammadk:
Fair point Ashmaroli. We should have skipped to `7.3.0` instead of `7.2.9`.

On the bright side, bundler will complain and let them know if their app is using Rails < 5.

Issue #390, opened 2017-02-22
Do some action upon omniauth confirmation


uurcankaya:
Hi everyone,

What is the best practice to enable some action right after customer installs the app?

There is a line called

 `mount ShopifyApp::Engine`

in routes file, i am assuming that's the code which adds all omniauth redirections. How can i override this? basically we need something like

`get 'auth/shopify/callback' => 'some_controller#some_method' `

uurcankaya:
i guess i am wrong, redirection happens in the session controller but not sure how to override that as well

kevinhughes27:
You can re-define the session controller in your app and do whatever you need to do. If you need to do something *after install* I would probably do it after the shop model is created rather than messing with the session controller.

This is the code in question btw: https://github.com/Shopify/shopify_app/blob/master/lib/shopify_app/sessions_concern.rb#L22

There was some talk in the past of adding a generic after install hook there that could be configured.

kevinzen:
I'm looking to solve this same issue -- setting up webhooks and customer accounts after a shop installs the app. What is the 'recommended' approach for this for someone using `shopify_app`? There doesn't seem to be a clean way to connect into the process directly.

kevinhughes27:
Webhooks can be managed using the provided class. Check out this section of the readme: https://github.com/Shopify/shopify_app#webhooksmanager

As for other actions I still think a generic after installed/authenticated hook for ShopifyApp would be useful. It would be pretty straight forward to add if you or someone is interested in doing it. It would need to hook in here: https://github.com/Shopify/shopify_app/blob/master/lib/shopify_app/sessions_concern.rb#L21-L22. I think ideally in after authenticated hooks would be configured in the initializer and likely just dispatched to a class controlled by the main application.

Dragas:
+1 on this. There's no elegant way to add assets (for example liquid templates) to shop using this framework. There's also no clear way how `WebHooksManager` and `ScriptTagManager` work, as they're written in WET manner. Instead you have to resort to hacks by rewriting half of `SessionConcerns` module regarding authentication callback and shop session login.

For example:
```ruby
module ShopifyApp
  class Configuration
    attr_accessor :assets

    def has_assets?
      assets.present?
    end

    def assets_manager_queue_name
      @assets_manager_queue_name ||= Rails.application.config.active_job.queue_name
    end
  end

  module SessionsConcern

    def callback
      if auth_hash
        login_shop
        install_webhooks
        install_scripttags
        install_assets
[...]

PR #428, opened 2017-07-07
Add hooks into installation lifecycle

alanhill:
### What?

Adds the ability to hook into the lifecycle of the installation process and fire off jobs after the app has installed as well as authenticated.

Closes https://github.com/Shopify/shopify_app/issues/390

### How?

- adds config options for `enable_after_authenticate_actions` and `enable_after_install_actions` 
  - when used with the generator `rails g shopify_app:enable_after_authenticate_actions` and `rails g shopify_app:enable_after_install_actions` will:
    - set the respective config to `true` and generate a `jobs/shopify/after_install_job.rb` or `jobs/shopify/after_authenticate_job.rb` 
- if enabled (the config = true), after the app installs/authenticates we then enqueue those jobs

@kevinhughes27 took a first pass at this, let me know what you think. The naming might need some work and I haven't added the options into the default generator yet `shopify_app:install --enable_after_authenticate_actions` for example.


lib/shopify_app/sessions_concern.rb
```diff
@@ -49,6 +50,8 @@ def login_shop
       sess = ShopifyAPI::Session.new(shop_name, token)
       session[:shopify] = ShopifyApp::SessionRepository.store(sess)
       session[:shopify_domain] = shop_name
+
+      perform_after_install_actions
```
kevinhughes27: I'm pretty sure this gets called even if the shop is logging in for a second time. The only way to do a true `after_install` is to check if the is creating the shop record but this assumes certain things about the SessionRepository. My opinion has always been that its better to keep it simple and use after login hooks but ensure that any actions performed are idempotent. This is how the WebhooksManager and ScriptTagsManager currently function.

All that is to say - lets just have the one `perform_after_authenticate_actions`

lib/shopify_app/sessions_concern.rb
```diff
@@ -87,5 +90,16 @@ def return_address
       session.delete(:return_to) || main_app.root_url
     end
 
+    def perform_after_install_actions
+      return unless ShopifyApp.configuration.enable_after_install_actions
```
kevinhughes27: instead of setting this configuration to `true` or `false` can we default it to `false` and set it to a job class when it should be enabled?

then this would look like:

```ruby
def perform_after_install_actions
  return unless ShopifyApp.configuration.after_install_job
  ShopifyApp.configuration.after_install_job.perform_later(shop_domain: session[:shopify_domain])
end
```

I like that you added a generator for the Job though we should keep that even though this connection point will be truly generic.

kevinhughes27: This is actually pretty similar to how we set the object for the SessionRepository https://github.com/Shopify/shopify_app/blob/master/example/config/initializers/shopify_session_repository.rb

Come to think of it why isn't the SessionRepository configured in the main ShopifyApp initializer? That would be a good follow up :)

kevinhughes27:
Nice work here 💖

kevinhughes27:
LGTM ping @Hammadk

alanhill:
I made so the job class is able to be changed, removed the `after_install` and went with just the `after_authenticate`. The command is changed to just `shopify_app:add_after_authenticate_job`

I was considering adding a command for when you're initializing the app, but I'm not sure if that's overkill - we could end up with a pretty massive list of flags for initializing an app if we keep adding to it. Thoughts?

This should be ready for re-review @kevinhughes27 


Sidenote:

It doesn't seem to overwrite a previous entry, for example, when it generates:
`config.after_authenticate_job = false` by default, if you then run `shopify_app:add_after_authenticate_job` you end up with:

```
  config.after_authenticate_job = Shopify::AfterAuthenticateJob
```

Will look into it, probably just missing something silly.

kevinhughes27:
I wouldn't add a flag to the base generator since this is more of an advanced feature.

kevinhughes27:
what would need to be inlined?

Hammadk:
Some apps do inline-configuration before they let the shop use the app. For example, checking if the shop is allowed to use the app.

Have a look at [Facebook](https://github.com/Shopify/facebook-commerce/blob/master/app/controllers/sessions_controller.rb#L7), [Messenger](https://github.com/Shopify/messenger-commerce/blob/master/app/controllers/sessions_controller.rb#L5) and [Pinterest](https://github.com/Shopify/pinterest-commerce/blob/master/app/controllers/sessions_controller.rb#L15-L18).

kevinhughes27:
ActiveJob supports `perform_now` so we could add an additional flag to this method that controls how the job is executed. @Hammadk I didn't look super hard but nothing from any of the ShopCreator objects look like it needs to be inlined. Most of what I saw was just queueing more jobs.

Checking if the shop is allowed to install should be a different hook added right at the start of the callback if we want to add/support something like that.

kevinhughes27:
yup! I'd go with `inline: false` as the default argument and then people ca override if they need to.

PR #430, opened 2017-07-10
Minor change: Expect params[:shop] to be a string

Hammadk:
`LoginProtection#login_again_if_different_shop` is a before action that forces re-login is another shop is already logged in. Currently it only works when `params[:shop].is_a?(String)`. As far as I can tell, there is no good reason to keep that around.

### DIFF
--- a/lib/shopify_app/login_protection.rb
+++ b/lib/shopify_app/login_protection.rb
@@ -27,7 +27,7 @@ module ShopifyApp
     end
 
     def login_again_if_different_shop
-      if shop_session && params[:shop] && params[:shop].is_a?(String) && shop_session.url != params[:shop]
+      if shop_session && params[:shop] && (shop_session.url != params[:shop])
         session[:shopify] = nil
         session[:shopify_domain] = nil
         redirect_to_login


lib/shopify_app/login_protection.rb
```diff
@@ -27,7 +27,7 @@ def shop_session
     end
 
     def login_again_if_different_shop
-      if shop_session && params[:shop] && params[:shop].is_a?(String) && shop_session.url != params[:shop]
+      if shop_session && params[:shop] && (shop_session.url != params[:shop])
```
kevinhughes27: agreed this seems weird to check for the type here. What do you think about abstracting this logic to a method `different_shop?`

Hammadk: I generally like extracting logic like this into well-named methods, but I don't think that we should do that in this case. The logic here is easy to understand because `#login_again_if_different_shop` gives us enough context.

I have been thinking about the `LoginProtection` module and I am a little worried about the methods that are exposed to the public API because they are `protected` and not `private`. I don't have a good solution for this yet but we should think about how to reduce the public API of this gem.

kevinhughes27: :ok: 

+1 for reducing the public interface. There is also some weird abstraction lines between this and some of the other concerns for legacy reasons.

Hammadk: Happy to help if you need! :)

PR #432, opened 2017-07-12
Add after authenticate job option

alanhill:
### What

Adds the ability to generate an after authenticate job which can perform async or inline once the shop has authenticated with the app.

Branch renamed from: https://github.com/Shopify/shopify_app/pull/428 to better reflect the changes. 

@Hammadk @kevinhughes27 ready for review again - made the config a hash accepting `job` and `inline`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+7.4.0
+-----
+* Add an after_authenticate job which will be run once the shop is authenticated. [[#431]](https://github.com/Shopify/shopify_app/pull/432)
+
 7.3.0
 -----
 * Bump required omniauth-shopify-oauth2 version to 1.2.0.
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Table of Contents
 * [**Managing Api Keys**](#managing-api-keys)
 * [**WebhooksManager**](#webhooksmanager)
 * [**ScripttagsManager**](#scripttagsmanager)
+* [**AfterAuthenticate Job**](#afterauthenticate-job)
 * [**ShopifyApp::SessionRepository**](#shopifyappsessionrepository)
 * [**AuthenticatedController**](#authenticatedcontroller)
 * [**AppProxyVerification**](#appproxyverification)
@@ -289,6 +290,31 @@ You also need to have write_script_tags permission in the config scope in order 
 Scripttags are created in the same way as the Webhooks, with a background job which will create the required scripttags.
 
 If `src` responds to `call` its return value will be used as the scripttag's source. It will be called on scripttag creation and deletion.
+
+AfterAuthenticate Job
+---------------------
+
+If your app needs to perform specific actions after it is installed ShopifyApp can queue or run a job of your choosing (note that we already provide support for automatically creating Webhooks and Scripttags). To configure the after authenticate job update your initializer as follows:
+
+```ruby
+ShopifyApp.configure do |config|
+  config.add_after_authenticate_job = { job: Shopify::AfterAuthenticateJob }
+end
+```
+
+If you need the job to run synchronously add the `inline` flag:
+
+```ruby
+ShopifyApp.configure do |config|
+  config.add_after_authenticate_job = { job: Shopify::AfterAuthenticateJob, inline: true }
+end
+```
+
+We've also provided a generator which creates a skeleton job and updates the initializer for you:
+
+```
+bin/rails g shopify_app:add_after_authenticate_job
+```
 
 ShopifyApp::SessionRepository
 -----------------------------
--- a/lib/generators/shopify_app/add_after_authenticate_job/add_after_authenticate_job_generator.rb
+++ b/lib/generators/shopify_app/add_after_authenticate_job/add_after_authenticate_job_generator.rb
@@ -1,0 +1,43 @@
+require 'rails/generators/base'
+
+module ShopifyApp
+  module Generators
+    class AddAfterAuthenticateJobGenerator < Rails::Generators::Base
+      source_root File.expand_path('../templates', __FILE__)
+
+      hook_for :test_framework, as: :job, in: :rails do |instance, generator|
+        instance.invoke generator, [ instance.send(:job_file_name) ]
+      end
+
+      def init_after_authenticate_config
+        initializer = load_initializer
+
+       after_authenticate_job_config = "  config.after_authenticate_job = { job: Shopify::AfterAuthenticateJob, inline: false }\n"
+
+        inject_into_file(
+          'config/initializers/shopify_app.rb',
+          after_authenticate_job_config,
+          before: 'end'
+        )
+
+        unless initializer.include?(after_authenticate_job_config)
+          shell.say "Error adding after_authneticate_job to config. Add this line manually: #{after_authenticate_job_config}", :red
+        end
+      end
+
+      def add_after_authenticate_job
+        template 'after_authenticate_job.rb', "app/jobs/#{job_file_name}_job.rb"
+      end
+
+      private
+
+      def load_initializer
+        File.read(File.join(destination_root, 'config/initializers/shopify_app.rb'))
+      end
+
+      def job_file_name
+        'shopify/after_authenticate'
+      end
+    end
+  end
+end
--- a/lib/shopify_app/sessions_concern.rb
+++ b/lib/shopify_app/sessions_concern.rb
@@ -20,6 +20,7 @@ module ShopifyApp
         login_shop
         install_webhooks
         install_scripttags
+        perform_after_authenticate_job
 
         redirect_to return_address
       else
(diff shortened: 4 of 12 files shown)


lib/generators/shopify_app/add_after_authenticate_job/add_after_authenticate_job_generator.rb
```diff
@@ -0,0 +1,38 @@
...
+    class AddAfterAuthenticateJobGenerator < Rails::Generators::Base
+      source_root File.expand_path('../templates', __FILE__)
+
+      hook_for :test_framework, as: :job, in: :rails do |instance, generator|
+        instance.invoke generator, [ instance.send(:job_file_name) ]
+      end
+
+      def init_after_authenticate_config
+        initializer = load_initializer
+        return if initializer.include?("config.webhooks")
```
kevinhughes27: why do you need to return if they have have webhooks configured?

lib/generators/shopify_app/add_after_authenticate_job/add_after_authenticate_job_generator.rb
```diff
@@ -0,0 +1,38 @@
...
+
+        inject_into_file(
+          'config/initializers/shopify_app.rb',
+          "  config.after_authenticate_job = { job: Shopify::AfterAuthenticateJob, inline: false }\n",
```
kevinhughes27: I would generate this without the inline arg since most people should use it as a job

kevinhughes27: oh I see - its not really a default argument the way I was thinking of it. Maybe it is better to have here since otherwise it would only live in documentation. What are your thoughts @Hammadk

lib/generators/shopify_app/add_after_authenticate_job/add_after_authenticate_job_generator.rb
alanhill: 🤦‍♂️ typo

lib/generators/shopify_app/add_after_authenticate_job/add_after_authenticate_job_generator.rb
Hammadk: I agree that we should keep it around for documentation purposes

lib/shopify_app/sessions_concern.rb
```diff
@@ -87,5 +88,16 @@ def return_address
       session.delete(:return_to) || main_app.root_url
     end
 
+    def perform_after_authenticate_job
+      return unless ShopifyApp.configuration.after_authenticate_job
+
+      job = ShopifyApp.configuration.after_authenticate_job[:job]
```
Hammadk: Is it possible for `ShopifyApp.configuration.after_authenticate_job` to be truthy, without setting `ShopifyApp.configuration.after_authenticate_job[:job]` ?

alanhill:
> Is it possible for ShopifyApp.configuration.after_authenticate_job to be truthy, without a job set in ShopifyApp.configuration.after_authenticate_job[:job]

Yes, I switched it up to check for the presence of the `config` in general, then `[:job]` and `[:inline]` otherwise it would fail on a nil class.

Also removed my typo and added the error handler like I originally intended. 

Ready for another quick round of 👀

lib/shopify_app/sessions_concern.rb
```diff
@@ -87,5 +88,16 @@ def return_address
       session.delete(:return_to) || main_app.root_url
     end
 
+    def perform_after_authenticate_job
+      config = ShopifyApp.configuration.after_authenticate_job
+
+      return unless config && config[:job].present? && config[:inline].present?
+
+      if config[:inline]
```
Hammadk: We should add tests for this conditional.

alanhill: I was pondering on this for a bit. There doesn't seem to be anything testing the authentication flow of the app - installing script tags and webhooks got lucky because they are contained in their own class instead of a job firing from within the app itself and not the gem, therefore you're able to unit test vs. integration test. 

I think the only way to properly tests this is to have an integration test which goes through the entire authentication flow, unless there's a simpler way that I'm missing here?

Hammadk: Have a look at the [`SessionControllerTests`](https://github.com/Shopify/shopify_app/blob/master/test/controllers/sessions_controller_test.rb#L97-L106). One way to do this is by testing the `#callback` method.

alanhill: 💡 ahhh that's what I was missing. TY.

alanhill:
Nice call on the test 😉 totally missed that one.

Should be good for review again!

Todo before :shipit: :

- [ ] Add to README
- [ ] Add to Changelog
- [ ] Bump version (minor I assume?)

.byebug_history
```diff
@@ -0,0 +1,5 @@
+exit
+config[:inline].present?
+config[:job].present?
+config
+next
```
kevinhughes27: 🔥

alanhill: 🤦‍♂️

lib/shopify_app/sessions_concern.rb
```diff
@@ -87,5 +88,16 @@ def return_address
       session.delete(:return_to) || main_app.root_url
     end
 
+    def perform_after_authenticate_job
+      config = ShopifyApp.configuration.after_authenticate_job
+
+      return unless config && config[:job].present? && config[:inline].in?([true, false])
+
+      if config[:inline]
```
Hammadk: Do we have a default value for `config[:inline]` ? If not, we won't get to this statement because most people won't set ` && config[:inline].in?([true, false])`. 

```
return unless config && config[:job].present?

if config[:inline] == true
...
```

test/controllers/sessions_controller_test.rb
```diff
@@ -149,6 +155,51 @@ class SessionsControllerTest < ActionController::TestCase
...
+        ShopifyApp.configure do |config|
+          config.after_authenticate_job = { job: Shopify::AfterAuthenticateJob, inline: nil }
+        end
+
+        Shopify::AfterAuthenticateJob.expects(:perform_later).never
+        Shopify::AfterAuthenticateJob.expects(:perform_now).never
+
+        mock_shopify_omniauth
+        get :callback, params: { shop: 'shop' }
+      end
```
Hammadk: Indentation is off here.

lib/shopify_app/sessions_concern.rb
alanhill: By default we add inline: false when you generate the job, but I see your point. So should we just consider anything but true to mean you want to perform asynchronously ?

Hammadk: Yeah, `config[:inline] == true` is good because it's self documenting code. If someone thinking about running an inline job, reading this code would tell them that they need to set that value to `true`.

alanhill:
This should be ready for review - comments are fixed up, README and Changelog added, version bumped a minor version.

README.md
```diff
@@ -23,6 +23,7 @@ Table of Contents
  * [Home Controller Generator](#home-controller-generator)
  * [App Proxy Controller Generator](#app-proxy-controller-generator)
  * [Controllers, Routes and Views](#controllers-routes-and-views)
+ * [After Authenticate Job and Generator](#after-authenticate-job-an-generator)
```
kevinhughes27: I would probably drop `Generator` from the section heading

kevinhughes27: Also I think this should go after the Webhooks and ScriptTags manager sections

README.md
```diff
@@ -323,6 +324,18 @@ end
 Create your app proxy url in the [Shopify Partners' Dashboard](https://app.shopify.com/services/partners/api_clients), making sure to point it to `https://your_app_website.com/app_proxy`.
 ![Creating an App Proxy](/images/app-proxy-screenshot.png)
 
+### After Authenticate Job and Generator
+
+You can generate an after authenticate job by running `shopify_app:add_after_authenticate_job` which will add a line in `shopify_app.rb` giving the option
+to run a job of your choosing either inline or asynchronously after the shop has authenticated. By default the job will run asynchronously unless `inline: false`
+is specified.
```
kevinhughes27: I would write this more like this:

If your app needs to perform specific actions after it is installed ShopifyApp can queue or run a job of your choosing (note that we already provide support for automatically creating Webhooks and Scripttags). To configure the after authenticate job update your initializer as follows:

```ruby
  config.add_after_authenticate_job = { job: Shopify::AfterAuthenticateJob }
end
```

If you need the job to run synchronously add the `inline` flag:

```ruby
  config.add_after_authenticate_job = { job: Shopify::AfterAuthenticateJob, inline: true }
end
```

We've also provided a generator which creates a skeleton job and updates the initializer for you:

```
bin/rails g shopify_app:add_after_authenticate_job
```

alanhill: 👍 Thanks for that, I'm awful with words.

alanhill:
README is updated with gooder ( 😉 ) words!

README.md
alanhill: Missed this one, updated!

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+7.3.1
```
kevinhughes27: I'd make this 7.4 since we added a feature

alanhill: 👍 bumped.

kevinhughes27:
🚢

PR #433, opened 2017-07-12
Package for release 7.2.12

ccleung:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+7.2.12
+-----
+* Bump required omniauth-shopify-oauth2 version to 1.2.0.
```
Hammadk: Can you please add the change from https://github.com/Shopify/shopify_app/pull/430 ?
Maybe `* Always expect params[:shop] to be a string`.

ccleung: updated

Hammadk:
@ccleung come to think of it, this should be v7.3 since this includes https://github.com/Shopify/shopify_app/pull/430. I think it's safe to assume that people are only passing the `:shop` params to authentication purposes, but there could be outliers.

ccleung:
@Hammadk i can close this and re-open a new PR with the proper branch name and version

PR #434, opened 2017-07-14
subclass from ActionController::Base not Application controller

kevinhughes27:
I'm working on mounting ShopifyApp in an application where most of the code is not part of the Shopify application. One of things I came across is that our ApplicationController defines a few before_actions which interfere with the engine. I think this change makes sense since we wouldn't want ApplicationController to interfere with anything inside the engine.

review
@Hammadk @alanhill

### DIFF
--- a/app/controllers/shopify_app/sessions_controller.rb
+++ b/app/controllers/shopify_app/sessions_controller.rb
@@ -1,5 +1,5 @@
 module ShopifyApp
-  class SessionsController < ApplicationController
+  class SessionsController < ActionController::Base
     include ShopifyApp::LoginProtection
     layout false, only: :new
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Removed the `shopify_session_repository` initializer. The SessionRepository is now configured through the main ShopifyApp configuration object and the generated initializer
 * Moved InMemorySessionStore into the ShopifyApp namespace
 * Remove ShopifySession concern. This module made the code internal to this engine harder to follow and we want do discourage over-writing the auth code now that we have generic hooks for all extra tasks during install.
+* Changed engine controllers to subclass ActionController::Base to avoid any possible conflict with the parent application
 
 7.3.0
 -----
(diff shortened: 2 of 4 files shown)


EiNSTeiN-:
How can you be sure no existing app rely on this to add before_actions to their app? If any app do rely on the current behaviour, how do they keep the existing behaviour?

kevinhughes27:
Thats a good point - there isn't really a good way to know if anyone is using this. Intuitively I doubt many people are since it would be a pretty complicated setup. 

In general if you need additions to your controllers you should subclass `ShopifyApp::AuthenticatedController` in your application and add things there. If you really need to add overrides to the controllers that are part of ShopifyApp you can override them entirely.

kevinhughes27:
We could do a bigger version bump to make people more aware of this change.

EiNSTeiN-:
Can you at least try this branch out on a couple of Shopify channels/apps? It should quickly be obvious if something break.

kevinhughes27:
@EiNSTeiN- yes we should definitely do that

app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -1,5 +1,5 @@
 module ShopifyApp
-  class SessionsController < ApplicationController
+  class SessionsController < ActionController::Base
     include ShopifyApp::SessionsConcern
```
Hammadk: We need to include `include ShopifyApp::LoginProtection` here no?

Hammadk: Actually, never mind. I just realized that `LoginProtection` is included in `SessionsConcern`

kevinhughes27: yeah I get tripped up by that too. Every time I always wonder if anyone is actually using this concern in their own controller or if we could simplify it and move this code into the controller.

kevinhughes27:
@swalkinshaw this one too!

kevinhughes27:
Merging into v8.0.0 branch. We are going to do a full QA after we're done with all the changes for the next major version.

PR #435, opened 2017-07-14
Add routes to Configuration

kevinhughes27:
Internally to ShopifyApp we attempted to determine how/where the engine was mounted and redirect accordingly. This code was complicated, likely didn't handle all edge cases and depended on Rails internals making it fragile. This doesn't affect the vast majority of consumers of this gem who likely mount the engine at the root as per the instructions. This only matters when the engine is not mounted at the root requiring additional but explicit configuration in the initializer eg:

```ruby
  config.root_url = '/nested'
  config.login_url = '/nested/login'
end

```

Review
@Hammadk @swalkinshaw


README.md
```diff
@@ -190,14 +190,6 @@ mount ShopifyApp::Engine, at: '/app-name'
...
 This will create the Shopify engine routes under the specified Subdirectory, as a result it will redirect new consumers to `/app-name/login` and following a similar format for the other engine routes.
 
-To use named routes with the engine so that it can route between the application and the engine's routes it should be prefixed with `main_app` or `shopify_app`.
-
-```ruby
-main_app.login_path # For a named login route on the rails app.
-
-shopify_app.login_path # For the shopify app store login route.
-```
-
```
kevinhughes27: I removed this from the Readme since I think this is more of an internal concern. I can't think of why this would be useful outside of the engine

lib/shopify_app/login_protection.rb
```diff
@@ -54,7 +54,7 @@ def close_session
     def main_or_engine_login_url(params = {})
       main_app.login_url(params)
     rescue NoMethodError
-      shopify_app.login_url(params)
+      shopify_app_engine.login_url(host: "https://#{request.host}", params: params)
```
Hammadk: Does this work in local environments without a port? For example, if my dev server is listening on port 3000 `myapp.com:3000` this would redirect to `myapp.com/app-name/login

README.md
swalkinshaw: I agree with this. If the engine is mounted then you just access routes as usual.

swalkinshaw:
Just wanted to summarize our chat about alternatives:

Before we were calling either `main_app.login_url` or falling back to `shopify_app.login_url`.

One alternative is the following:

1. Require that a named route `login_url` exists and raise if it doesn't.
2. Check and raise if *both* app and engine `login_url` exist. Either mount the engine or do it manually.
3. Only ever call the bare `login_url`.

README.md
```diff
@@ -183,28 +183,39 @@ Mounting the Engine will provide the basic routes to authenticating a shop with
...
-This will create the Shopify engine routes under the specified subpath, as a result it will redirect new consumers to `/nested/login`. If you mount the engine at a subpath you'll also need to update the omniauth initializer to include a custom `callback_path` e.g:
+or
 
 ```ruby
-  provider :shopify,
-    ShopifyApp.configuration.api_key,
-    ShopifyApp.configuration.secret,
-    scope: ShopifyApp.configuration.scope,
-    callback_path: '/nested/auth/shopify/callback'
+namespace :nested do
```
swalkinshaw: I'm not sure we should encourage or document this.

lib/shopify_app/login_protection.rb
```diff
@@ -41,20 +41,25 @@ def redirect_to_login
...
-    def main_or_engine_login_url(params = {})
-      main_app.login_url(params)
-    rescue NoMethodError
-      shopify_app.login_url(params)
+    def login_url
+      url = ShopifyApp.configuration.login_url
+
+      if params[:shop].present?
+        query = { shop: params[:shop] }.to_query
+        url = "#{url}?#{query}"
```
Hammadk: We'll lose all parameters except for the `:shop` param. Is that intentional?

kevinhughes27: this is how it worked before - we always re-created the params hash and it only included `shop`

README.md
kevinhughes27: 👍

lib/shopify_app/login_protection.rb
Hammadk: That sounds like a mistake no? Consider the case where someone is accessing the app through a [admin link](https://help.shopify.com/partners/making-apps/app-admin-access):

1. The link will send them to `app.com?shop=SHOP&action=ACTION`
2. If they are not logged in, they'll be redirected to `app.com/login?shop=SHOP`.
3. Once logged in, they'll probably be redirect to `app.com`

README.md
Hammadk: What do we think about changing that namespace to `shopify_app` ? I like that because it gives developers immediate context on what's being accessed.

lib/shopify_app/login_protection.rb
kevinhughes27: I think that case is covered by the `return_to`

Hammadk: I see! We set that [here](https://github.com/Shopify/shopify_app/blob/7a654313848c7c0a5e0ceceacde1969a1400b03a/lib/shopify_app/login_protection.rb#L43)

README.md
```diff
@@ -183,28 +183,31 @@ Mounting the Engine will provide the basic routes to authenticating a shop with
...
+If required the engine can be mounted at a nested route, eg:
 
 ```ruby
 mount ShopifyApp::Engine, at: '/nested'
 ```
 
-This will create the Shopify engine routes under the specified subpath, as a result it will redirect new consumers to `/nested/login`. If you mount the engine at a subpath you'll also need to update the omniauth initializer to include a custom `callback_path` e.g:
+This will create the Shopify engine routes under the specified subpath. You'll also need to make some updates to your `shopify_app.rb` and `omniauth.rb` initializers.
+
+update the shopify_app initializer to include a custom `root_url` and `login_url` e.g:
```
Hammadk: We should remove `login_url` from the documentation (here and in the lines below)

kevinhughes27: derp

swalkinshaw:
README still has a reference to `config.login_url = '/nested/login'`

kevinhughes27:
Missed another one - fixed up now!

PR #438, opened 2017-07-17
Configure SessionRepository in main initializer

kevinhughes27:
There is no reason to have a separate initializer. The reason it is done this way is for legacy reasons before ShopifyApp had its own configuration. While make this change I also found out that the InMemorySessionStore class wasn't in the ShopifyApp namespace so I moved it there. I also removed the `favicon` from our default generated view since it breaks the latest version of rails and isn't necessary.

@Hammadk @swalkinshaw


lib/shopify_app/configuration.rb
```diff
@@ -24,6 +25,18 @@ def initialize
       @myshopify_domain = 'myshopify.com'
     end
 
+    def session_repository=(klass)
+      if Rails.configuration.cache_classes
+        ShopifyApp::SessionRepository.storage = klass
+      else
+        reloader = defined?(ActiveSupport::Reloader) ? ActiveSupport::Reloader : ActionDispatch::Reloader
```
swalkinshaw: Can we get rid of this now? Gem relies on >= 5.0 so we can assume `ActiveSupport::Reloader`

lib/generators/shopify_app/install/templates/shopify_session_repository.rb
```diff
@@ -1,23 +0,0 @@
-# You should replace InMemorySessionStore with what you will be using
```
swalkinshaw: Only thing we lose is this documentation now. Is there any other place we could put it?

lib/shopify_app/configuration.rb
kevinhughes27: good call!

lib/generators/shopify_app/install/templates/shopify_session_repository.rb
kevinhughes27: Its fairly well documented in the Readme itself. I'll read through this and make sure its all captured

kevinhughes27: This is basically just describing the ShopifyApp::Shop concern which is part of the gem now so I think we are covered.

swalkinshaw:
I guess some tests just need references updated now.

README.md
```diff
@@ -293,9 +293,9 @@ If `src` responds to `call` its return value will be used as the scripttag's sou
 ShopifyApp::SessionRepository
 -----------------------------
 
-`ShopifyApp::SessionRepository` allows you as a developer to define how your sessions are retrieved and stored for a shop. The `SessionRepository` is configured using the `config/initializers/shopify_session_repository.rb` file and can be set to any object that implements `self.store(shopify_session)` which stores the session and returns a unique identifier and `self.retrieve(id)` which returns a `ShopifyAPI::Session` for the passed id. See either the `InMemorySessionStore` or the `SessionStorage` module for examples.
+`ShopifyApp::SessionRepository` allows you as a developer to define how your sessions are retrieved and stored for a shop. The `SessionRepository` is configured in the `config/initializers/shopify_app.rb` file and can be set to any object that implements `self.store(shopify_session)` which stores the session and returns a unique identifier and `self.retrieve(id)` which returns a `ShopifyAPI::Session` for the passed id. See either the `InMemorySessionStore` class or the `ShopifyApp::Shop` concern for examples.
```
Hammadk: Let's add the new namespace in this text: `ShopifyApp::InMemorySessionStore`

lib/shopify_app/in_memory_session_store.rb
```diff
@@ -1,25 +1,27 @@
...
+    def self.clear
+      @@repo = nil
+    end
 
-  def self.repo
-    if Rails.env.production?
-      raise EnvironmentError.new("Cannot use InMemorySessionStore in a Production environment")
+    def self.repo
+      if Rails.env.production?
+        raise EnvironmentError.new("Cannot use InMemorySessionStore in a Production environment")
```
Hammadk: I really like that we explicitly raise this error here. When people see this error during a deploy, they'll think about what they need to do to fix this. We can make this error more actionable by changing it to:

```
raise EnvironmentError.new(
  "Cannot use InMemorySessionStore in a Production environment . Please initialize ShopifyApp with a model that can store and retrieve sessions"
)
```
We can also add a link to relevant documentation in that error. Maybe a link to a wiki page.

PR #439, opened 2017-07-17
remove session concern

kevinhughes27:
As part of a push for simplification in 8.0.0 I propose that we remove the SessionsConcern. This was initially added to help with flexibility in defining your own session controller. With our current session controller having support for installing webhooks, scripttags and a generic after_auth hook I think we should discourage including these methods in your own controller. This controller can still be sub-classed and overwritten if desired.

The tests already only hit the controller so no changes are required to the test suite.

@Hammadk


kevinhughes27:
@swalkinshaw for a second review

PR #447, opened 2017-07-26
EmbeddedApp Concern

kevinhughes27:
One thing I didn't like when I went to use this engine on a project where the embedded app is a small thing compared to the whole app was how we still configured headers for the ESDK globally. I looked into how to avoid this because I didn't want to add these headers for my whole application. I found it was pretty easy to do locally in newer versions of rails (this may have been hard in the past I am not sure).

I also updated the Localization concern to fully abstract itself while I was nearby.

@swalkinshaw @Hammadk

### DIFF
--- a/lib/shopify_app/controller_concerns/embedded_app.rb
+++ b/lib/shopify_app/controller_concerns/embedded_app.rb
@@ -1,0 +1,19 @@
+module ShopifyApp
+  module EmbeddedApp
+    extend ActiveSupport::Concern
+
+    included do
+      if ShopifyApp.configuration.embedded_app?
+        after_action :set_esdk_headers
+        layout 'embedded_app'
+      end
+    end
+
+    private
+
+    def set_esdk_headers
+      response.set_header('P3P', 'CP="Not used"')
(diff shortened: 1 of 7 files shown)


lib/shopify_app/controller_concerns/embedded_app.rb
```diff
@@ -0,0 +1,19 @@
...
+        after_action :set_esdk_headers
+        layout 'embedded_app'
+      end
+    end
+
+    private
+
+    def set_esdk_headers
+      response.set_header('P3P', 'CP="Not used"')
+      response.default_headers.delete('X-Frame-Options')
```
Hammadk: This updates the headers on the response of the request. Weren't we previously setting headers on the request?

Hammadk: Eh, I thought about this a bit more and this should be safe to do since `X-Frame-Options` doesn't have anything to do with the server.

kevinhughes27: For what its worth I did try without the headers and the ESDK does not work so this is still required

PR #457, opened 2017-08-23
basic setup for per-user authentication

kevinhughes27:
related to #385

This PR adds some inner workings to store the current user if ShopifyApp is configured to use per user permissions. It doesn't add everything that we could possibly support for Online mode but its a start.

For review:
@Hammadk @EiNSTeiN-

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,22 @@ provider :shopify,
   callback_path: '/nested/auth/shopify/callback'
 ```
 
+Per User Authentication
+-----------------------
+To enable per user authentication you need to update the `omniauth.rb` initializer:
+
+```ruby
+provider :shopify,
+  ShopifyApp.configuration.api_key,
+  ShopifyApp.configuration.secret,
+  scope: ShopifyApp.configuration.scope,
+  per_user_permissions: true
+```
+
+The current Shopify user will be stored in the rails session at `session[:shopify_user]`
+
+This will change the type of token that Shopify returns and it will only be valid for a short time. Read more about `Online access` [here](https://help.shopify.com/api/getting-started/authentication/oauth). Note that this means you won't be able to use this token to respond to Webhooks.
+
 Managing Api Keys
 -----------------
 
--- a/app/controllers/shopify_app/sessions_controller.rb
+++ b/app/controllers/shopify_app/sessions_controller.rb
@@ -26,8 +26,7 @@ module ShopifyApp
     end
 
     def destroy
-      session[:shopify] = nil
-      session[:shopify_domain] = nil
+      reset_session
       flash[:notice] = I18n.t('.logged_out')
       redirect_to login_url
     end
@@ -46,6 +45,7 @@ module ShopifyApp
       sess = ShopifyAPI::Session.new(shop_name, token)
       session[:shopify] = ShopifyApp::SessionRepository.store(sess)
       session[:shopify_domain] = shop_name
+      session[:shopify_user] = associated_user if associated_user.present?
     end
 
     def auth_hash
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -207,6 +207,22 @@ provider :shopify,
...
+
+```ruby
+provider :shopify,
+  ShopifyApp.configuration.api_key,
+  ShopifyApp.configuration.secret,
+  scope: ShopifyApp.configuration.scope,
+  per_user_permissions: true
+```
+
+The current Shopify user will be store in the rails session at `session[:shopify_user]`
```
EiNSTeiN-: typo with `stored` ?

app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -36,6 +35,7 @@ def destroy
 
     def authenticate
       if sanitized_shop_name.present?
+        session['shopify.omniauth_params'] = { shop: sanitized_shop_name }
         fullpage_redirect_to "#{main_app.root_path}auth/shopify?shop=#{sanitized_shop_name}"
```
EiNSTeiN-: You can remove `shop=#{sanitized_shop_name}` from the url and only keep it in the session; so that there's no possible mistake about which one is the right one to use

kevinhughes27:
FYI I'll ship these commits separately since they are unrelated. I need them in one branch for now for our app.

app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -46,6 +46,7 @@ def login_shop
       sess = ShopifyAPI::Session.new(shop_name, token)
       session[:shopify] = ShopifyApp::SessionRepository.store(sess)
       session[:shopify_domain] = shop_name
+      session[:shopify_user] = associated_user if associated_user.present?
```
EiNSTeiN-: While I'm looking at this... I believe this code path should be protected against [session fixation](https://www.owasp.org/index.php/Session_fixation), this can be done with this:
```
    session.options[:renew] = true
    session.delete(:_csrf_token)
```

EiNSTeiN-:
The code in this branch looks good 👍

app/controllers/shopify_app/sessions_controller.rb
kevinhughes27: That causing `NoMethodError: undefined method `session_options' for nil:NilClass` Can we open an issue to tackle this later?

app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -36,16 +35,22 @@ def destroy
 
     def authenticate
       if sanitized_shop_name.present?
-        fullpage_redirect_to "#{main_app.root_path}auth/shopify?shop=#{sanitized_shop_name}"
+        session['shopify.omniauth_params'] = { shop: sanitized_shop_name }
```
Hammadk: I don't think we can set the session from the child-frame until the parent frame redirects to our app. If I remember correctly, this was an issue in Safari. Safari is more strict about third party cookies by default.

kevinhughes27: Sounds like we need to 🎩  with safari.

Hammadk: This was Safari 11 I believe (and some version of IE)

kevinhughes27: I was able to install my app using this branch using Safari on my laptop. I check the version and it is `Version 10.1.2 (12603.3.8)` - is my safari out of date?

kevinhughes27: Apple's download link for Safari 11 is broken. @Hammadk suggested I try changing this setting:

image

to `Allow from current website only`

I was able to install the app fine. I also cleared all the cookies in Safari and tried again which was also successful. I think this change is okay.

PR #459, opened 2017-08-28
basic setup for per-user authentication

kevinhughes27:
code extraction from #457

cc @Hammadk @EiNSTeiN-

### DIFF
--- a/app/controllers/shopify_app/sessions_controller.rb
+++ b/app/controllers/shopify_app/sessions_controller.rb
@@ -26,8 +26,7 @@ module ShopifyApp
     end
 
     def destroy
-      session[:shopify] = nil
-      session[:shopify_domain] = nil
+      reset_session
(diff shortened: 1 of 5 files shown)


app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -46,6 +45,7 @@ def login_shop
       sess = ShopifyAPI::Session.new(shop_name, token)
       session[:shopify] = ShopifyApp::SessionRepository.store(sess)
       session[:shopify_domain] = shop_name
+      session[:shopify_user] = associated_user if associated_user.present?
```
sshaw: I think this should be passed to `ShopifyAPI::Session.new` as [its 3rd argument](http://www.rubydoc.info/gems/shopify_api/ShopifyAPI%2FSession:initialize). This way, one can persist email, locale, etc... when storing a session.

PR #460, opened 2017-08-28
Pass shop in the session

kevinhughes27:
extracted from #457

@Hammadk @EiNSTeiN-

### DIFF
--- a/app/controllers/shopify_app/sessions_controller.rb
+++ b/app/controllers/shopify_app/sessions_controller.rb
@@ -35,7 +35,8 @@ module ShopifyApp
 
     def authenticate
       if sanitized_shop_name.present?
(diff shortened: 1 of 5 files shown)


app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -35,14 +35,19 @@ def destroy
 
     def authenticate
       if sanitized_shop_name.present?
-        fullpage_redirect_to "#{main_app.root_path}auth/shopify?shop=#{sanitized_shop_name}"
+        session['shopify.omniauth_params'] = { shop: sanitized_shop_name }
```
kevinhughes27: from the other PR there is some concern that this won't work in Safari. We need to test and confirm.

PR #483, opened 2017-10-03
change how the X-Frame-Options header is removed

### LINKED ISSUE #468: First request hangs
Using the new 8.1 gem, the idea is to use a session to store some OmniAuth info, rather than using a redirect with a GET parameter. Ok... 

I still have the same issue as before now, and cannot quite grasp the solution. If I start a new session with the App, it hangs on first hit. If I refresh, the second hit works instantly. If I stop Rails server in my console and restart it, and refresh the App, it hangs. It works perfect second hit. 

So it seems that something is up with the authentication code? I am not sure this is related to the Asset pipeline hanging. Why is it that any new Rails server on localhost will hang the App. The console shows the delivery of the App with a 200 OK status... but nothing is rendered. 

So this remains a mystery to me, on how to solve. Is annoying to have to manually refresh the browser every time a session expires and not have the App just refresh itself to get a new session going.

kevinhughes27:
fixes #468

@resistorsoftware / @meulmees lets try this fix suggested [here](https://stackoverflow.com/questions/18445782/how-to-override-x-frame-options-for-a-controller-or-action-in-rails-4)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+8.2.2
+-----
+* Changes how the ESDK concern allows iframes. Fixes an issue with the first request for some people
+
 8.2.1
 -----
 * Bugfix: Don't logout shops from `login_again_if_different_shop` when Rails
--- a/lib/shopify_app/controller_concerns/embedded_app.rb
+++ b/lib/shopify_app/controller_concerns/embedded_app.rb
@@ -13,7 +13,7 @@ module ShopifyApp
 
     def set_esdk_headers
       response.set_header('P3P', 'CP="Not used"')
-      response.default_headers.delete('X-Frame-Options')
+      response.headers.except!('X-Frame-Options')
     end
   end
 end


kevinhughes27:
cc @yaworsk

kevinhughes27:
I tested in our recent app (which actually needs this since I don't want the other controllers to be iframeable). The first request took a bit longer and then finished (I suspect asset compilation). I should point out that I get the same result with the old code as well.

yaworsk:
Recognizing I'm not familiar with this code, the secure_headers gem does provide per controller overrides https://github.com/twitter/secureheaders/blob/master/docs/named_overrides_and_appends.md

kevinhughes27:
ideally we can override this using vanilla rails methods. It would suck to have to detect the secure headers gem and do something extra.

resistorsoftware:
I just tried this out. Using this change the browser still whined... 

Refused to display 'https://taghomedecor.dev/login?shop=swift-braun3430.myshopify.com' in a frame because it set 'X-Frame-Options' to 'sameorigin'.

My shopify App embedded_app code:

    def set_esdk_headers
      response.headers.except!('X-Frame-Options')
    end

The Rails render seemed to fine, but nothing actually rendered... 

    Started GET "/login?shop=swift-braun3430.myshopify.com" for 127.0.0.1 at 2017-10-05 13:34:48 -0400
    Processing by ShopifyApp::SessionsController#new as HTML
      Parameters: {"shop"=>"swift-braun3430.myshopify.com"}
      Rendering inline template
      Rendered inline template (1.1ms)
      Completed 200 OK in 8ms (Views: 5.1ms | Models: 0.0ms)

kevinhughes27:
Why is your login page begin iframed? The login page needs to be outside of the iframe.

resistorsoftware:
It is called from Shopify... clicking on the App icon in the merchant
shop... what makes you think the login page is in an iframe? The fact the
browser spit that out in the console?


514.795.9601
Book an Appointment!
https://calendly.com/dlazar

kevinhughes27:
Yeah I am confused why the browser would print that message

resistorsoftware:
It is a stock Shopify App gem 8.2 running a Rails 5.1 stock App with almost nothing special going on... a very basic install... I did not suddenly re-program the internal routes to render login in an iframe or anything.. and the App works fine with the headers in application.rb, I simply tinkered with the patch code... and output some logs to ensure all that code is working.

kevinhughes27:
Can you try adding the headers to the login page as well and remove them from application.rb. That could be the issue.

meulmees:
I just tried this out, and it is consistently fixing the issue for me!

Like you mentioned, after restart the first request is slower likely because of asset compilation, but it definitely loads. Where as without this fix the first request never loads.

kevinhughes27:
Okay I am going to merge this then since it fixes it for one of the two cases. @meulmees I am assuming you tested in our codebase? If you have a second can you try a vanilla app as well? This might help you out https://github.com/kevinhughes27/dotfiles/blob/master/shell/macros.sh#L80 :)

kevinhughes27:
@yaworsk we'll handle how this will work with secure headers later

resistorsoftware:
Would be nice to see an example that works. When I use the patch my browser stops on an frame sameorigin header, so this patch hardly seems ideal when I see it failing with a standard Rails 5.1 App using the gem. Back to the drawing board it seems. I will simply live with the header management in application.rb till I can get the alternative universe version to work. Thanks for all the effort and attention paid to the issue!! Every little bit helps.

flavio-b:
@kevinhughes27 , I was also getting the same error `Refused to display... in a frame because it set 'X-Frame-Options' to 'sameorigin'.` 

It happens when you clear the session (or when it expires) and you try to the open the app, from the App's section within the Shopify Admin. The controller tries to GET /login inside the iframe, and we get that error. At least that's what it looks like in the tests I did.

I "fixed" it by generating the shopify app controllers in my app, and adding an after filter to the `SessionsController`, like so:

`
after_action only: [:new, :create] do |controller|
      controller.response.headers.except!('X-Frame-Options')
end
`

It's working now all the time.

kevinhughes27:
Thanks! sounds like we should add that to the SessionsController. Unfortunately I don't think we can include the EmbeddedApp concern as is since there are other endpoints we don't want this removed from. Is that something you'd be interested in submitting a PR for?

flavio-b:
Ok, I think this should do: #504 . It's limited to the GET /login route.

PR #485, opened 2017-10-07
Update Gemfile

limsammy:
Version of rails is outdated causing bundler to fail. Latest version causes all to work well again!

### DIFF
--- a/example/Gemfile
+++ b/example/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'rails', '4.2.5.1'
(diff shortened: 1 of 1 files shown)


example/Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'rails', '4.2.5.1'
+gem 'rails' # v4.2.5.1 is outdated, causing bundler to fail. all works well with latest version of rails
```
kevinhughes27: can you update this all the way to the latest rails please. Thanks!

tylerball:
Confirmed that `bundle install` works fine in the example app on `master`, closing.

PR #487, opened 2017-10-17
Adjustment to README on Custom Webhooks section

ohmybrew:
Updated README example for custom webhook controller to more closely match how `webhooks_controller.rb` works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ class CustomWebhooksController < ApplicationController
   include ShopifyApp::WebhookVerification
 
   def carts_update
-    params.try(:permit!)
+    params.permit!
     SomeJob.perform_later(shop_domain: shop_domain, webhook: webhook_params.to_h)
     head :no_content
   end
--- a/README.md
+++ b/README.md
@@ -275,20 +275,27 @@ ShopifyApp.configure do |config|
 end
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -275,20 +275,27 @@ ShopifyApp.configure do |config|
...
+If you'd rather implement your own controller then you'll want to use the WebhookVerfication module to verify your webhooks, example:
 
 ```ruby
 class CustomWebhooksController < ApplicationController
   include ShopifyApp::WebhookVerification
 
   def carts_update
-    SomeJob.perform_later(shopify_domain: shop_domain, webhook: params)
-    head :ok
+    params.try(:permit!)
```
kevinhughes27: can you explain what this is for?

ohmybrew: Ah, it was just to match closely to how the `WebhookController` is currently doing it in the lib. However, we're not filtering anything so permit wouldn't make sense, I could remove it if that would be best?

kevinhughes27: Oh nvm I see now - it permits everything. I think the `try` threw me off. I bet that was to try and support old versions of Rails which we don't do anymore. Can you update this in both places to be `params.permit!`

ohmybrew: Sure, this is done. That makes sense, thanks.

PR #495, opened 2017-10-24
Make compatible with Turbolinks >= 5.0.0

flavio-b:
Update event name to trigger the handler on newer Turbolinks versions. Explicitly remove the event listener after the flash notice (or error) is triggered once.


kevinhughes27:
Thanks!

PR #497, opened 2017-10-25
add /callback for oauth callback

karlcoelho:
A lot of tutorials use the /callback endpoint for oAuth requests, so I thought it would be a useful addition.


kevinhughes27:
Can you supply a few examples of tutorials doing this? The url is currently done according to the omniauth pattern I've seen in a lot of places. I think `/callback` is usually used on really minimal examples since the route isn't very descriptive for a larger project. I also don't like the idea of two routes that do the same thing pointing to the same controller.

jamiemtdwyer:
I could be wrong, but I don't think there are any tutorials written specifically for `shopify_app` that include this route. As Kevin said, I think it's best not to add any duplicate routes here.

We appreciate the contribution as always, but I'll be closing this for now.

PR #504, opened 2017-11-15
Remove X-Frame-Options header on login route

flavio-b:
This prevents a browser error where the login route is called from the embedded app iframe.


kevinhughes27:
Thanks!

PR #519, opened 2017-12-05
Make sessions controller embeddable

### LINKED ISSUE #518: Refused to display login page
When I open an embedded app from Shopify admin with cleaned cookies, for example, I see this error in js console:
```
Refused to display 'https://APP_URL/login?shop=oh-my-flowers.myshopify.com' in a frame because it set 'X-Frame-Options' to 'sameorigin'.
```

Looks like we need to reset 'X-Frame-Options' for the login page too.

I using Rails 5.2.0 beta 2 (maybe this also causing this issue)

kirillplatonov:
Fixes #518

### DIFF
--- a/app/controllers/shopify_app/sessions_controller.rb
+++ b/app/controllers/shopify_app/sessions_controller.rb
@@ -1,6 +1,8 @@
 module ShopifyApp
   class SessionsController < ActionController::Base
     include ShopifyApp::LoginProtection
+    include ShopifyApp::EmbeddedApp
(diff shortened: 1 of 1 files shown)


kevinhughes27:
Its not in a release yet but we already shipped a fix for this: https://github.com/Shopify/shopify_app/commit/2f8051fea45b38e6c83b84ca6bf3826e4c31e836. I'll cut a release today. Please re-open if we missed something.

PR #521, opened 2017-12-06
don't require 'rails/generators/active_record' in install generator

mikeyhew:
The `install` generator doesn't do anything involving ActiveRecord, so this `require` is unnecessary here


jamiemtdwyer:
Makes sense to me – @kevinhughes27 for sanity

PR #534, opened 2017-12-27
Use form_tag helper to create login form

flavio-b:
On #513, the login form method was changed from GET to POST. This causes Rails to raise an invalid token error. Using the form_tag helper, the form automatically gets a hidden input with the token, to protect from CSRF attacks.

### DIFF
--- a/app/views/shopify_app/sessions/new.html.erb
+++ b/app/views/shopify_app/sessions/new.html.erb
@@ -110,13 +110,13 @@
       <label for="shop">Enter your shop domain to log in or install this app.</label>
     </p>
 
-    <form method="POST" action="login">
+    <%= form_tag login_path do %>
       <% if flash[:error] %>
         <div class=error><%= flash[:error] %></div>
       <% end %>
       <input id="shop" name="shop" type="text" autofocus="autofocus" placeholder="example.myshopify.com" class="marketing-input">
       <button type="submit" class="marketing-button">Install</button>
-    </form>
+    <% end %>
   </main>
 
 </body>


damhonglinh:
I'm configuring a new Shopify app and I kept getting `ActionController::InvalidAuthenticityToken (ActionController::InvalidAuthenticityToken)`. It took me a while to know it's an issue from the gem but not from my new app. It would be great for later developers if this PR is merged soon.

kevinhughes27:
ping @ShayneP @jordanliddle @meulmees

jamiemtdwyer:
Sorry, catching up on emails from the holidays. Thanks for reviewing @kevinhughes27 !

jamiemtdwyer:
Sounds like this is probably worth releasing a new version for?

benbonnet:
using rails  5.2 (idk if this is related to the version,I'm just unable to `POST` anything (meaning, opening cors, with `skip_before_action :verify_authenticity_token`, etc…) Is that related to this issue ?

meulmees:
I've got an app running on rails `5.1.3` without any issues.

@benbonnet can you provide more context around what you're trying to do, and a stack trace for the error(s) you're seeing?

benbonnet:
super standard; so for the stacktrace (you only get a single error line for this one). To correct my previous comment, it only happens on 5.2; tried again on 5.1. Solve it for now by prepending my picking the value of the csrf meta available in the layout header

```
headers: {
        'Content-Type': 'application/json',
        'X-CSRF-Token': document.querySelector('meta[name="csrf-token"]').getAttribute('content')
      }
```

PR #537, opened 2018-01-08
Fix #535

### LINKED ISSUE #535: Fullpage redirect with JavaScript doesn't work in Rails 5.2 due to `Content Security Policy`
Reproduce steps:
1. In Chrome, go to https://example.com/login?shop=abc.myshopify.com
2. You will see an error in DevTools console: `Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' https:". Either the 'unsafe-inline' keyword, a hash ('sha256-sometokenhere='), or a nonce ('nonce-...') is required to enable inline execution`

This happens in Rails 5.2 (specifically for example I use `5.2.0.beta2`) and is due to a new security feature in Rails 5.2: https://github.com/rails/rails/pull/31162

EiNSTeiN-:
Move inline javascript into a .js asset so it plays nicely with CSP.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,5 @@ source "https://rubygems.org"
 
 # Specify your gem's dependencies in shopify_app.gemspec
 gemspec
+
+gem 'rails-controller-testing', group: :test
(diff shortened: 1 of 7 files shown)


jamiemtdwyer:
This doesn't seem to solve the problem in the referenced issue.

When testing this branch with Ruby 2.3.4 and Rails 5.2.0beta, attempting to install from the `/login` route, after inputting my shop name the title bar, the browser hangs and I get the following in console:

```
Uncaught TypeError: Cannot read property 'dataset' of null at HTMLDocument.<anonymous>
```

Seems that `redirectTargetElement` is null here: https://github.com/Shopify/shopify_app/blob/52c25213e70378b5ee9482571f954943ff11f70e/app/assets/javascripts/shopify_app/redirect.js#L3

jamiemtdwyer:
🚢

PR #538, opened 2018-01-09
Prevent flash script from running on Turbolinks preview

flavio-b:
Adds a data attribute to prevent the flash code from running when Turbolinks is displaying a preview of the page. Otherwise, the flash code will run again during that short instant of the preview.

The same can be achieve by wrapping the flash code in a conditional to check if:
`document.documentElement.hasAttribute("data-turbolinks-preview")`

Just annotating with an attribute seems cleaner, though not so clear.

### DIFF
--- a/lib/generators/shopify_app/install/templates/_flash_messages.html.erb
+++ b/lib/generators/shopify_app/install/templates/_flash_messages.html.erb
@@ -1,5 +1,5 @@
 <% content_for :javascript do %>
-<script type="text/javascript">
+<script type="text/javascript" data-turbolinks-eval="false">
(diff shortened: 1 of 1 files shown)


flavio-b:
Sorry! After doing some more tests, it seems like just adding the Javascript conditional to prevent the code from running during Turbolinks preview works better.

jamiemtdwyer:
Hey @flavio-b, do you want to squash these commits? `git rebase -i ~HEAD~3` and then force push to your branch.

flavio-b:
@jamiemtdwyer ,  thanks. I actually just sent a new PR with a single commit and closed this one.

PR #540, opened 2018-01-15
Prevent flash messages on Turbolinks preview

flavio-b:


jamiemtdwyer:
See https://github.com/Shopify/shopify_app/pull/538 for context

Issue #541, opened 2018-01-19
Documentation not accurate about AuthenticatedController


nickmealey:
Readme currently [says about](https://github.com/Shopify/shopify_app#authenticatedcontroller) the AuthenticatedController that it inherits from `ApplicationController`, when it actually inherits from `ActionController::Base` which was changed in [this commit](https://github.com/Shopify/shopify_app/blob/master/app/controllers/shopify_app/authenticated_controller.rb). What was the reason for this change? Controllers that inherit from `AuthenticatedController` now can't use `ApplicationController methods`

kevinhughes27:
It was done to avoid a situation where a developer or another gem adds methods to application controller which interferes with ShopifyApp. If you need custom behavior then you can still make your own controller from application controller and include the concern.

The readme should be updated.

vfonic:
I fixed this here: https://github.com/Shopify/shopify_app/pull/550

This issue can be closed now.

0ptimusrhyme:
Thanks!

PR #548, opened 2018-02-24
Prefer AuthenticatedByShopify concern over AuthenticatedController

kdonovan:
I ran into the bug pointed out in #541, where the docs incorrectly claim `ShopifyApp::AuthenticatedController` inherits from `ApplicationController`. 

I understand the logic behind changing this to avoid accidentally importing breaking logic changes from the user's application, but it still causes issues as e.g. my local app has customizations to the ApplicationController to add support for Pundit and many other things. Any "userland" additions to Shopify integration will need _both_ my expected `ApplicationController` logic _and_ the default Shopify authentication logic; I'd originally worked around that by extracting all my `ApplicationController` code into a new `ApplicationControllerConcern` concern of my own, and injecting that into both the standard `ApplicationController` and a separate `Shopify::ApplicationController` that inherits from `ShopifyApp::AuthenticatedController` (and using that as the base for all my additional Shopify controllers). It works... but it's hacky and unpleasant, and I was inspired once I saw from #541 that this behavior wasn't exactly completely polished/decided upon.

After this PR, all my local workarounds become unnecessary. Userland Shopify controllers inherit from `ApplicationController`, like any other controller, and merely include the new concern:

```ruby
class SomeNewShopifyController < ApplicationController
  include ShopifyApp::AuthenticatedByShopify
end
```

The test suite is still green, and I can't see any obvious issues... but I'd appreciate another set of eyes to make sure I'm not missing anything obvious. 👀


vfonic:
This also makes it consistent with [`AppProxyVerification`](https://github.com/Shopify/shopify_app/blob/master/lib/shopify_app/controller_concerns/app_proxy_verification.rb) and [`WebhookVerification`](https://github.com/Shopify/shopify_app/blob/master/lib/shopify_app/controller_concerns/webhook_verification.rb).

I would just change the name slightly (following the two above-mentioned concerns). Perhaps `ShopifyApp::AuthenticationVerification`?

lib/shopify_app/controller_concerns/authenticated_by_shopify.rb
```diff
@@ -0,0 +1,16 @@
...
+      include ShopifyApp::LoginProtection
+      include ShopifyApp::EmbeddedApp
+
+      protect_from_forgery with: :exception
+      before_action :login_again_if_different_shop
+      around_action :shopify_session
+    end
+
+  end
+end
```
kevinhughes27: can you add the newline char here please

kdonovan:
Apologies on the delay here -- day job went a little haywire -- but I'll definitely circle back to this project shortly and I'll update the PR as discussed above then. 👍

kdonovan:
@EiNSTeiN- Apologies again for the delay, but I was able to circle back to this today and update the generators + add a deprecation warning.

Note: I started looking at the failing code-review/policial checks, but implementing those (e.g. frozen string literal comment) in these files actually diverges the style from the existing codebase.

PR #550, opened 2018-03-02
Fix: AuthenticatedController inheritance docs

vfonic:
AuthenticatedController inherits from ActionController::Base


EiNSTeiN-:
Thanks

PR #560, opened 2018-03-13
Clarified steps for default generator

MisinformedDNA:


0ptimusrhyme:
Is this super necessary? I feel like it could be interpreted that `rake db:migrate` and `bundle exec rails server` are required as part of running the generator. While it's true that you'll have to run `db:migrate` eventually (and `bundle exec rails server` to run the app), the purpose of this section is intended to describe the generator itself, not what comes next.

jamiemtdwyer:
Maybe we could add another section for "Running the application" instead?

0ptimusrhyme:
I think that'd be a good idea. Would make sense to remove `After running the install generator, you can start your app with bundle exec rails server and install your app by visiting localhost.` from the Install Generator section as well, as my thoughts apply to that in the same way.

vfonic:
There's also this at the bottom of [Shop Generator](https://github.com/Shopify/shopify_app#shop-model-generator):

_Note that you will need to run rake db:migrate after this generator_

That could be added at the bottom of install generator. The completely new section idea sounds good too!

MisinformedDNA:
Right, Shop Generator says what to do next, but Default Generator just leaves you hanging.

PR #564, opened 2018-03-24
Quickstart.md: Update snippet MD to shell syntax, add notes on credentials to step 5

0ptimusrhyme:
Finishes #224 and resolves MD snippet nitpicks on Quickstart.md

CC @jamiemtdwyer for 👀

### DIFF
--- a/docs/Quickstart.md
+++ b/docs/Quickstart.md
@@ -6,12 +6,12 @@ Build and deploy a new Shopify App to Heroku in minutes
 1. New Rails App (with postgres)
 --------------------------------
 
-```
-rails new test-app --database=postgresql
-cd test-app
-git init
-git add .
-git commit -m 'new rails app'
+```sh
+$ rails new test-app --database=postgresql
+$ cd test-app
+$ git init
+$ git add .
+$ git commit -m 'new rails app'
 ```
 
 2. Create a new Heroku app
@@ -20,17 +20,17 @@ git commit -m 'new rails app'
 The next step is to create a new heroku app. Pull up your heroku dashboard and make a new app!
 
 cli:
+```sh
+$ heroku create name
+$ heroku git:remote -a name
 ```
-heroku create name
-heroku git:remote -a name
-  ```
 
 now we need to let git know where the remote server is so we'll be able to deploy later
 
 web:
-```
-https://dashboard.heroku.com/new
-git remote add heroku git@heroku.com:appinfive.git
+```sh
+# https://dashboard.heroku.com/new
+$ git remote add heroku git@heroku.com:appinfive.git
 ```
 
 3. Create a new App in the partners area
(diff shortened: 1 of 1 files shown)


docs/Quickstart.md
```diff
@@ -43,28 +43,31 @@ git remote add heroku git@heroku.com:appinfive.git
...
+$ git add .
+$ git commit -m 'generated shopify app'
 ```
-use the keys from your app in the partners area
-rails generate shopify_app --api_key a366cbafaccebd2f615aebdfc932fa1c --secret 8750306a895b3dbc7f4136c2ae2ea293
-git add .
-git commit -m 'generated shopify app'
-```
+
+If you forget to set your keys or redirect uri above you will find them in the shopify_app initializer at: `/config/initializers/.`
```
vfonic: I'd write the full initializer name, as it's hardcoded and won't be changing anytime soon (or ever): `config/initializers/shopify_app.rb` (also put the dot '.' out of backticks)

docs/Quickstart.md
```diff
@@ -43,28 +43,31 @@ git remote add heroku git@heroku.com:appinfive.git
...
+
+We recommend adding a gem or utilizing ENV variables to handle your keys before releasing your app.
```
vfonic: I agree with this, and maybe it would be better to have the example above NOT show adding keys, which makes them added to `shopify_app.rb` initializer.

Perhaps change the above to:

`$ rails generate shopify_app --api_key <shopify_api_key> --secret <shopify_api_secret>`

I believe these are the default values in the generator itself if nothing is passed as command line argument.

0ptimusrhyme: The defaults in the initializer are `<api_key>` and `<secret>` respectively. Do you think it'd be better to mirror those completely, or to use your values to be more descriptive? I'm leaning toward your recommendation, but interested to hear your thoughts. :D

docs/Quickstart.md
0ptimusrhyme: Good catch! Will be in next commit pending any further feedback.

docs/Quickstart.md
vfonic: I'm indifferent. I just know that a lot of people just copy-paste everything so it might be good to use obviously dummy values and point them in the right direction (which might even be pointing them towards gems like [figaro](https://github.com/laserlemon/figaro) or [dotenv](https://github.com/bkeepers/dotenv)

jamiemtdwyer: I think this is a good suggestion 👍

jamiemtdwyer:
LGTM pending @vfonic's suggestions being addressed

docs/Quickstart.md
0ptimusrhyme: Done and done.

PR #571, opened 2018-04-04
Sanitizes the shop query param in case it's missing the .myshopify.com

meulmees:
We're currently seeing an issue where if we leave off the `.myshopify.com` bit of the shop query parameter we end up with the following error after a few redirects:

screen capture on 2018-04-04 at 14-04-44

Example: `https://www.myapp.com/shopify_app/login?shop=shop1`

This PR uses adds a new `sanitized_params` that uses the `sanitize_shop_param` to make sure that the `sanitized_params[:shop]` is correct and contains the appropriate shopify domain.

cc/ @TejasM

### DIFF
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -40,7 +40,9 @@ module ShopifyApp
       if request.xhr?
         head :unauthorized
       else
-        session[:return_to] = request.fullpath if request.get?
+        if request.get?
+          session[:return_to] = "#{request.path}?#{sanitized_params.to_query}"
+        end
         redirect_to login_url
       end
     end
@@ -55,7 +57,7 @@ module ShopifyApp
       url = ShopifyApp.configuration.login_url
 
       if params[:shop].present?
-        query = { shop: params[:shop] }.to_query
+        query = { shop: sanitized_params[:shop] }.to_query
         url = "#{url}?#{query}"
       end
 
@@ -84,6 +86,14 @@ module ShopifyApp
     def sanitize_shop_param(params)
       return unless params[:shop].present?
       ShopifyApp::Utils.sanitize_shop_domain(params[:shop])
+    end
+
+    def sanitized_params
+      request.query_parameters.clone.tap do |query_params|
+        if params[:shop].is_a?(String)
+          query_params[:shop] = sanitize_shop_param(params)
+        end
+      end
     end
   end
 end
--- a/test/shopify_app/controller_concerns/login_protection_test.rb
+++ b/test/shopify_app/controller_concerns/login_protection_test.rb
@@ -66,8 +66,8 @@ class LoginProtectionTest < ActionController::TestCase
       session[:shopify_domain] = "foobar"
       sess = stub(url: 'https://foobar.myshopify.com')
       ShopifyApp::SessionRepository.expects(:retrieve).returns(sess).once
-      get :second_login, params: { shop: 'other_shop' }
-      assert_redirected_to '/login?shop=other_shop'
+      get :second_login, params: { shop: 'other-shop' }
+      assert_redirected_to '/login?shop=other-shop.myshopify.com'
       assert_nil session[:shopify]
       assert_nil session[:shopify_domain]
     end
@@ -88,14 +88,32 @@ class LoginProtectionTest < ActionController::TestCase
   test '#shopify_session with no Shopify session, redirects to the login url' do
     with_application_test_routes do
       get :index, params: { shop: 'foobar' }
-      assert_redirected_to '/login?shop=foobar'
+      assert_redirected_to '/login?shop=foobar.myshopify.com'
+    end
+  end
+
+  test '#shopify_session with no Shopify session, redirects to the login url \
+        with non-String shop param' do
+    with_application_test_routes do
+      params = { shop: { id: 123 } }
+      get :index, params: params
+      assert_redirected_to "/login?#{params.to_query}"
     end
   end
 
   test '#shopify_session with no Shopify session, sets session[:return_to]' do
     with_application_test_routes do
       get :index, params: { shop: 'foobar' }
-      assert_equal '/?shop=foobar', session[:return_to]
+      assert_equal '/?shop=foobar.myshopify.com', session[:return_to]
+    end
+  end
+
+  test '#shopify_session with no Shopify session, sets session[:return_to]\
+        with non-String shop param' do
+    with_application_test_routes do
+      params = { shop: { id: 123 } }
+      get :index, params: params
+      assert_equal "/?#{params.to_query}", session[:return_to]
     end
   end
 
@@ -109,7 +127,7 @@ class LoginProtectionTest < ActionController::TestCase
   test '#shopify_session when rescuing from unauthorized access, redirects to the login url' do
     with_application_test_routes do
       get :raise_unauthorized, params: { shop: 'foobar' }
-      assert_redirected_to '/login?shop=foobar'
+      assert_redirected_to '/login?shop=foobar.myshopify.com'
     end
   end
 


jamiemtdwyer:
> https://www.myapp.com/shopify_app/login?shop=shop1

should this ever actually happen under normal circumstances?

meulmees:
it happens because experts small tasks (ex: https://experts.shopify.com/tasks/1) asks the merchant to enter their shop name
screen shot 2018-04-04 at 2 30 17 pm

jamiemtdwyer:
cool, makes sense. 👍

EiNSTeiN-:
shopify will always set `shop` as qualified domain, so the embedded app can expect shop to always contain `myshopify.com`

Your app should always check that `.myshopify.com` is part of the url _before_ redirecting to it, otherwise you may redirect to a domain that is not controlled by shopify.

jamiemtdwyer:
Sorry, good catch

meulmees:
Isn't that what `ShopifyApp::Utils.sanitize_shop_domain` is doing? And if you tried to do that (`https://www.myapp.com/shopify_app/login?shop=dereks-tester.foo.bar`) then you get:
screen shot 2018-04-04 at 3 00 23 pm


I don't see why my app shouldn't be able to just add
```ruby
before_action :login_again_if_different_shop
```

To my base controller and then you can hit `https://www.myapp.com/shopify_app/login?shop=dereks-tester` and it works the same as `dereks-tester.myshopify.com`

meulmees:
You can test this right now in prod with this URL: https://partners.shopify.com/services_marketplace

Works: https://partners.shopify.com/services_marketplace?shop=dereks-tester.myshopify.com
No Works: https://partners.shopify.com/services_marketplace?shop=dereks-tester

(Of course you can use your own shop)

test/shopify_app/controller_concerns/login_protection_test.rb
```diff
@@ -66,8 +66,8 @@ class LoginProtectionTest < ActionController::TestCase
       session[:shopify_domain] = "foobar"
       sess = stub(url: 'https://foobar.myshopify.com')
       ShopifyApp::SessionRepository.expects(:retrieve).returns(sess).once
-      get :second_login, params: { shop: 'other_shop' }
```
meulmees: Separately I think there's a bug with `ShopifyApp::Utils.sanitize_shop_domain`
It doesn't like `_` characters - is that expected?

EiNSTeiN-: `_` is not a legal character in domain names, so yes this is expected

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -85,5 +85,11 @@ def sanitize_shop_param(params)
       return unless params[:shop].present?
       ShopifyApp::Utils.sanitize_shop_domain(params[:shop])
     end
+
+    def sanitized_params
+      query_params = request.query_parameters
+      query_params[:shop] = sanitize_shop_param(params) if params[:shop].present?
```
Hammadk: I think this should only happen if `params[:shop]` is a string similar to our [check here](https://github.com/Shopify/shopify_app/blob/3fb589d71bc03a11a8bb48bf87e613f6ce0210ea/lib/shopify_app/controller_concerns/login_protection.rb#L30).

When `params[:shop]` is not a string, that means that it could be a hash. This will happen when someone has a `Shop` model in their Rails app, and there are some parameters being passed for it.

nickhoffman:
I thought the exact same thing as @Hammadk: the API client should be responsible for providing a valid myshopify domain. However, if parts of this gem already correct the domain by appending `.myshopify.com` , we should be consistent by doing one of two things:

1. Update this part of the gem to append `.myshopify.com` when that's missing; or
2. Update the other parts of the gem to not append `.myshopify.com` when it's missing.

meulmees:
@nickhoffman - given that we allow you to just enter `dereks-tester` (without `.myshopify.com`) into the login page:
screen shot 2018-04-05 at 2 11 30 pm

I feel like it follows that you should also be able to pass the `/login?shop=dereks-tester` and it work as well.

Do you have strong opinions?

jamiemtdwyer:
As one of the de-facto owners of this repo currently, I don't have any strong opinions. I'll echo @EiNSTeiN- concerns that he stated earlier

> Your app should always check that .myshopify.com is part of the url before redirecting to it, otherwise you may redirect to a domain that is not controlled by shopify.

As long as we're confident that this will continue to be the case, I'm in favour of this change.

meulmees:
Yep definitely, and I have tested to ensure that's still taking place

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -85,5 +85,13 @@ def sanitize_shop_param(params)
       return unless params[:shop].present?
       ShopifyApp::Utils.sanitize_shop_domain(params[:shop])
     end
+
+    def sanitized_params
+      query_params = request.query_parameters
+      return query_params unless params[:shop].is_a?(String)
+
+      query_params[:shop] = sanitize_shop_param(params)
```
nickhoffman: This will modify the object that `request.query_parameters` returns, which will affect other consumers of `request.query_parameters` . I suggest we clone the value before modifying it.

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -85,5 +85,13 @@ def sanitize_shop_param(params)
...
+      query_params
```
nickhoffman: The code in this method is fairly procedural. I think we can make it more OO and concise. How about something like this?

```ruby
request.query_parameters.clone.tap do |query_params|
  query_params[:shop] = sanitize_shop_params(params) if params[:shop].is_a?(String)
end
```

Hammadk: Can we also write a test for the `params[:shop].is_a?(String)` check. That'll serve as documentation for when someone tries to remove this.

lib/shopify_app/controller_concerns/login_protection.rb
meulmees: Excellent point

lib/shopify_app/controller_concerns/login_protection.rb
meulmees: Thanks very much @nickhoffman - `tap` is definitely a much nicer approach.

I've also added a couple tests for the `login_url` and `return_to` URL's containing a non-String shop param

nickhoffman:
Policial's failing due to several `Metrics/LineLength: Line is too long.` violations. Would you mind clearing this up before shipping, please?

meulmees:
Cleaned up policial, the only one remaining is around the class file being to large for the test:
`test/shopify_app/controller_concerns/login_protection_test.rb:29 | Metrics/ClassLength: Class has too many lines. [140/100]`

It would seem to me that we might want to disable that rule for test files but 🤷‍♂️

vfonic:
I was thinking about similar issue. I wanted to build my own solution and test what users are typing in the login input field. I understand there are many users out there and I'd expect some of them to type only the shop name without `.myshopify.com`.

More importantly, I'd expect them to also type full qualified url, with protocol (`https`) and maybe even following slash `/` and maybe even they just picked up the url on some random, say, product page. I'd want to provide the best UX to my users by allowing them to add their url any way they want, as long as I can figure out which shop we're talking about here. I don't want to force them to have to type in exactly correct format.

It would be great to have this feature in this gem. I understand it doesn't fit under this PR, but it just popped in my mind as I was reading this PR. If you want, I can open an issue and we can discuss it. Otherwise, I'll add on top of gem's functionality and provide this for my own customers.

EiNSTeiN-:
Viktor, regarding shop name, I don't think it's worth investing time into
this now, for reasons I won't go into on this public issue.

Derek, for class length error you can ignore for now, this repo's rubocop
config should be updated to inherit from our common style guide which
doesn't have this rule.

PR #579, opened 2018-04-19
clear shopify_user in session

ccleung:
https://github.com/Shopify/shopify_app/pull/459 introduced shopify_user in session, and it's cleared out here: https://github.com/Shopify/shopify_app/blob/master/app/controllers/shopify_app/sessions_controller.rb#L32

However, we should clear it out as well in login_protection

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,7 @@
+LineLength:
+  Exclude:
+    - test/**/*
+
+Metrics/ClassLength:
+  Exclude:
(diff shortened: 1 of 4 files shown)


alanhill:
As an aside.. `close_session`, `login_as_different_shop`, `redirect_to login_url`, `login_url` and `clear_shop_session` are a handful of methods repeating a bunch of logic and making the source code a bit trickier to consume. We should look at simplifying this.

Issue #552, opened 2018-03-05
Issue with `login_url` being redefined


vfonic:
I'm using a rails engine gem that is built on top of `shopify_app`. I've got some custom additional logic that I've put in my gem and I'm directly inheriting, instead of from `ShopifyApp::AuthenticatedController`, from `BulletTrain::AuthenticatedController` (my gem's awesome name), which inherits from `ShopifyApp::AuthenticatedController`.

The issue I'm currently having is because of this line:
```rb
controller :sessions do
  get 'login' => :new, :as => :login
end
```

What happens is that this line causes rails url helper to create a `login_url` method. When called through my gem, it looks like the app overrides `login_url` defined in `shopify_app` gem in [`LoginProtection`](https://github.com/Shopify/shopify_app/blob/master/lib/shopify_app/controller_concerns/login_protection.rb#L54)

This causes a bug, where a user, without session (in my app), gets a login screen (due to rails url helper `login_url` returning login url without the shop param, therefore preventing auto-login), instead of being logged in immediately.

Related issue: https://github.com/Shopify/shopify_app/issues/524

A simple solution would be to rename the `login_url` method in `ShopifyApp::LoginProtection` to a non-conflicting name. I don't know how to fix this error otherwise. Any ideas?

Should I create a PR? I suggest `login_url_for_shop` as it creates the url based on the `params[:shop]`.

EiNSTeiN-:
I totally understand that the name clash is unexpected. I believe you should be able to solve this problem without changes to this gem:

Assuming you mount the engine as our generator suggests:
```ruby
mount ShopifyApp::Engine, at: '/'
```
Add a namespace to this line:
```ruby
mount ShopifyApp::Engine, at: '/', as: :shopify_app
```
this should work as well:
```ruby
namespace :shopify_app do
  mount ShopifyApp::Engine, at: '/'
end
```
then you'll be able to avoid the name conflict between LoginProtection's login_url and the route helper:
```ruby
shopify_app_login_url
```

vfonic:
@EiNSTeiN- ah nice! I didn't know that all the routes used internally inside the gem will still work (for example `sessions/new.html.erb` uses `form_for login_path, ...`.
Is this because of `isolate_namespace` or it's just a standard gem behavior?

I guess I could close this issue.

EiNSTeiN-:
In the view context you would only have access to route helpers, so the LoginProtection methods would not be available unless you also add `helper_method :login_url` somewhere in your controller.
[...]

PR #585, opened 2018-05-02
Rename `login_url` method to `login_url_with_optional_shop` to avoid ambiguity with Rails' route helper method of the same name

swalkinshaw:
`login_url` potentially conflicts with Rails' route helper methods which are auto generated. Since this engine defines a `login` route, `login_url` would be the generated method.

While this doesn't cause problems in normal use cases, if the same route is defined in the main app's route to point to a customized sessions controller (for example), the route helper method `login_url` will take precedence causing redirects to happen *without* the `shop` param.

ie:

```ruby
  get 'login', to: 'sessions#new', as: :login

end
```

Even though this is guarding against an uncommon edge case, there's no reason to keep the conflicting method name.

This simply renames it to `login_url_with_optional_shop`.

NOTE: I guess this is a potentially breaking change if anyone had overridden `login_url` previously even though it's a protected method.

### DIFF
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -15,7 +15,7 @@ module ShopifyApp
         redirect_to return_address
       else
         flash[:error] = I18n.t('could_not_log_in')
-        redirect_to login_url_with_optional_shop
+        redirect_to(login_url_with_optional_shop)
       end
     end
 
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -15,7 +15,7 @@ module ShopifyApp
         redirect_to return_address
       else
         flash[:error] = I18n.t('could_not_log_in')
-        redirect_to login_url
+        redirect_to login_url_with_optional_shop
       end
     end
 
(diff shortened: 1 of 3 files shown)


vfonic:
I've also hit this issue: [Issue with `login_url` being redefined #552](https://github.com/Shopify/shopify_app/issues/552)

I'd love to see this change. Maybe in v9.

vfonic:
I've again spent several *days* tracking down this issue. Can you please merge it? That would be really appreciated!

chrisbutcher:
I'll rebase and ship this today.

chrisbutcher:
cc: @Hammadk @EiNSTeiN- Can I get another quick review of this PR?

chrisbutcher:
👋 @Shopify/platform-dev-tools-education Can I also get another review from one of you?

chrisbutcher:
Thanks @nwtn ! 

> lmk if you need me to 🎩
Actually, that would be super helpful :)

vfonic:
Thanks guys for getting back on this right away! :+1:

chrisbutcher:
@swalkinshaw would like to get another PR merged into `shopify_app` alongside this change before we bump major gem versions

vfonic:
I was really hoping this change got merged into the new major version bump... :(

vfonic:
I just saw this got merged in v11. Wow! Thank you so much, for making a major version bump just for this PR! I really appreciate it! ❤️ 🎉

chrisbutcher:
No worries! Sorry it slipped from our list of changes to include in v10 :)

ric2b:
Could this be documented in the README as a breaking change when upgrading to v11? It's only mentioned in the changelog.

PR #587, opened 2018-05-10
Update spring to be compatible with changes

sumith:
Gem.path does not take parameters.  Fixing to be compliant with latest changes


EiNSTeiN-:
thanks!

PR #617, opened 2018-08-10
Top level partitioning for safari 12

ragalie:
Webkit is releasing ITP 2.0 as part of the OS updates scheduled for Fall 2018. These updates, which will affect merchants using Safari 12, break embedded apps in Shopify admin by preventing apps from using cookies in the iframe. Apps rely on these cookies to authenticate the merchant when the app loads in an iframe.

This PR works around ITP 2.0 by triggering "cookie partitioning". Partitioning is triggered when the user interacts with a domain at the top-level before interacting with it in the iframe. In the partitioned state apps are able to use cookies in the iframe, but those cookies aren't available outside of the iframe context.

These limitations required us to implement a more convoluted flow to ensure that we auth correctly. That flow is diagrammed below, and this PR implements that flow:

image

### DIFF
--- a/app/assets/javascripts/shopify_app/itp_polyfill.js
+++ b/app/assets/javascripts/shopify_app/itp_polyfill.js
@@ -1,0 +1,16 @@
+function setCookieAndRedirect() {
+  document.cookie = "shopify.cookies_persist=true";
+  window.location.href = window.shopOrigin + "/admin/apps/" + window.apiKey;
+}
+
+document.addEventListener("DOMContentLoaded", function() {
+  if (document.hasStorageAccess) {
+    var itpContent = document.querySelector('#CookiePartitionPrompt');
+    itpContent.style.display = 'block';
+
+    var button = document.querySelector('#AcceptCookies');
+    button.addEventListener('click', setCookieAndRedirect);
+  } else {
+    setCookieAndRedirect();
+  }
+});
--- a/app/assets/javascripts/shopify_app/itp_polyfill.js
+++ b/app/assets/javascripts/shopify_app/itp_polyfill.js
@@ -1,16 +1,30 @@
-function setCookieAndRedirect() {
-  document.cookie = "shopify.cookies_persist=true";
-  window.location.href = window.shopOrigin + "/admin/apps/" + window.apiKey;
-}
-
-document.addEventListener("DOMContentLoaded", function() {
-  if (document.hasStorageAccess) {
-    var itpContent = document.querySelector('#CookiePartitionPrompt');
-    itpContent.style.display = 'block';
-
-    var button = document.querySelector('#AcceptCookies');
-    button.addEventListener('click', setCookieAndRedirect);
-  } else {
-    setCookieAndRedirect();
+(function() {
+  function setCookieAndRedirect() {
+    document.cookie = "shopify.cookies_persist=true";
+    window.location.href = window.shopOrigin + "/admin/apps/" + window.apiKey;
   }
-});
+
+  function shouldDisplayPrompt() {
+    if (navigator.userAgent.indexOf('com.jadedpixel.pos') !== -1) {
+      return false;
+    }
+
+    if (navigator.userAgent.indexOf('Shopify Mobile/iOS') !== -1) {
+      return false;
+    }
+
+    return Boolean(document.hasStorageAccess);
+  }
+
+  document.addEventListener("DOMContentLoaded", function() {
+    if (shouldDisplayPrompt()) {
+      var itpContent = document.querySelector('#CookiePartitionPrompt');
+      itpContent.style.display = 'block';
+
+      var button = document.querySelector('#AcceptCookies');
+      button.addEventListener('click', setCookieAndRedirect);
+    } else {
+      setCookieAndRedirect();
+    }
+  });
+})();
--- a/app/assets/javascripts/shopify_app/redirect.js
+++ b/app/assets/javascripts/shopify_app/redirect.js
@@ -1,31 +1,33 @@
-function redirect() {
-  var redirectTargetElement = document.getElementById("redirection-target");
+(function() {
+  function redirect() {
+    var redirectTargetElement = document.getElementById("redirection-target");
 
-  if (!redirectTargetElement) {
-    return;
+    if (!redirectTargetElement) {
+      return;
+    }
+
+    var targetInfo = JSON.parse(redirectTargetElement.dataset.target)
+
+    if (window.top == window.self) {
+      // If the current window is the 'parent', change the URL by setting location.href
+      window.top.location.href = targetInfo.url;
+    } else {
+      // If the current window is the 'child', change the parent's URL with postMessage
+      normalizedLink = document.createElement('a');
+      normalizedLink.href = targetInfo.url;
+
+      data = JSON.stringify({
+        message: 'Shopify.API.remoteRedirect',
+        data: {location: normalizedLink.href}
+      });
+      window.parent.postMessage(data, targetInfo.myshopifyUrl);
+    }
   }
 
-  var targetInfo = JSON.parse(redirectTargetElement.dataset.target)
+  document.addEventListener("DOMContentLoaded", redirect);
 
-  if (window.top == window.self) {
-    // If the current window is the 'parent', change the URL by setting location.href
-    window.top.location.href = targetInfo.url;
-  } else {
-    // If the current window is the 'child', change the parent's URL with postMessage
-    normalizedLink = document.createElement('a');
-    normalizedLink.href = targetInfo.url;
-
-    data = JSON.stringify({
-      message: 'Shopify.API.remoteRedirect',
-      data: { location: normalizedLink.href }
-    });
-    window.parent.postMessage(data, targetInfo.myshopifyUrl);
-  }
-}
-
-document.addEventListener("DOMContentLoaded", redirect);
-
-// In the turbolinks context, neither DOMContentLoaded nor turbolinks:load
-// consistently fires. This ensures that we at least attempt to fire in the
-// turbolinks situation as well.
-redirect();
+  // In the turbolinks context, neither DOMContentLoaded nor turbolinks:load
+  // consistently fires. This ensures that we at least attempt to fire in the
+  // turbolinks situation as well.
+  redirect();
+})();
--- a/app/assets/javascripts/shopify_app/redirect.js
+++ b/app/assets/javascripts/shopify_app/redirect.js
@@ -1,5 +1,10 @@
-document.addEventListener("DOMContentLoaded", function() {
+function redirect() {
   var redirectTargetElement = document.getElementById("redirection-target");
+
+  if (!redirectTargetElement) {
+    return;
+  }
+
   var targetInfo = JSON.parse(redirectTargetElement.dataset.target)
 
   if (window.top == window.self) {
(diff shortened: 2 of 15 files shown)


app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -36,10 +36,14 @@ def destroy
 
     private
 
-    def authenticate
+    def authenticate      
       if sanitized_shop_name.present?
         session['shopify.omniauth_params'] = { shop: sanitized_shop_name }
-        fullpage_redirect_to "#{main_app.root_path}auth/shopify"
+        if session['shopify.cookies_persist'] || !ShopifyApp.configuration.embedded_app?
```
ragalie: I think there's a subtle bug here.

When we're in fullpage mode, we might already have the `shopify.cookies_persist` cookie, which would mean we'd take this path. But the whole reason we're in fullpage mode is that we want to trigger a "fullpage redirect", since that's where our ITP logic currently lives.

We should separate these two concerns by fullpage redirecting to a `set_top_level_cookie` action or passing a similarly-named parameter that will always trigger the top-level cookie (ITP) logic.

app/assets/javascripts/shopify_app/itp_polyfill.js
```diff
@@ -0,0 +1,8 @@
+function shouldTriggerCookiePartitioning() {
+  var userAgent =  window.navigator.userAgent;
+  var isSafari = userAgent.indexOf('Safari');
+  var versionNumber = isSafari ? parseFloat(userAgent.match(/Version\/(\d+\.?\d*)/)[1]) : null;
+  
+  // TODO: Replace with library for checking user-agents
+  return document.hasStorageAccess && isSafari && versionNumber >= 12;
```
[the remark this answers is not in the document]
davidmuzi: What about on Mobile?
POS: `com.jadedpixel.pos Shopify POS/4.8.3 (iPhone; iOS 11.4; Scale/3.00) MobileMiddlewareSupported`
Shopify Mobile:`Shopify Mobile/iOS/7.6.0 (iPhone10,5 Simulator/com.shopify.ShopifyInternal/11.4.0)`
Safari: `Mozilla/5.0 (iPhone; CPU iPhone OS 12_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Mobile/15E148 Safari/604.1`

Do you have a way to test the behaviour on an iOS 12 device? Ive thus far been unable to see auth broken on my device running iOS 12 b6.

cc. @andrewapperley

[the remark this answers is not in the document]
andrewapperley: @davidmuzi I've also not seen apps broken. I'm running the latest iOS 12 beta as well. If we're using UserAgent checks then we would need to check for `iOS 12` as well. I'll reply here if I see any announcement from Apple about ITP 2.0 on iOS or if its in the release notes of a new beta.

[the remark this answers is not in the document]
EiNSTeiN-: I agree that hasStorageAccess should be sufficient, otherwise this may miss other webkit based browsers that implement `hasStorageAccess`. Since storage access is baked right into webkit I expect other browsers will be affected over time. The list of webkit-based browsers is long, including Chrome on iOS according to this: https://en.wikipedia.org/wiki/List_of_web_browsers#WebKit-based

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -57,14 +57,14 @@ def clear_shop_session
       session[:shopify_user] = nil
     end
 
-    def login_url
+    def login_url(no_cookie_redirect = false)
```
vfonic: Would be better if `no_cookie_redirect` was a [keyword argument](https://robots.thoughtbot.com/ruby-2-keyword-arguments).

Right now the API to call this method is: https://github.com/Shopify/shopify_app/pull/617/files#diff-350e7187d4b8bbcd1eaee3ff44fc7fb3R68
`login_url(true)`
while, with keyword argument, it would change to:
`login_url(no_cookie_redirect: true)`

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -5,19 +5,19 @@ module LoginProtection
     class ShopifyDomainNotFound < StandardError; end
 
     included do
+      after_action :set_test_cookie if ShopifyApp.configuration.embedded_app?
```
vfonic: I don't see any tests added for these changes (just some slight modifications). Can you please add some?

config/locales/ja.yml
```diff
@@ -1,3 +1,7 @@
 ja:
   logged_out: 'ログインに成功しました'
   could_not_log_in: 'Shopifyストアにログインできませんでした'
+  enable_cookies_heading: "Enable cookies from %{app}"
+  enable_cookies_body: "You must manually enable cookies in this browser in order to use %{app} within Shopify."
+  enable_cookies_footer: 'Cookies let the app authenticate you by temporarily storing your preferences and personal information. They expire after 30 days.'
+  enable_cookies_action: 'Enable cookies'
```
vfonic: These all are in English (I guess you know that :) ), and you're missing a new line here.

app/assets/javascripts/shopify_app/itp_polyfill.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+function setCookieAndRedirect() {
+  document.cookie = "shopify.cookies_persist=true";
+  window.location.href = window.shopOrigin + "/admin/apps/" + window.apiKey;
+
+  return false;
```
vfonic: You don't need this. The button you've attached the "click" event at, it's `type="button"`, therefore won't do any for submission or navigation. `return false` to prevent clicks is generally considered to be a bad practice (due to stopping event propagation)

app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -14,6 +14,10 @@ def create
       authenticate
     end
 
+    def enable_cookies
+      @shop = sanitize_shop_param(params)
```
vfonic: You could just reuse (call) `sanitized_shop_name` from `LoginProtection`. It does almost exactly and I like to avoid using `@shop` variable name as a name for shop's domain.

app/views/shopify_app/sessions/enable_cookies.html.erb
```diff
@@ -0,0 +1,66 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <link rel="stylesheet" href="https://sdks.shopifycdn.com/polaris/2.5.0/polaris.min.css" />
```
vfonic: This "redirect" page is loading 134KB of styles. Can you please extract the required styles and inline them like you did right here below?

app/views/shopify_app/sessions/enable_cookies.html.erb
```diff
@@ -0,0 +1,66 @@
...
+
+    .Polaris-Page {
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      height: 100vh;
+    }
+  </style>
+  <base target="_top">
+  <title>Redirecting…</title>
```
vfonic: Nitpick: You could use `...` (three characters) instead of `…` (one character)

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -5,19 +5,19 @@ module LoginProtection
...
-        begin
-          ShopifyAPI::Base.activate_session(shop_session)
-          yield
-        ensure
-          ShopifyAPI::Base.clear_session
-        end
-      else
-        redirect_to_login
+      return redirect_to_login unless shop_session
+      clear_top_level_oauth_cookie
```
vfonic: Is this necessary to call on every request?

Maybe it's better to simply ignore the cookie after you no longer need it? Or clear it somewhere during the authentication/cookie-setting process, once.

config/locales/en.yml
```diff
@@ -2,3 +2,7 @@ en:
   logged_out: 'Successfully logged out'
   could_not_log_in: 'Could not log in to Shopify store'
   invalid_shop_url: 'Invalid shop domain'
+  enable_cookies_heading: "Enable cookies from %{app}"
+  enable_cookies_body: "You must manually enable cookies in this browser in order to use %{app} within Shopify."
+  enable_cookies_footer: 'Cookies let the app authenticate you by temporarily storing your preferences and personal information. They expire after 30 days.'
```
vfonic: Do cookies really expire after 30 days? I think this depends on the app, and, as far as I can tell, rails doesn't set expiration on cookies. Sorry I couldn't find anything more useful. :/

https://github.com/rails/rails/blob/39cb84e8bdd4b9a35d68b33222a174d2350aee89/actionpack/lib/action_dispatch/middleware/session/cookie_store.rb#L44

EiNSTeiN-: Hey @vfonic, this is enforced by ITP: https://webkit.org/blog/8311/intelligent-tracking-prevention-2-0/, and is the reason this branch is being worked on.

lib/shopify_app/controller_concerns/login_protection.rb
ragalie: We can't ignore it; we need to remove it or there's a risk that we won't run the full flow the next time we can't recover the session.

We need to wait until we confirm that we've actually recovered the session to remove the cookie, and this seems like the least bad place to do that. We could check that the cookie doesn't exist, but that's probably slower than just deleting the key from the hash regardless.

ragalie:
One thing I think we should do: we should check the user agent for POS/mobile and skip the "enable cookies" button in that case.

We know that ITP doesn't apply because we're not using SafariViewControllers to render the embedded apps, but there's no easy way for the app to know that.

I know we were opposed to this for the purposes of determining whether ITP applied _in general_, but if the user agents are stable enough I think it makes sense in this context.

francinen:
cc @ShayneP

app/assets/javascripts/shopify_app/itp_polyfill.js
```diff
@@ -0,0 +1,16 @@
+function setCookieAndRedirect() {
```
tylerball: it may be wise to wrap all this code in an IIFE just for safety and to avoid interference from browser extensions, etc.

app/assets/javascripts/shopify_app/redirect.js
```diff
@@ -1,5 +1,10 @@
-document.addEventListener("DOMContentLoaded", function() {
+function redirect() {
```
tylerball: same suggestion as above

app/views/shopify_app/sessions/enable_cookies.html.erb
```diff
@@ -0,0 +1,388 @@
...
+      background:linear-gradient(to bottom, #bac0e6, #bac0e6);
+      border-color:#a7aedf;
+      box-shadow:none;
+      color:white;
+    }
+  </style>
+  <base target="_top">
+  <title>Redirecting…</title>
+
+  <script src="https://cdn.shopify.com/s/assets/external/app.js?<%= Time.now.strftime('%Y%m%d%H') %>"></script>
```
tylerball: Do we need to load the EASDK here since we're not initializing it?

app/views/shopify_app/sessions/enable_cookies.html.erb
```diff
@@ -0,0 +1,388 @@
+<!DOCTYPE html>
+<html lang="en">
```
nwtn: presumably this should change based on locale

app/views/shopify_app/sessions/enable_cookies.html.erb
```diff
@@ -0,0 +1,388 @@
...
+<head>
+  <meta charset="utf-8" />
+  <style>
```
nwtn: This feels super fragile. I’m assuming you’re embedding this here for performance, so we’re not loading all the Polaris CSS for a single redirect page?

ragalie: Yes, I think that was the logic. I know we got some pushback around including all of the CSS given the size, especially for something that's relatively static like this page.

PR #629, opened 2018-09-05
Removing the read_orders scope from list of default scopes provided to a generated app

TerencedzLi:
With reference to: [Issue](https://github.com/Shopify/shopify/issues/166090)

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ ShopifyApp.configure do |config|
   config.application_name = 'Your app name' # Optional
   config.api_key = ENV['SHOPIFY_CLIENT_API_KEY']
   config.secret = ENV['SHOPIFY_CLIENT_API_SECRET']
-  config.scope = 'read_customers, read_orders, write_products'
+  config.scope = 'read_customers, write_products'
   config.embedded_app = true
 end
 ```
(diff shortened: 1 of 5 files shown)


vfonic:
There are couple of more `read_orders` in the scope throughout the codebase. To find them all, it's the best to do a project-wide search for `read_orders`.

TerencedzLi:
AFAIK, the remaining references are either to showcase usage of scopes (in docs) or a part of some tests that allow you to specify scopes to be used (i.e for CLI)

tylerball:
I don't have a lot of context here but I think we should have a plan to update those docs as well.

TerencedzLi:
Note to self: Here's one part of the documentation that needs to be updated. https://help.shopify.com/en/api/getting-started/authentication/public-authentication

jamiemtdwyer:
> Note to self: Here's one part of the documentation that needs to be updated. https://help.shopify.com/en/api/getting-started/authentication/public-authentication
> 
> Node example docs only use read_products. https://help.shopify.com/en/api/tutorials/building-> node-app

I don't see any reason why any documentation on the help site would need to be updated, including either of these.

There are no tutorials or guides on the help site that use `shopify_app`.

TerencedzLi:
> > Note to self: Here's one part of the documentation that needs to be updated. https://help.shopify.com/en/api/getting-started/authentication/public-authentication
> > Node example docs only use read_products. [https://help.shopify.com/en/api/tutorials/building->](https://help.shopify.com/en/api/tutorials/building-%3E) node-app
> 
> I don't see any reason why any documentation on the help site would need to be updated, including either of these.
> 
> There are no tutorials or guides on the help site that use `shopify_app`.

Yeaup, those links I put are irrelevant. But as a whole effort, as seen in the initial issue, we wanted to remove read_orders as a default scope. I agree probably not relevant to this gem :) thanks for the feedback

PR #636, opened 2018-09-21
Adds the ability to configure the login_url

meulmees:
### Problem

While working on our embedded app we ran into a requirement where we want to redirect users to a separate landing page in order to go through the login flow.

Currently the `shopify_app` gem redirects to `login_url` but that's hard coded to be essentially `"#{root_url}/login"`

### Solution

This PR adds a `login_url` `attr_accessor` so that clients can configure the login URL. It defaults to the old behaviour until you've customized the URL.

ex:

```ruby
  config.login_url = 'https://my.domain.com/some/route/login
end
```

### DIFF
--- a/test/shopify_app/controller_concerns/login_protection_test.rb
+++ b/test/shopify_app/controller_concerns/login_protection_test.rb
@@ -124,6 +124,15 @@ class LoginProtectionTest < ActionController::TestCase
     end
   end
 
+  test '#shopify_session with no Shopify session, redirects to a custom config login url' do
+    with_custom_login_url 'https://domain.com/custom/route/login' do
+      with_application_test_routes do
+        get :index, params: { shop: 'foobar' }
+        assert_redirected_to 'https://domain.com/custom/route/login?shop=foobar.myshopify.com'
+      end
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


meulmees:
@EiNSTeiN- I don't see any security issues, but just wanted to get your 👀 on it

test/shopify_app/controller_concerns/login_protection_test.rb
```diff
@@ -255,4 +288,12 @@ def with_application_test_routes
       yield
     end
   end
+
+  def with_custom_login_url(url)
+    original_url = ShopifyApp.configuration.login_url.dup
+
+    ShopifyApp.configure { |config| config.login_url = url }
+    yield
+    ShopifyApp.configure { |config| config.login_url = original_url }
```
EiNSTeiN-: I believe this last line should be behind `ensure` so the url is restored even if an exception occurs

EiNSTeiN-:
Perhaps a line in the readme about this would be good.

test/shopify_app/controller_concerns/login_protection_test.rb
meulmees: Ah yep, thanks!

PR #647, opened 2018-10-18
Implement ITP 2.1 solution with the Storage Access API for Safari 12.1+ users

francinen:
Closes https://github.com/Shopify/shopify/issues/174867

## Context
**TL;DR: ITP 2.1 will require us to use the Storage Access API to ensure embedded apps can access authentication cookies in the iframe. WebKit will release an updated version of the API that allows access to persist for same-origin navigation in iframes. The relaxed Storage Access API, as well as ITP 2.1, will be rolled out with Safari 12.1 in spring 2019.**

When we started exploring possible solutions for ITP 2.0, we ruled out the Storage Access API because, at the time, it did not work for multi-page embedded apps: it did not allow storage access to persist for same- or cross-origin navigation within an iframe.

After we released our ITP 2.0 solution, which relies on the ability to partition cookies by getting a user to interact with the app's top-level domain, we learned WebKit would introduce two updates to ITP 2.1:
1. They will be removing the partitioning mechanism (which breaks our current implementation for ITP 2.0)
2. They are relaxing their Storage Access API rules to allow access to persist in iframes for _same-origin_ navigation (which means it will work for multi-page embedded apps, insofar as they do not do cross-origin navigation)

These updates will be available in Safari 12.1, which is expected to be released in spring 2019. This means that we still need to execute our existing ITP 2.0 implementation for merchants using Safari 12.0.

## Solution
**TL;DR: Use the Storage Access API if the user agent is Safari 12.1+. For Safari 12.0, continue executing our solution for ITP 2.0, which relies on cookie partitioning.**

Screen capture: https://drive.google.com/a/shopify.com/file/d/1WiUUVBwWzsCepJqJnN_HAHQMtMJPjYPE/view?usp=sharing

The Storage Access API provides the following methods to manage access to third-party cookies:

- `document.hasStorageAccess`: A promise that returns `true` if a domain can access its cookies in a third-party context. It always returns `true` if you are on the app's top-level domain.

- `document.requestStorageAccess`: When called, it will render a browser prompt that gives users the option to allow or not allow a domain to access its cookies in a third-party context. If the user clicks `Allow`, the promise will resolve and storage access is granted for ~30 days. If they click `Don't allow`, the promise rejects and storage access is denied. The promise will also reject if the user has not interacted with the app's top-level domain in the last 30 days.

The following conditions must be met in order to successfully request storage access using the Storage Access API:
1. The user must have _interacted with the app's top-level domain_ in the last 30 days in order to render a browser prompt when `document.requestStorageAccess` is called.
2. Calling `document.requestStorageAccess` must be _triggered by a user gesture_ in the third-part context (e.g., the iframe).

These conditions require user gestures in two interfaces: the app's top-level domain and in the iframe.

With these constraints in mind, this is the workflow we are proposing for ITP 2.1:

If the merchant has not interacted with the app in the last 30 days and cookies from the app's domain is blocked by ITP...
1. Merchant clicks on link to embedded app from the Apps index page
2. When we load the embedded app view (empty iframe), we call `document.hasStorageAccess` to check if the app can access its cookies in the iframe. If it does have storage access, it will load as expected.
3. If the app does not have storage access, we will check if the merchant has interacted with the app's top-level domain (using a `shopify.top_level_interaction` item we set in sessionStorage).
4. If the identifier is not found in sessionStorage, we redirect them to a view where they can interact with the app's TLD.  **This is required for the `document.requestStorageAccess` promise to resolve.**
5. Merchant clicks the `Continue` button on the app's TLD, which sets the `shopify.top_level_interaction` item in sessionStorage to denote that the merchant has performed the user gesture, and redirects them to the embedded context.
image

6. Back in the embedded context, we call `document.hasStorageAccess`, which still returns `false`. But now when we check for the `shopify.top_level_interaction` item in sessionStorage, we find it is defined, which tells us that we can successfully call `document.requestStorageAccess`.
7. Still in the embedded context, we render an interface for the merchant to allow the app to request storage access. When they click `Continue`, we call `document.requestStorageAccess` and it renders a prompt for the merchant to allow or not allow cookies from the app.
image

8. If the merchant clicks `Allow`, the app will load in the iframe. If the merchant clicks `Don't allow`, they will be redirected to the Apps Index page.
image

Screen capture (Allows storage access): https://drive.google.com/a/shopify.com/file/d/1WiUUVBwWzsCepJqJnN_HAHQMtMJPjYPE/view?usp=sharing

Screen capture (Does not allow storage access): https://drive.google.com/a/shopify.com/file/d/1uL_jqpwqRoDV-w01DjIVbNHfR6X8e-3t/view?usp=sharing

**The merchant does not need to go through this entire flow again for another 30 days. If they clear their browser history before the 30-day window ends, they will need to repeat all 8 steps again.**

If the merchant has interacted with the app in the last 30 days and has already allowed cookies from the app, _but_ navigates to another page in the Shopify admin (or away from Shopify)...
1. Merchant clicks on link to embedded app from Apps Index
2. When we call `document.hasStorageAccess`, it returns `false` because storage access does not persist for cross-origin navigation. But, if we check sessionStorage, we see that `shopify.top_level_interaction` is defined in sessionStorage. This means we do not have to redirect them to the app's TLD; we can simply render the interface for the merchant to allow the app to request storage access in the iframe.
image

3. Merchant clicks `Continue`, and we call `document.requestStorageAccess`. Because they've previously clicked `Allow` in the browser prompt, it doesn't need to render the prompt again, and simply loads the embedded app in the iframe.

You can refer to these mockups for a simplified version of the UX flow: https://sketch.cloud/s/baVw4

**This workflow will only affect merchants if the app's domain has been classified by ITP. ITP runs on a machine learning model, so it could affect different apps at different times.**

## Tophatting
1. You will need to set up a [test app using the `shopify_app` gem](https://github.com/Shopify/shopify_app#becoming-a-shopify-app-developer). After you set it up, install it on your store.
2. Checkout this branch of the gem. Re-start your server in the repo for your test app.
3. If you haven't already, download [Safari Technology Preview](https://developer.apple.com/safari/download/) and [enable ITP Debug Mode](https://webkit.org/blog/8387/itp-debug-mode-in-safari-technology-preview-62/)
3. In another tab/window in your Terminal, run the following command to tell ITP to block cookies from your test app's domain (Replace `1234.ngrok.io` with your ngrok domain):
```
defaults write com.apple.SafariTechnologyPreview ResourceLoadStatisticsManualPrevalentResource 1234.ngrok.io
```
4. Run the following command to start your ITP Debug logs:
```
log stream -info | grep ResourceLoadStatisticsDebug
```
5. When you open Safari Technology Preview, check the logs to confirm that your app's domain has been classified by ITP:
image

6. Login to your Shopify admin account and click on the link to your test app in the Apps Index page.

7. Go through the UX flow outlined above.

## TO-DO's:
- Blocker: Our backend changes appear to break the app installation flow. We are looking into it now.
- We are working with the Illustration team to get a more appropriate visual to render in the top level interaction page. This is not a blocker.


tylerball:
I setup a dummy app to test this out and I'm getting an error on shop install:

2018-10-18 at 5 36 pm

logs:
```
Started POST "/login" for 184.94.110.66 at 2018-10-18 17:34:29 -0400
Cannot render console from 184.94.110.66! Allowed networks: 127.0.0.1, ::1, 127.0.0.0/127.255.255.255
Processing by ShopifyApp::SessionsController#create as HTML
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"k6/Ew6grce6gkDfDjioUAi2tfWgbPqq3eF+3Lf7SfyersSVFVA5SOdUFH/J43Xprctj8SUEGUgf/GKoSgqgS9g==", "shop"=>"tb-test.myshopify.com"}
  Rendering /Users/tball/src/github.com/Shopify/shopify_app/app/views/shopify_app/sessions/request_storage_access.html.erb
  Rendered /Users/tball/src/github.com/Shopify/shopify_app/app/views/shopify_app/sessions/request_storage_access.html.erb (308.2ms)
Completed 200 OK in 314ms (Views: 312.7ms | ActiveRecord: 0.0ms)


Started GET "/undefined" for 184.94.110.66 at 2018-10-18 17:34:30 -0400
Cannot render console from 184.94.110.66! Allowed networks: 127.0.0.1, ::1, 127.0.0.0/127.255.255.255

ActionController::RoutingError (No route matches [GET] "/undefined"):

actionpack (5.0.7) lib/action_dispatch/middleware/debug_exceptions.rb:53:in `call'
web-console (2.3.0) lib/web_console/middleware.rb:20:in `block in call'
web-console (2.3.0) lib/web_console/middleware.rb:18:in `catch'
web-console (2.3.0) lib/web_console/middleware.rb:18:in `call'
actionpack (5.0.7) lib/action_dispatch/middleware/show_exceptions.rb:31:in `call'
railties (5.0.7) lib/rails/rack/logger.rb:36:in `call_app'
railties (5.0.7) lib/rails/rack/logger.rb:24:in `block in call'
activesupport (5.0.7) lib/active_support/tagged_logging.rb:69:in `block in tagged'
activesupport (5.0.7) lib/active_support/tagged_logging.rb:26:in `tagged'
activesupport (5.0.7) lib/active_support/tagged_logging.rb:69:in `tagged'
railties (5.0.7) lib/rails/rack/logger.rb:24:in `call'
actionpack (5.0.7) lib/action_dispatch/middleware/request_id.rb:24:in `call'
rack (2.0.5) lib/rack/method_override.rb:22:in `call'
rack (2.0.5) lib/rack/runtime.rb:22:in `call'
activesupport (5.0.7) lib/active_support/cache/strategy/local_cache_middleware.rb:28:in `call'
actionpack (5.0.7) lib/action_dispatch/middleware/executor.rb:12:in `call'
actionpack (5.0.7) lib/action_dispatch/middleware/static.rb:136:in `call'
rack (2.0.5) lib/rack/sendfile.rb:111:in `call'
railties (5.0.7) lib/rails/engine.rb:522:in `call'
rack (2.0.5) lib/rack/handler/webrick.rb:86:in `service'
/opt/rubies/2.5.1/lib/ruby/2.5.0/webrick/httpserver.rb:140:in `service'
/opt/rubies/2.5.1/lib/ruby/2.5.0/webrick/httpserver.rb:96:in `run'
/opt/rubies/2.5.1/lib/ruby/2.5.0/webrick/server.rb:307:in `block in start_thread'
  Rendering /Users/tball/.gem/ruby/2.5.1/gems/actionpack-5.0.7/lib/action_dispatch/middleware/templates/rescues/routing_error.html.erb within rescues/layout
  Rendering /Users/tball/.gem/ruby/2.5.1/gems/actionpack-5.0.7/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb
  Rendered /Users/tball/.gem/ruby/2.5.1/gems/actionpack-5.0.7/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb (2.7ms)
  Rendered collection of /Users/tball/.gem/ruby/2.5.1/gems/actionpack-5.0.7/lib/action_dispatch/middleware/templates/routes/_route.html.erb [12 times] (5.8ms)
  Rendered collection of /Users/tball/.gem/ruby/2.5.1/gems/actionpack-5.0.7/lib/action_dispatch/middleware/templates/routes/_route.html.erb [8 times] (4.2ms)
  Rendered /Users/tball/.gem/ruby/2.5.1/gems/actionpack-5.0.7/lib/action_dispatch/middleware/templates/routes/_table.html.erb (20.7ms)
  Rendering /Users/tball/.gem/ruby/2.5.1/gems/actionpack-5.0.7/lib/action_dispatch/middleware/templates/rescues/_request_and_response.html.erb
  Rendered /Users/tball/.gem/ruby/2.5.1/gems/actionpack-5.0.7/lib/action_dispatch/middleware/templates/rescues/_request_and_response.html.erb (1.2ms)
  Rendered /Users/tball/.gem/ruby/2.5.1/gems/actionpack-5.0.7/lib/action_dispatch/middleware/templates/rescues/routing_error.html.erb within rescues/layout (382.1ms)
```

francinen:
@tylerball We're aware of the issue and listed in as a To-Do near the end of the PR description. Sorry if it wasn't apparent enough! I should have communicated it more clearly. Mike and I will investigate it on Monday. In the meantime, you can checkout master and install the app on your shop before checking out this one.

tylerball:
@francinen No worries, I didn't read the description thoroughly enough, that's why I deleted the comment.

app/views/shopify_app/sessions/request_storage_access.html.erb
```diff
@@ -0,0 +1,395 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <base target="_top">
+  <title>Redirecting…</title>
+  <style>
+    html,
```
tylerball: I'm rethinking my suggestion about inlining the styles since we have to do it three times now.

app/assets/javascripts/shopify_app/storage_access.js
```diff
@@ -0,0 +1,114 @@
+var ACCESS_GRANTED_STATUS = 'access granted';
```
tylerball: This file also isn't wrapped in an IIFE

app/assets/javascripts/shopify_app/storage_access.js
```diff
@@ -0,0 +1,114 @@
...
+var ACCESS_DENIED_STATUS = 'access denied';
+
+function StorageAccessHelper(redirectInfo) {
+  this.redirectInfo = redirectInfo;
```
tylerball: Not crazy about the word `redirectInfo` as a name for a configuration-like object. Maybe `config` or `redirectData`.

app/assets/javascripts/shopify_app/storage_access.js
```diff
@@ -0,0 +1,114 @@
...
+
+StorageAccessHelper.prototype.setUpCookiePartitioning = function() {
+  var itpContent = document.querySelector('#CookiePartitionPrompt');
+  itpContent.style.display = 'block';
+
+  var button = document.querySelector('#AcceptCookies');
+  button.addEventListener('click', this.setCookieAndRedirect.bind(this));
+}
+
+
```
tylerball: extra lines

app/assets/javascripts/shopify_app/top_level_interaction.js
```diff
@@ -0,0 +1,21 @@
+(function() {
+  function setUpTopLevelInteraction() {
+    var TopLevelInteraction = new ITPHelper({
+      content: '#TopLevelInteractionContent',
+      action: '#TopLevelInteractionButton',
```
tylerball: I wouldn't consider this code resusable so I wouldn't bother making these selectors configurable, just make them constants in the ITPHelper klass.

app/assets/javascripts/shopify_app/storage_access_redirect.js
```diff
@@ -0,0 +1,22 @@
...
+        storageAccessHelper.execute();
+    }
+  }
+
+  document.addEventListener("DOMContentLoaded", redirect);
+
+  // In the turbolinks context, neither DOMContentLoaded nor turbolinks:load
+  // consistently fires. This ensures that we at least attempt to fire in the
+  // turbolinks situation as well.
+  redirect();
```
tylerball: What about just including this script at the bottom of `<body>` like oldentimes?

test/javascripts/shopify_app/storage_access_test.js
```diff
@@ -0,0 +1,306 @@
...
+      var canPartitionCookiesStub = sinon.stub(ITPHelper.prototype, 'canPartitionCookies').callsFake(() => true);
+
+      const setUpCookiePartitioningStub = sinon.stub(storageAccessHelper, 'setUpCookiePartitioning');
+
+      storageAccessHelper.execute();
+
+      sinon.assert.called(setUpCookiePartitioningStub);
+
+      canPartitionCookiesStub.restore();
+      setUpCookiePartitioningStub.restore();
```
tylerball: Look into using [sinon sandboxes](https://sinonjs.org/releases/latest/sandbox/) to eliminate the need to restore stubs.

app/assets/javascripts/shopify_app/top_level_interaction.js
```diff
@@ -0,0 +1,21 @@
...
+
+    if (!TopLevelInteraction.itpContent) {
+      return;
+    }
+
+    if (TopLevelInteraction.userAgentIsAffected()) {
+      TopLevelInteraction.setUpContent();
+    } else {
+      TopLevelInteraction.redirect();
+    }
```
tylerball: I think we could move everything from line 9 to here into `ITPHelper` itself, similar to `execute` in `StorageAccessHelper`, and just have DOMContentLoaded call it. I think it would make it easier to follow what's happening.

app/assets/javascripts/shopify_app/storage_access.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+}
+
+/* ITP 2.0 solution: handles cookie partitioning */
+StorageAccessHelper.prototype.setUpHelper = function() {
+  return new ITPHelper({redirectUrl: window.shopOrigin + "/admin/apps/" + window.apiKey});
+}
+
+StorageAccessHelper.prototype.setCookieAndRedirect = function() {
+  document.cookie = "shopify.cookies_persist=true";
```
tylerball: What's the difference between what happens here and in `partition_cookies.js`? Can it be combined somehow?

app/assets/javascripts/shopify_app/itp_polyfill.js
```diff
@@ -1,30 +1,32 @@
...
+ITPHelper.prototype.redirect = function() {
+  sessionStorage.setItem('shopify.top_level_interaction', 'true');
+  window.location.href = this.redirectUrl;
+}
 
-    if (navigator.userAgent.indexOf('Shopify Mobile/iOS') !== -1) {
-      return false;
-    }
+ITPHelper.prototype.userAgentIsAffected = function() {
+  if (navigator.userAgent.indexOf('com.jadedpixel.pos') !== -1) {
```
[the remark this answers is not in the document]
ragalie: I think it would be more performant, and we're already doing user agent checking in the backend, so we might as well move this, agreed!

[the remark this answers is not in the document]
ragalie: Wouldn't we just not render this page if the user agent isn't affected?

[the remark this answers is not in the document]
francinen: We did not build it that way for this view. I will need to split it up into separate views if that's the case.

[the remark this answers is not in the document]
francinen: Will we be able to check if `document.hasStorageAccess` is true from the server?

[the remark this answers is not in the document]
francinen: For context, here is the full method:
```javascript
  
  ITPHelper.prototype.userAgentIsAffected = function() {
    if (navigator.userAgent.indexOf('com.jadedpixel.pos') !== -1) {
      return false;
    }
  
    if (navigator.userAgent.indexOf('Shopify Mobile/iOS') !== -1) {
      return false;
    }
  
    return Boolean(document.hasStorageAccess);
  }
```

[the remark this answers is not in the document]
ragalie: I think the idea is that we would remove the user agent checks here and move them to https://github.com/Shopify/shopify_app/pull/647/files#diff-350e7187d4b8bbcd1eaee3ff44fc7fb3R103

We would keep the `hasStorageAccess` check on the frontend.

app/assets/stylesheets/itp.css
```diff
@@ -0,0 +1,5 @@
+#RequestStorageAccess,
+#CookiePartitionPrompt,
+#TopLevelInteractionContent {
+  display: none;
+}
```
nwtn: why do these need to be in a separate CSS file? actually feels like we might be able to inline these to cut down on HTTP roundtrips.

app/assets/stylesheets/styles.css
```diff
@@ -0,0 +1,478 @@
+html,
```
nwtn: i get why you extracted this, but _if_ this + the rest of our header content is under 14kb it can be inlined so that it all comes through on the first http round trip, which would improve render performance.

app/views/shopify_app/sessions/enable_cookies.html.erb
```diff
@@ -2,350 +2,20 @@
...
+
   <script>
     window.apiKey = "<%= ShopifyApp.configuration.api_key %>";
     window.shopOrigin = "https://<%= @shop %>";
   </script>
 
-  <%= javascript_include_tag('shopify_app/itp_polyfill', crossorigin: 'anonymous', integrity: true) %>
+  <%= javascript_include_tag('shopify_app/itp_helper', crossorigin: 'anonymous', integrity: true) %>
+  <%= javascript_include_tag('shopify_app/storage_access', crossorigin: 'anonymous', integrity: true) %>
+  <%= javascript_include_tag('shopify_app/partition_cookies', crossorigin: 'anonymous', integrity: true) %>
```
nwtn: similar question to css stuff above — can we combine these into a single bundle to improve performance?

app/views/shopify_app/sessions/request_storage_access.html.erb
```diff
@@ -0,0 +1,62 @@
+<!DOCTYPE html>
+<html lang="en">
```
nwtn: `lang="<%= I18n.locale %>"`

app/assets/stylesheets/styles.css
francinen: @tylerball What do you think about keeping the CSS inline? https://github.com/Shopify/shopify_app/pull/647#discussion_r228277238

app/views/shopify_app/sessions/enable_cookies.html.erb
francinen: @tylerball Since we're using Webpack for our Karma tests, should we use it to bundle our JS as well?

ragalie: How would that work in practice, since this is a gem? Would the app developer need to set up their environments to run webpack in dev/prod?

app/assets/javascripts/shopify_app/partition_cookies.js
```diff
@@ -0,0 +1,13 @@
+(function() {
+  function setCookieAndRedirect() {
```
tylerball: I don't think this function gets called anymore since it's [in StorageAccessHelper](https://github.com/Shopify/shopify_app/pull/647/files#diff-57a97a7498955129995e2cfe66ab38f5R101) right?

app/assets/javascripts/shopify_app/storage_access_redirect.js
```diff
@@ -0,0 +1,17 @@
...
+  function redirect() {
+    var redirectTargetElement = document.getElementById("redirection-target");
+
+    var targetInfo = JSON.parse(redirectTargetElement.dataset.target)
+
+    if (window.top == window.self) {
+      // If the current window is the 'parent', change the URL by setting location.href
+      window.top.location.href = targetInfo.hasStorageAccessUrl;
+    } else {
+        var storageAccessHelper = new StorageAccessHelper(targetInfo);
```
tylerball: I think we can simplify this even further. Both times StorageAccessHelper is constructed `.execute()` is called on it right after. The only difference is the additional logic above that instead redirects immediately if called as the parent frame. Why don't we move this logic in the access helper itself, which could be moved into the helper itself and allow us to test it. I think that along with my [other comment above](https://github.com/Shopify/shopify_app/pull/647/files#r229396355) would allow us to get rid of this file and `partition_cookies.js` and save a couple http requests.

app/views/shopify_app/sessions/enable_cookies.html.erb
tylerball: Mike is right, we need to keep these as sprockets includes because this is a gem, but we could use `//= require`s inside of the scripts to only import a single JS file.

app/assets/javascripts/shopify_app/storage_access.js
```diff
@@ -0,0 +1,116 @@
+(function() {
+  var ACCESS_GRANTED_STATUS = 'access granted';
```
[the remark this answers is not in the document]
tylerball: It looks like it's being passed to `redirectToAppTLD` which is using it to construct a URL and send a postMessage which I guess redirects via the EASDK. Why not just use window.parent? It seems like it would simplify this a lot.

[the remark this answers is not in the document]
ragalie: I think the idea is that since the windows are on two different domains, we can't use `window.parent.location`, we have to send a message to the top frame asking it to redirect.

Gemfile
```diff
@@ -4,3 +4,4 @@ source "https://rubygems.org"
 gemspec
 
 gem 'rails-controller-testing', group: :test
+gem 'shopify_browser_sniffer', github: 'Shopify/shopify_browser_sniffer', ref: 'shopify-matchers-and-monkey-patch'
```
francinen: Waiting on this PR to ship so we can release the gem and make it available to `shopify_app`: https://github.com/Shopify/shopify_browser_sniffer/pull/2/files

app/views/shopify_app/sessions/request_storage_access.html.erb
nwtn: can we put them in a partial? reusability + inlining

PR #660, opened 2018-10-31
Use shop_domain for dynamic script tags example

karlcoelho:
I was rather confused with the `-> (domain)` parameter for the dynamic script tags config, thinking it was referring to the Rails app domain itself rather than the shop domain. `shop_domain` is used as a parameter and variable in a lot of the other examples so I thought let's keep it consistent with everywhere else.


nwtn:
Closing because the CLA hasn’t been signed. Please feel free to sign the CLA and then re-open.

PR #661, opened 2018-11-05
[WIP]Add old_client_secret to omniauth builder

theundeadmonk:


nwtn:
@theundeadmonk is this something we still need, or can this be closed?

PR #663, opened 2018-11-05
Add service.yml

nwtn:


nwtn:
not critical, but for future reference: it turns out the slack channel shouldn’t have the `#`

PR #664, opened 2018-11-06
Fixes issue with ITP circumvention and redirecting to requested URL

joshreeves:
I realize that the ITP issues and workarounds are a larger ongoing discussion; however, the problems this PR addresses required us to move our app onto a fork of the gem. I'd like to be able to switch back to the official release as soon as this issue is resolved.

The issue is that in user agents which are subject to the ITP workarounds (Safari 12, Safari WebView in the Shopify POS), the `session[:redirect_to]` value in `ShopifyApp::LoginProtection` is saved prior to the user agent allowing cookies to be set. Once the `fullpage_redirect_to enable_cookies_path` happens at the top level, the original `session[:return_to]` URL is lost. This means that if an unauthenticated user follows a deep link into the app (like from a `More Actions` link or from a POS integration link), then after authentication is successful they are returned to the root path of the app and **not** to the path they requested. This is a significant issue for us.

The changes in this PR ensure that the originally requested URL is saved throughout the redirection process, and the user is eventually taken to the appropriate content that was requested. If there's a simpler way of solving this, I'd love to hear it. If not, the changes below do solve the problem.

### DIFF
--- a/app/controllers/shopify_app/sessions_controller.rb
+++ b/app/controllers/shopify_app/sessions_controller.rb
@@ -54,6 +54,8 @@ module ShopifyApp
       return render_invalid_shop_error unless sanitized_shop_name.present?
       session['shopify.omniauth_params'] = { shop: sanitized_shop_name }
 
+      session[:return_to] = params[:return_to] if params[:return_to]
+
       if user_agent_can_partition_cookies
         authenticate_with_partitioning
       else
@@ -77,7 +79,7 @@ module ShopifyApp
         authenticate_in_context
       else
         set_top_level_oauth_cookie
-        fullpage_redirect_to enable_cookies_path(shop: sanitized_shop_name)
+        enable_cookie_access
       end
     end
 
@@ -94,6 +96,13 @@ module ShopifyApp
     def render_invalid_shop_error
       flash[:error] = I18n.t('invalid_shop_url')
       redirect_to return_address
+    end
+
+    def enable_cookie_access
+      fullpage_redirect_to(enable_cookies_path(
+        shop: sanitized_shop_name,
+        return_to: session[:return_to]
+      ))
     end
 
     def authenticate_in_context
--- a/app/assets/javascripts/shopify_app/storage_access.js
+++ b/app/assets/javascripts/shopify_app/storage_access.js
@@ -103,7 +103,7 @@
 
   /* ITP 2.0 solution: handles cookie partitioning */
   StorageAccessHelper.prototype.setUpHelper = function() {
-    return new ITPHelper({redirectUrl: window.shopOrigin + "/admin/apps/" + window.apiKey});
+    return new ITPHelper({redirectUrl: window.shopOrigin + "/admin/apps/" + window.apiKey + window.returnTo});
   }
 
   StorageAccessHelper.prototype.setCookieAndRedirect = function() {
--- a/app/views/shopify_app/sessions/enable_cookies.html.erb
+++ b/app/views/shopify_app/sessions/enable_cookies.html.erb
@@ -17,6 +17,7 @@
   <script>
     window.apiKey = "<%= ShopifyApp.configuration.api_key %>";
     window.shopOrigin = "https://<%= @shop %>";
+    window.returnTo = "<%= params[:return_to] %>"
   </script>
 
   <%= javascript_include_tag('shopify_app/enable_cookies', crossorigin: 'anonymous', integrity: true) %>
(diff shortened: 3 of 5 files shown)


francinen:
Hi there! Thanks for raising this issue. 

We will be shipping some updates to our current ITP solution soon that will address changes coming with ITP 2.1: https://github.com/Shopify/shopify_app/pull/647

Since we refactored a lot of the code this PR touches, we feel it would be better to hold off on merging this PR until we ship the updates for ITP 2.1.

In terms of the approach taken here, it might be possible to simplify it a bit. Instead of passing the `return_to` URL as a query parameter, could we redirect to something like `/apps/{api_key}/path/that/we/want/to/return/to`? This might eliminate the need to modify the `return_to` logic in `login_protection.rb`.

joshreeves:
Hi @francinen! Thanks for the response. That makes sense. I'll look for those updates.

As far as your suggestion, the problem is that after the `redirect_to_login` happens, both `session[:return_to]` and `request.referer` are `nil` in Safari 12.0.1. However, I did follow your recommendation with the JavaScript redirect; which, as you expected, did allow me to simplify things quite a bit. I made a second commit with those changes. The solution is simpler now, but still requires including the `return_to` param inside of `login_url_params` due to the fact that Safari 12 is not passing along the `HTTP referer` when redirecting to the login path.

francinen:
Hey @joshreeves! We merged our updates for ITP 2.1. Feel free to update your branch with the latest refactors 👍

joshreeves:
Hey @francinen! Thanks for the heads up. I updated my branch to the latest.

app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -87,6 +89,13 @@ def render_invalid_shop_error
       redirect_to return_address
     end
 
+    def enable_cookie_access
+      fullpage_redirect_to enable_cookies_path(
+        shop: sanitized_shop_name,
+        return_to: params[:return_to]
```
ragalie: Should this be `session[:return_to]`?

ragalie:
Does this support the ITP 2.1 use case too, or just ITP 2.0?

tylerball:
@ragalie this only covers the ITP2.0 flow I believe, we'll need to update the ITP2.1 flow to pass the `return_to` param around as well. I'm okay with this PR only covering current ITP @joshreeves, I can open an issue and look at implementing it for the rest of the routes later.

app/controllers/shopify_app/sessions_controller.rb
joshreeves: @ragalie - This is correct because at this point in the execution, the user has been redirected before Safari allows the session to be set, so `session[:return_to]` is blank here, even though it was set within `LoginProtection`.

joshreeves:
@ragalie @tylerball - Thanks for looking at this. I haven't tested it against ITP 2.1, only currently released browsers, including Mobile Safari (through the Shopify POS app) and Safari 12.

Let me know if there's anything I need to do to get this merged. We'd like to get our app back onto the upstream branch.

app/controllers/shopify_app/sessions_controller.rb
ragalie: Isn't it set here though? https://github.com/Shopify/shopify_app/pull/664/files#diff-350e7187d4b8bbcd1eaee3ff44fc7fb3R48

I'd rather us handle `params[:return_to]` in a limited number of places and use `session[:return_to]` as the source of truth for this value except where we do the params processing.

joshreeves:
@ragalie - Looks like the fail in Travis is an unrelated intermittent spec fail related to the configuration test for `/nested/` routes not being reset properly.

ragalie:
@joshreeves Agreed, there's a flapping test :(

It looks like there are some style/code complexity violations; maybe you can try to resolve those and then it'll re-run the tests and we'll get lucky.

nwtn:
What’s the state of this now? Are we ok to close? Should this be merged?

andrewlo:
We're seeing this as a major UX issue for POS Channel as well. If the user tries to go to a given page, the ITP page redirects to the root page instead of the page they are trying to get to.

@joshreeves @ragalie @nwtn can this be merged?

joshreeves:
@nwtn - I haven't had the time to go back and reconcile with the updates that were pushed in November. That being said, it's beginning to be problematic for my app that we are on a custom branched version of `shopify_app`, so I am moving this up in my list of priorities. I'll try to take a look at this in the next week.

@andrewlo

joshreeves:
@ragalie @nwtn - I've rebased with `master` and tested that everything works in Safari 12.0.2 and Shopify POS. Ready for final review.

@andrewlo

joshreeves:
@francinen

andrewlo:
@rahilvirani we may want to update Shopify App gem again after this fix goes in

nwtn:
Thanks! We’ll cut a new version with this shortly

PR #668, opened 2018-11-09
Pass shop param to omniauth strategy

tylerball:
In doing some local testing with shopify_app I discovered that the shop param isn't being passed to the oauth path properly, which makes the app unable to authenticate.

### DIFF
--- a/app/controllers/shopify_app/sessions_controller.rb
+++ b/app/controllers/shopify_app/sessions_controller.rb
@@ -61,7 +61,7 @@ module ShopifyApp
 
     def authenticate_in_context
       clear_top_level_oauth_cookie
-      redirect_to "#{main_app.root_path}auth/shopify"
+      redirect_to URI::Generic.build(path: "#{main_app.root_path}auth/shopify", query: {shop: params[:shop]}.to_param).to_s
     end
 
     def authenticate_at_top_level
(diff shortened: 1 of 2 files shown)


ragalie:
What's the situation where this occurs? I think by the time this path is hit we should already have the shop_domain stored in `session['shopify.omniauth_params']` (we put it in there right beforehand in `#authenticate`).

I suspect one of the following is the root cause:
1. The shop name isn't passing the sanitize step (I remember seeing a separate PR about this), and you're skipping that sanitize step in this PR.
2. Cookies aren't actually available, but they should always be available at this point in the process.

tylerball:
@ragalie Yes I need to investigate this more.

tylerball:
Here's the problem. The matcher in `omniauth-shopify-oauth2` [doesn't use `session['shopify.omniauth_params']` at all](https://github.com/Shopify/omniauth-shopify-oauth2/blob/7750af00c6f784f18335329529937e33edf77f50/lib/omniauth/strategies/shopify.rb#L43), in fact I see no reference to that env key in that repo or in the oauth2 middleware. IMHO using the shop query param is simpler since that's what the default is.

However, I found https://github.com/Shopify/shopify_app/pull/460 which introduced this and the need to override the `setup` key in the initializer to pull the shop from the session instead, apparently for CSRF reasons. I think the real fix here is to update the default behaviour in `omniauth-shopify-oauth2` to pull shop from session and make this simpler.

To reproduce the broken flow you can use the generator to create a basic app and point `shopify_app` at `master` and try the auth flow against local shopify.

ragalie:
Did you update the myshopify_domain? `strategy.options[:myshopify_domain] = 'myshopify.io'`

Someone made this list of steps to connect to local shopify: https://docs.google.com/document/d/1DjyvBrHfJCojQk3jVt7GT8rRWo5UyQYTGsfpAk485eI/edit

PR #671, opened 2018-11-16
Update shipit.rubygems.yml

katiedavis:
As per this: https://shopify.slack.com/archives/C21F9TJPK/p1542396290013400 conversation! This should remove policial from shipit, and apparently it shouldn't be there anyway. I am not sure this is actually how to do it? Do you just list the name? I do not know.

### DIFF
--- a/shipit.rubygems.yml
+++ b/shipit.rubygems.yml
@@ -1,3 +1,3 @@
 # using the default shipit config
  hide:
-  -policial
+  - policial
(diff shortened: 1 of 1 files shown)


shipit.rubygems.yml
```diff
@@ -1 +1,3 @@
 # using the default shipit config
+ hide:
+  -policial
```
francinen: Looks right, based on the docs: https://github.com/Shopify/shipit-engine#ci

Noob question: Does it matter if there isn't a space between the `-` and `policial`?

katiedavis: I think it probably needs to be there! I'll add

PR #681, opened 2019-01-11
Update README to clarify tunneling/embedding

### LINKED ISSUE #680: Content Security Policy prevents iframe from being rendered
In the README for starting development [it is suggested](https://github.com/Shopify/shopify_app#becoming-a-shopify-app-developer) to set the Application Callback URL to http://localhost:3000/

However, it seems like this prevents rendering of the app in an iframe after the app has been installed in a store because it violates the Content Security Policy 

screenshot 2019-01-11 12 57 54

Resulting in this being shown
screenshot 2019-01-11 12 58 12

It would be great to be able to do development without any proxy to localhost, and I find it a bit misleading that taking the steps in the README doesn't actually allow me to test the app inside the admin (embedded).

This *can* be circumvented by setting the `forceRedirect` to false, although it's not exactly the same.

tylerball:
Update the wording of the `Tunnelling` section of the README to:
- emphasize the requirements for local tunnelling
- show an example of using ngrok
- update the instructions for setting up app urls to match the field names in the Partners Dashboard

Closes #680

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ Table of Contents
 * [**Description**](#description)
 * [**Quickstart**](#quickstart)
 * [**Becoming a Shopify App Developer**](#becoming-a-shopify-app-developer)
+* [**App Tunneling**](#app-tunneling)
 * [**Installation**](#installation)
   * [Rails Compatibility](#rails-compatibility)
 * [**Generators**](#generators)
@@ -36,7 +37,6 @@ Table of Contents
 * [**Troubleshooting**](#troubleshooting)
  * [Generator shopify_app:install hangs](#generator-shopify_appinstall-hangs)
 * [**Testing an embedded app outside the Shopify admin**](#testing-an-embedded-app-outside-the-shopify-admin)
-* [**App Tunneling**](#app-tunneling)
 * [**Questions or problems?**](#questions-or-problems)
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,26 +58,22 @@ Check out this screencast on how to create and deploy a new Shopify App to Herok
...
-If you don't have a Shopify Partner account yet head over to http://shopify.com/partners to create one, you'll need it before you can start developing apps.
-
-Once you have a Partner account create a new application to get an Api key and other Api credentials. To create a development application set the Application Callback URL to
+App Tunneling
+-------------
 
-```
-http://localhost:3000/
-```
+In order to install your local app on a shop, use the [App Proxy Controller](#app-proxy-controller-generator) or receive Webhooks from Shopify, your app needs to be on a publicly visible URL. Use a tunneling service like [Forward](https://forwardhq.com/), [RequestBin](https://requestb.in/), [ngrok](https://ngrok.com/) etc. These tools allow you to create a secure tunnel from the public Internet to your local machine.
```
nwtn: First sentence is a [garden-path sentence](https://en.wikipedia.org/wiki/Garden-path_sentence) for me — I interpreted it as “In order to install your local app on a shop, [you should] use the App Proxy Controller…”.

Maybe reword as “Your local app needs to be on a publicly visible URL in order to install it on a shop, use the App Proxy Controller, or receive Webhooks.”

README.md
```diff
@@ -58,26 +58,22 @@ Check out this screencast on how to create and deploy a new Shopify App to Herok
...
 
-and the `redirect_uri` to
+For example with ngrok, run this command to set up proxying to Rails' default port:
```
nwtn: can we add a link to ngrok here?

nwtn: maybe also add a note that they need to use the `https` ngrok link, not the http one

README.md
tylerball: Ah yes, updated.

jamiemtdwyer: Yes, requestb.in does not exist anymore.

You can recommend one of these as an alternative:
https://beeceptor.com/
http://mockbin.org/
https://hookbin.com/

nwtn: good catch, i'll do a new PR to fix that

nwtn: i pushed it to master instead of doing a PR

PR #684, opened 2019-02-05
Fix infinite loop when third-party cookies are disabled

### LINKED ISSUE #677: Infinite loop if Chrome blocks third-party cookies
Hello,

I've observed a couple of issues when trying to authenticate to a shopify_app if Chrome has third-party cookies blocked. It has been a bit more common than I expected.

On 8.3.1, `SessionsController#redirect_for_cookie_access?` never returns false, so the user keeps getting redirect to `/ `-> `/login` -> `/enable_cookies`   infinitely. 

On 8.4.0, upon running `storage_access.js` it throws an error because it can't access `window.sessionStorage`, on line 36.

Steps to recreate: 

1. Visit  chrome://settings/content/cookies  and choose the option to block third-party cookies.
2. Try to open an embedded app running 8.3.1 and 8.4.0

I'm still running 8.3.1 in production because I've also seen some infinite redirections in 8.4.0, though I didn't have a chance to look at the cause yet. 

A possible solution for these issues would be:

1. Add a counter query param to check the number of attempts at setting a cookie.
2. Handle exceptions in the ITP Javascript side. 

I'm not too familiar with the current ITP workaround implementation, so I can't even say if these ideas make much sense. Any input is appreciated.

nwtn:
Fixes #677.

I’m not that familiar with our ITP solution, so I’m looking for feedback on the direction here.

We already have a check on the number of install redirects; this try/catch block just makes sure that this is the path we go down if users have explicitly turned off third-party cookies.

chrome-storage-demo

### DIFF
--- a/app/assets/javascripts/shopify_app/storage_access.js
+++ b/app/assets/javascripts/shopify_app/storage_access.js
@@ -33,9 +33,14 @@
   }
 
   StorageAccessHelper.prototype.grantedStorageAccess = function() {
-    sessionStorage.setItem('shopify.granted_storage_access', true);
-    document.cookie = 'shopify.granted_storage_access=true';
-    this.redirectToAppHome();
+    try {
+      sessionStorage.setItem('shopify.granted_storage_access', true);
+      document.cookie = 'shopify.granted_storage_access=true';
+      this.redirectToAppHome();
+    } catch (error) {
+      console.warn('Third party cookies may be blocked.', error);
+      this.redirectToAppTLD(ACCESS_DENIED_STATUS);
+    }
   }
 
   StorageAccessHelper.prototype.handleRequestStorageAccess = function() {
--- a/shopify_app.gemspec
+++ b/shopify_app.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency('rake')
   s.add_development_dependency('byebug')
-  s.add_development_dependency('sqlite3')
+  s.add_development_dependency('sqlite3', '~> 1.3.6')
   s.add_development_dependency('minitest')
   s.add_development_dependency('mocha')
 


nwtn:
@flavio-b @sergey-alekseev let me know if this works for you. If so, I can cut a new release tomorrow.

nwtn:
Unfortunately it looks like this breaks our ITP flow so I’ll need to keep digging.

broken-safari

nwtn:
Crisis averted: the broken ITP flow was because I was testing in the wrong version of Safari Technology Preview. All is good. I’ll merge this shortly and cut a new version.

sergey-alekseev:
@nwtn thanks for the quick fix! Sorry for not checking it yesterday, I'm not available for a couple of days, but I've passed this to our dev team and will get back to you after they confirm everything is alright.

flavio-b:
I did a quick test and this works great. Thanks for working on this!

dhalai:
@nwtn Thank you for your fix. I tested it and it works exactly as you described 👍

sergey-alekseev:
@nwtn thanks for https://github.com/Shopify/shopify_app/pull/685 as well! Hope to see 8.4.1 on https://rubygems.org/gems/shopify_app soon. As @dhalai confirmed the solution works.

nwtn:
8.4.1 should now be deployed on rubygems! please feel free to create a new issue if you experience any problems.

SockFuzz:
I am still getting an error when trying to access certain Shopify applications that reads "The application can't be loaded, check that your browser allows third-party cookies". I am trying to access "Recurring Orders & Subscription Box App by Bold".

My Chrome settings are already configured to allow 3rd party cookies and I have gone ahead and whitelisted all shopify and myshopify URLs with wildcards for subdomains and paths.

It keeps cycling between "https://ro.boldapps.net/s/life-unpacked/payment_settings" and "https://my-store.myshopify.com/admin/apps/5a3cb5ac82a21b14f0e3fa026a88eb29/s/life-unpacked/payment_settings".

I am up to date using Chrome Version 72.0.3626.121 (Official Build) (64-bit) on Mac.

flavio-b:
@SockFuzz , I'd suggest reaching out to Bold for support. Also, try closing the browser and opening the app again from the Shopify Admin. Clear out all Session Storage and cookies too.

yu-coder:
Hi, I am getting the same issue.
I am using shopify_app gem version v12.0.2

SockFuzz:
> @SockFuzz , I'd suggest reaching out to Bold for support. Also, try closing the browser and opening the app again from the Shopify Admin. Clear out all Session Storage and cookies too.

This issue wound up being with my Chrome browser and local profile. There is something messed up with how sessions and my Chrome profile for some reason. Every other browser works and private Chrome tabs work fine.

> Hi, I am getting the same issue.
> I am using shopify_app gem version v12.0.2

Maybe try using a different browser to see if the issue persists. Another quick test would be to use private tab or incognito mode as these bypass any local profile settings and extensions that might be configured.

PR #694, opened 2019-02-22
Abstract support to secrets rotation

jmignac:
Relates to https://github.com/Shopify/buy-button/issues/2739

- Extend the hmac validation to support both previous and new token when a secret rotation is happening. 
- Add a job generator, to make requests to all `shops` to update their `shopify_token` with the use of a `refresh_token`. This should work for any consumer that has `shopify_token` and `shopify_domain` in their `shops` table.

PR by @jamiemtdwyer and @jmignac

### DIFF
--- a/lib/generators/shopify_app/add_after_authenticate_job/add_after_authenticate_job_generator.rb
+++ b/lib/generators/shopify_app/add_after_authenticate_job/add_after_authenticate_job_generator.rb
@@ -12,7 +12,9 @@ module ShopifyApp
       def init_after_authenticate_config
         initializer = load_initializer
 
-       after_authenticate_job_config = "  config.after_authenticate_job = { job: Shopify::AfterAuthenticateJob, inline: false }\n"
+        after_authenticate_job_config =
+          "  config.after_authenticate_job = "\
+          "{ job: Shopify::AfterAuthenticateJob, inline: false }\n"
 
         inject_into_file(
           'config/initializers/shopify_app.rb',
--- a/lib/generators/shopify_app/rotate_shopify_token_job/rotate_shopify_token_job_generator.rb
+++ b/lib/generators/shopify_app/rotate_shopify_token_job/rotate_shopify_token_job_generator.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'rails/generators/base'
+
+module ShopifyApp
+  module Generators
+    class RotateShopifyTokenJobGenerator < Rails::Generators::Base
+      source_root File.expand_path('../templates', __FILE__)
+
+      def add_rotate_shopify_token_job
+        copy_file('rotate_shopify_token_job.rb', "app/jobs/shopify/rotate_shopify_token_job.rb")
+        copy_file('rotate_shopify_token.rake', "lib/tasks/shopify/rotate_shopify_token.rake")
+      end
+    end
+  end
+end
--- a/lib/generators/shopify_app/rotate_shopify_token_job/templates/rotate_shopify_token.rake
+++ b/lib/generators/shopify_app/rotate_shopify_token_job/templates/rotate_shopify_token.rake
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+namespace :shopify do
+  desc "Rotate shopify tokens for all active shops"
+  task :rotate_shopify_tokens, [:refresh_token] => :environment do |_t, args|
+    all_active_shops.find_each do |shop|
+      Shopify::RotateShopifyTokenJob.perform_later(
+        shop_domain: shop.shopify_domain,
+        refresh_token: args[:refresh_token]
+      )
+    end
+  end
+
+  # Its implementation will depend on the app configuration. Change accordingly.
+  def all_active_shops
+    Shop.all
+  end
+end
--- a/lib/generators/shopify_app/rotate_shopify_token_job/templates/rotate_shopify_token_job.rb
+++ b/lib/generators/shopify_app/rotate_shopify_token_job/templates/rotate_shopify_token_job.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module Shopify
+  class RotateShopifyTokenJob < ActiveJob::Base
+    def perform(params)
+      @shop = Shop.find_by(shopify_domain: params[:shop_domain])
+      return unless @shop
+
+      config = ShopifyApp.configuration
+      uri = URI("https://#{@shop.shopify_domain}/admin/oauth/access_token")
+      post_data = {
+        client_id: config.api_key,
+        client_secret: config.secret,
+        refresh_token: params[:refresh_token],
+        access_token: @shop.shopify_token,
+      }
+
+      @response = Net::HTTP.post_form(uri, post_data)
+      return log_error(response_expcetion_error_message) unless @response.is_a?(Net::HTTPSuccess)
+
+      access_token = JSON.parse(@response.body)['access_token']
+      return log_error(no_access_token_error_message) unless access_token
+
+      @shop.update(shopify_token: access_token)
+    end
+
+    private
+
+    def log_error(message)
+      Rails.logger.error(message)
+    end
+
+    def no_access_token_error_message
+      "RotateShopifyTokenJob response returned no access token for shop: #{@shop.shopify_domain}"
+    end
+
+    def response_exception_error_message
+      "RotateShopifyTokenJob failed for shop: #{@shop.shopify_domain}." \
+        "Response returned status: #{@response.code}. Error message: #{@response.message}. "
+    end
+  end
+end
--- a/lib/shopify_app/controller_concerns/webhook_verification.rb
+++ b/lib/shopify_app/controller_concerns/webhook_verification.rb
@@ -15,12 +15,16 @@ module ShopifyApp
     end
 
     def hmac_valid?(data)
-      secret = ShopifyApp.configuration.secret
-      digest = OpenSSL::Digest.new('sha256')
-      ActiveSupport::SecurityUtils.secure_compare(
-        shopify_hmac,
-        Base64.encode64(OpenSSL::HMAC.digest(digest, secret, data)).strip
-      )
+      secrets = [ShopifyApp.configuration.secret, ShopifyApp.configuration.old_secret].reject(&:blank?)
+
+      secrets.any? do |secret|
+        digest = OpenSSL::Digest.new('sha256')
+
+        ActiveSupport::SecurityUtils.secure_compare(
+          shopify_hmac,
+          Base64.strict_encode64(OpenSSL::HMAC.digest(digest, secret, data))
+        )
+      end
     end
 
     def shop_domain
--- a/shopify_app.gemspec
+++ b/shopify_app.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('browser_sniffer', '~> 1.1.0')
   s.add_runtime_dependency('rails', '>= 5.0.0')
   s.add_runtime_dependency('shopify_api', '>= 4.3.5')
-  s.add_runtime_dependency('omniauth-shopify-oauth2', '~> 2.0.0')
+  s.add_runtime_dependency('omniauth-shopify-oauth2', '~> 2.1.0')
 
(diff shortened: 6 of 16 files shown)


lib/shopify_app/controller_concerns/webhook_verification.rb
```diff
@@ -15,12 +15,18 @@ def verify_request
     end
 
     def hmac_valid?(data)
-      secret = ShopifyApp.configuration.secret
-      digest = OpenSSL::Digest.new('sha256')
-      ActiveSupport::SecurityUtils.secure_compare(
-        shopify_hmac,
-        Base64.encode64(OpenSSL::HMAC.digest(digest, secret, data)).strip
-      )
+      secrets = [ShopifyApp.configuration.secret, ShopifyApp.configuration.old_secret].map(&:to_s)
```
kmcphillips: I don't think you want `to_s` here. Are you guarding against `nil`? Because if so you basically want:

```ruby
secrets = [ShopifyApp.configuration.secret, ShopifyApp.configuration.old_secret].reject(&:blank)
```

kmcphillips:
🙏 Thank you for working on this.

lib/shopify_app/controller_concerns/webhook_verification.rb
jmignac: That was exactly why. Thanks!

ragalie:
:wave: I think we ran into a similar problem a few months ago with the OAuth flow during secret rotation (https://github.com/Shopify/omniauth-shopify-oauth2/pull/70).

Unfortunately we haven't had a chance to roll it out yet, but it would be nice to make it a companion to your work. @theundeadmonk do you think this is something we could finish up and get out alongside this?

jmignac:
Once https://github.com/Shopify/omniauth-shopify-oauth2/pull/70/files#diff-20fb14b92659da69e72effc9ab8f72fe gets merged, we need to add a named parameter `old_client_secret: ShopifyApp.configuration.old_secret` to [lib/generators/shopify_app/install/templates/shopify_provider.rb:4](https://github.com/Shopify/shopify_app/blob/master/lib/generators/shopify_app/install/templates/shopify_provider.rb#L4)

lib/generators/shopify_app/rotate_shopify_token_job/rotate_shopify_token_job_generator.rb
```diff
@@ -0,0 +1,16 @@
...
+
+require 'rails/generators/base'
+
+module ShopifyApp
+  module Generators
+    class RotateShopifyTokenJobGenerator < Rails::Generators::Base
+      source_root File.expand_path('../templates', __FILE__)
+
+      def add_rotate_shopify_token_job
+        copy_file('rotate_shopify_token_job.rb', "app/jobs/shopify/rotate_shopify_token_job.rb")
```
jmignac: I wonder if they should they be `app/jobs/shopify_app/...`, but the way they are now is the pattern though.

jamiemtdwyer: I think the `shopify` namespace makes enough sense, since the job deals with Shopify concerns 👍

test/shopify_app/controller_concerns/webhook_verification_test.rb
```diff
@@ -27,7 +38,21 @@ class WebhookVerificationTest < ActionController::TestCase
       params = { foo: 'anything' }
       digest = OpenSSL::Digest.new('sha256')
       secret = ShopifyApp.configuration.secret
-      valid_hmac = Base64.encode64(OpenSSL::HMAC.digest(digest, secret, params.to_query)).strip
+      valid_hmac = Base64.strict_encode64(OpenSSL::HMAC.digest(digest, secret, params.to_query))
```
rebeccajfriedman: Just a suggestion, but I feel like this should be a helper method since you use it in the code and in multiple tests.

lib/generators/shopify_app/rotate_shopify_token_job/templates/rotate_shopify_token.rake
```diff
@@ -0,0 +1,17 @@
...
+      Shopify::RotateShopifyTokenJob.new.perform(
+        shop_domain: shop.shopify_domain,
+        refresh_token: args[:refresh_token]
+      )
+    end
+  end
+
+  # Its implementation will depend on the app configuration. Change accordingly.
+  def all_active_shops
+    Shop.all
```
rebeccajfriedman: This is a bit dangerous. Even though this is only a suggestion, I would recommend using batching (maybe https://apidock.com/rails/ActiveRecord/Batches/find_in_batches).

lib/generators/shopify_app/rotate_shopify_token_job/templates/rotate_shopify_token.rake
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+namespace :shopify do
+  desc "Rotate shopify tokens for all active shops"
+  task :rotate_shopify_tokens, [:refresh_token] => :environment do |_t, args|
+    all_active_shops.each do |shop|
+      Shopify::RotateShopifyTokenJob.new.perform(
```
rebeccajfriedman: Just to confirm: Is this being run asynchronously, or in the loop?

lib/generators/shopify_app/rotate_shopify_token_job/templates/rotate_shopify_token.rake
```diff
@@ -0,0 +1,17 @@
...
+        shop_domain: shop.shopify_domain,
```
rebeccajfriedman: I think this makes the assumption that apps are calling the primary domain `shopify_domain` locally. Some apps might be using `shop_domain` or `domain` or `primary_domain`.

lib/generators/shopify_app/rotate_shopify_token_job/rotate_shopify_token_job_generator.rb
rebeccajfriedman: I think it's okay to maintain the existing pattern.

lib/generators/shopify_app/rotate_shopify_token_job/templates/rotate_shopify_token.rake
jamiemtdwyer: `shopify_domain` is the column that the Shop model generator in `shopify_app` provides by default.

The rake task is only meant to be an example implementation, and isn't intended to be able to be used out of the box by all apps. The job is what we are really aiming to make re-usable here.

lib/generators/shopify_app/rotate_shopify_token_job/templates/rotate_shopify_token.rake
jamiemtdwyer: Good catch – this should be async

lib/shopify_app/controller_concerns/webhook_verification.rb
```diff
@@ -15,12 +15,16 @@ def verify_request
...
-        Base64.encode64(OpenSSL::HMAC.digest(digest, secret, data)).strip
-      )
+      secrets = [ShopifyApp.configuration.secret, ShopifyApp.configuration.old_secret].reject(&:blank?)
+
+      secrets.any? do |secret|
+        digest = OpenSSL::Digest.new('sha256')
+
+        ActiveSupport::SecurityUtils.secure_compare(
+          shopify_hmac,
+          Base64.strict_encode64(OpenSSL::HMAC.digest(digest, secret, data))
```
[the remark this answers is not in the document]
jamiemtdwyer: When you use `encode64`, a CRLF is added after every 60 encoded characters. Using `strict_encode64` avoids this 👍

lib/shopify_app/controller_concerns/webhook_verification.rb
```diff
@@ -15,12 +15,16 @@ def verify_request
...
     def hmac_valid?(data)
-      secret = ShopifyApp.configuration.secret
-      digest = OpenSSL::Digest.new('sha256')
-      ActiveSupport::SecurityUtils.secure_compare(
-        shopify_hmac,
-        Base64.encode64(OpenSSL::HMAC.digest(digest, secret, data)).strip
-      )
+      secrets = [ShopifyApp.configuration.secret, ShopifyApp.configuration.old_secret].reject(&:blank?)
+
+      secrets.any? do |secret|
```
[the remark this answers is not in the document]
jamiemtdwyer: The old secret will be deleted from the partner dashboard after all of the access tokens are refreshed, so an hour at most (the refresh token is only valid for an hour).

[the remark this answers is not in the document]
jamiemtdwyer: Actually, I see what you mean – in the instructions we provide, we'll have to tell folks to remove the `old_secret` value after rotation has been completed.

test/shopify_app/controller_concerns/webhook_verification_test.rb
jmignac: I'm actually going to change the tests to assert the right result, instead of using code similar to the method implementation. 👍

lib/generators/shopify_app/rotate_shopify_token_job/templates/rotate_shopify_token.rake
jmignac: Fixed it in line 5, `all_active_shops.find_each`

lib/generators/shopify_app/rotate_shopify_token_job/templates/rotate_shopify_token_job.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    private
+
+    def log_error(message)
+      Rails.logger.error(message)
+    end
+
+    def no_access_token_error_message
+      "RotateShopifyTokenJob response returned no access token for shop: #{shop.shopify_domain}"
```
rebeccajfriedman: Wouldn't this need to be `@shop`?

shopify_app.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('browser_sniffer', '~> 1.1.0')
   s.add_runtime_dependency('rails', '>= 5.0.0')
   s.add_runtime_dependency('shopify_api', '>= 4.3.5')
-  s.add_runtime_dependency('omniauth-shopify-oauth2', '~> 2.0.0')
+  s.add_runtime_dependency('omniauth-shopify-oauth2', '~> 2.1.0')
```
kmcphillips: 👍 https://github.com/Shopify/omniauth-shopify-oauth2/commit/033bc7f81443496320066367e52359373ccc9683

lib/generators/shopify_app/add_after_authenticate_job/add_after_authenticate_job_generator.rb
```diff
@@ -12,7 +12,9 @@ class AddAfterAuthenticateJobGenerator < Rails::Generators::Base
       def init_after_authenticate_config
         initializer = load_initializer
 
-       after_authenticate_job_config = "  config.after_authenticate_job = { job: Shopify::AfterAuthenticateJob, inline: false }\n"
```
kmcphillips: Is this just a line length linting change?

lib/generators/shopify_app/rotate_shopify_token_job/templates/rotate_shopify_token.rake
michaelsunglee: Could be wrong, but I think you want just `Shop` here. 
I think if you call Shop.all.find_each, you're still loading `all` the shops

sambostock: `Shop.all` returns a relation, which is lazy, so this is fine.

lib/generators/shopify_app/add_after_authenticate_job/add_after_authenticate_job_generator.rb
jmignac: Yup hehehe. Lint wasn't activated when this gem was last touched, for some reason. It started complaining about lots of things here. It was event using some general cops, we had to add Shopify's

PR #695, opened 2019-02-25
Moving the inline js to a js asset file

Abdulwahaab710:
## What is the problem you are trying to solve?

[Content Security Policy(CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) is an additional layer of security that helps detecting and migrating from Cross-site Scripting (XSS). When generate a new Shopify app through the gem, it's generate few files and one of the files is `app/views/layouts/embedded_app.html.erb` and `app/views/layouts/_flash_messages.html.erb`. This file `lib/generators/shopify_app/install/install_generator.rb` copies the following files to your app.
- `lib/generators/shopify_app/install/templates/embedded_app.html.erb`
- `lib/generators/shopify_app/install/templates/_flash_messages.html.erb` 

The `embedded_app.html.erb` and `_flash_messages.html.erb` contains an inline script tag and it's doesn't play well with CSP, where the first option is to replace the `<script>` tag with a `<%= nonced_javascript_tag %>` if you are using [secure_headers](https://github.com/twitter/secure_headers) or `<%= javascript_tag nonce: true %>` if you are using Rails 5.2 CSP. The second option is to allow `'unsafe-inline'` which is definitely not recommend because it defeats the purpose of CSP detecting and migrating from XSS.

## What are you trying to accomplish with this PR?
Move inline javascript into a `.js` asset and pass the ERB variables through a data attribute in a div so it plays nicely with CSP.

## How to 🎩 this PR?

The :tophat:ing for this PR is divide into 3 parts. Part 1 will be making sure the shopify app loads and doesn't throw any errors. Part 2 will be checking the files that the shopify app generates and finally part 3 will be making sure that flashes works.

### Setting up Shopify_app

1. First clone the repo

2. Create a new Ruby on Rails project
  ```bash
  rails new test-project
  ```

3. Add the shopify_app gem into your Gemfile and make it point to the local shopify_app
  ```ruby
  gem 'shopify_app', path: 'THE GEM LOCAL PATH'
  ```

4. Install the gem by running:
  ```bash
  bundle install
  ```

5. Generate a shopify_app using rails generator
  ```
  rails generate shopify_app --api_key <YOUR API KEY>  --secret <YOUR SECRET>
  ```

6. Run the application
  ```
  rails s
  ```

### Part 1 (Shopify app init)

1. Install the application on your development store

2. Go to the main page and check that there is no errors in the console. You should see warnings but no errors. Check the following :point_down:  screenshot for reference.

test0x04___discountify___shopify

### Part 2

1. open the file `app/views/layouts/embedded_app.html.erb` and make sure that the file doesn't have the following script tag.

https://github.com/Shopify/shopify_app/blob/b7fd181c73378b3956dbce95dab90b06e12b9244/lib/generators/shopify_app/install/templates/embedded_app.html.erb#L25-L32

But it has a `content_tag(:div....)`

https://github.com/Shopify/shopify_app/blob/d19a02835c95917c625072b0007d34d4c05059c2/lib/generators/shopify_app/install/templates/embedded_app.html.erb#L25-L29

2. make sure that the gem generate a JavaScript file in the assets folder `app/assets/javascripts/shopify_app.js` and it matches the following file.

https://github.com/Shopify/shopify_app/blob/d19a02835c95917c625072b0007d34d4c05059c2/lib/generators/shopify_app/install/templates/shopify_app.js#L1-L9

### Part 3 (Flashes)

To make sure that application displays the flash messages, we have to pass them to the view. The easiest way would be to pass the flashes through a get parameter and than take that parameter and pass it to flash

#### Setup

In `app/controllers/home_controllers.rb` modify the index action to be like the following:

```ruby
  def index
    @products = ShopifyAPI::Product.find(:all, params: { limit: 10 })
    @webhooks = ShopifyAPI::Webhook.find(:all)
    flash[:error] = params[:error] if params[:error]
    flash[:notice] = params[:notice] if params[:notice]
  end
```

#### Testing

To view the flashes goto https://<YOURR DEVELOPMENT STORE>.myshopify.com/admin/apps/<YOUR APPLICATION>/?error=This+is+an+error. You should see an error flash message with `This is an error` and in the console you should see `ShopifyApp client sent {"message":"Shopify.API.flash.error","data":{"message":"This is an error"}} to https://test0x04.myshopify.com`

test0x04___discountify___shopify

To view the flashes goto https://<YOURR DEVELOPMENT STORE>.myshopify.com/admin/apps/<YOUR APPLICATION>/?notice=This is a notice. You should see an error flash message with `This is a notice` and in the console you should see `ShopifyApp client sent {"message":"Shopify.API.flash.notice","data":{"message":"This is a notice"}} to https://test0x04.myshopify.com`

test0x04___discountify___shopify

### DIFF
--- a/example/app/assets/javascripts/flash_messages.js
+++ b/example/app/assets/javascripts/flash_messages.js
@@ -1,0 +1,17 @@
+var eventName = typeof(Turbolinks) !== 'undefined' ? 'turbolinks:load' : 'DOMContentLoaded';
+
+if (!document.documentElement.hasAttribute("data-turbolinks-preview")) {
+  document.addEventListener(eventName, function flash() {
+    var flash = JSON.parse(document.getElementById('flash').dataset.flash);
+
+    if (flash.notice) {
+      ShopifyApp.flashNotice(flash.notice);
+    }
+
+    if (flash.error) {
+      ShopifyApp.flashError(flash.error);
+    }
+
+    document.removeEventListener(eventName, flash)
+  });
+}
--- a/example/app/assets/javascripts/flash_messages.js
+++ b/example/app/assets/javascripts/flash_messages.js
@@ -2,7 +2,7 @@ var eventName = typeof(Turbolinks) !== 'undefined' ? 'turbolinks:load' : 'DOMCon
 
 if (!document.documentElement.hasAttribute("data-turbolinks-preview")) {
   document.addEventListener(eventName, function flash() {
-    var flash = JSON.parse(document.getElementById('flash').dataset.flash);
+    var flash = JSON.parse(document.getElementById('shopify-app-flash').dataset.flash);
 
     if (flash.notice) {
       ShopifyApp.flashNotice(flash.notice);
--- a/example/app/views/layouts/embedded_app.html.erb
+++ b/example/app/views/layouts/embedded_app.html.erb
@@ -28,6 +28,11 @@
         forceRedirect: true
       });
     </script>
+    <%= content_tag(:div, nil, id: 'shopify-app-init', data: {
+      api_key: ShopifyApp.configuration.api_key,
+      shop_origin: ("https://#{ @shop_session.url }" if @shop_session),
+      debug: (Rails.env.development? ? 'true' : 'false')
+    } ) %>
 
     <%= javascript_include_tag 'application', 'data-turbolinks-track' => true %>
 
--- a/example/app/views/layouts/embedded_app.html.erb
+++ b/example/app/views/layouts/embedded_app.html.erb
@@ -20,18 +20,10 @@
 
     <script src="https://cdn.shopify.com/s/assets/external/app.js?<%= Time.now.strftime('%Y%m%d%H') %>"></script>
 
-    <script type="text/javascript">
-      ShopifyApp.init({
-        apiKey: "<%= ShopifyApp.configuration.api_key %>",
-        shopOrigin: "<%= "https://#{ @shop_session.domain }" if @shop_session %>",
-        debug: false,
-        forceRedirect: true
-      });
-    </script>
     <%= content_tag(:div, nil, id: 'shopify-app-init', data: {
       api_key: ShopifyApp.configuration.api_key,
       shop_origin: ("https://#{ @shop_session.url }" if @shop_session),
-      debug: (Rails.env.development? ? 'true' : 'false')
+      debug: Rails.env.development?
     } ) %>
 
     <%= javascript_include_tag 'application', 'data-turbolinks-track' => true %>
--- a/lib/generators/shopify_app/install/templates/_flash_messages.html.erb
+++ b/lib/generators/shopify_app/install/templates/_flash_messages.html.erb
@@ -1,3 +1,3 @@
 <% content_for :javascript do %>
-  <%= content_tag(:div, nil, id: 'flash', data: { flash: { notice: flash[:notice], error: flash[:error] } }) %>
+  <%= content_tag(:div, nil, id: 'shopify-app-flash', data: { flash: { notice: flash[:notice], error: flash[:error] } }) %>
 <% end %>
--- a/lib/generators/shopify_app/install/templates/_flash_messages.html.erb
+++ b/lib/generators/shopify_app/install/templates/_flash_messages.html.erb
@@ -1,19 +1,3 @@
 <% content_for :javascript do %>
-<script type="text/javascript">
-  var eventName = typeof(Turbolinks) !== 'undefined' ? 'turbolinks:load' : 'DOMContentLoaded';
-
-  if (!document.documentElement.hasAttribute("data-turbolinks-preview")) {
-    document.addEventListener(eventName, function flash() {
-      <% if flash[:notice] %>
-        ShopifyApp.flashNotice(<%== flash[:notice].to_json %>);
-      <% end %>
-
-      <% if flash[:error] %>
-        ShopifyApp.flashError(<%== flash[:error].to_json %>);
-      <% end %>
-                             
-      document.removeEventListener(eventName, flash)
-    });
-  }
-</script>
+  <%= content_tag(:div, nil, id: 'flash', data: { flash: { notice: flash[:notice], error: flash[:error] } }) %>
 <% end %>
--- a/lib/generators/shopify_app/install/templates/embedded_app.html.erb
+++ b/lib/generators/shopify_app/install/templates/embedded_app.html.erb
@@ -30,6 +30,11 @@
         forceRedirect: true
       });
     </script>
+    <%= content_tag(:div, nil, id: 'shopify-app-init', data: {
+      api_key: ShopifyApp.configuration.api_key,
+      shop_origin: ("https://#{ @shop_session.url }" if @shop_session),
+      debug: (Rails.env.development? ? 'true' : 'false')
+    } ) %>
 
     <% if content_for?(:javascript) %>
       <div id="ContentForJavascript" data-turbolinks-temporary>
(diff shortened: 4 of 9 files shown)


example/app/views/layouts/embedded_app.html.erb
```diff
@@ -20,14 +20,11 @@
...
-        apiKey: "<%= ShopifyApp.configuration.api_key %>",
-        shopOrigin: "<%= "https://#{ @shop_session.url }" if @shop_session %>",
-        debug: false,
-        forceRedirect: true
-      });
-    </script>
+    <%= content_tag(:div, nil, id: 'shopify-app-init', data: {
+      api_key: ShopifyApp.configuration.api_key,
+      shop_origin: ("https://#{ @shop_session.url }" if @shop_session),
+      debug: (Rails.env.development? ? 'true' : 'false')
```
EiNSTeiN-: I don't think this should be a string otherwise it will always evaluate to true:
```suggestion
      debug: !!Rails.env.development?
```

lib/generators/shopify_app/install/templates/embedded_app.html.erb
```diff
@@ -22,14 +22,12 @@
...
-        apiKey: "<%= ShopifyApp.configuration.api_key %>",
-        shopOrigin: "<%= "https://#{ @shop_session.url }" if @shop_session %>",
-        debug: <%= Rails.env.development? ? 'true' : 'false' %>,
-        forceRedirect: true
-      });
-    </script>
+    <%= content_tag(:div, nil, id: 'shopify-app-init', data: {
+      api_key: ShopifyApp.configuration.api_key,
+      shop_origin: ("https://#{ @shop_session.url }" if @shop_session),
+      debug: (Rails.env.development? ? 'true' : 'false')
```
EiNSTeiN-:
```suggestion
      debug: !!Rails.env.development?
```

example/app/views/layouts/embedded_app.html.erb
clayton-shopify: I expect it's already boolean, so is the `!!` even needed?

example/app/assets/javascripts/flash_messages.js
```diff
@@ -0,0 +1,17 @@
+var eventName = typeof(Turbolinks) !== 'undefined' ? 'turbolinks:load' : 'DOMContentLoaded';
+
+if (!document.documentElement.hasAttribute("data-turbolinks-preview")) {
+  document.addEventListener(eventName, function flash() {
+    var flash = JSON.parse(document.getElementById('flash').dataset.flash);
+
+    if (flash.notice) {
+      ShopifyApp.flashNotice(flash.notice);
```
nwtn: Not a dealbreaker on merging this PR, but my preference is that (if we’re editing this stuff anyway) we stop using EASDK and start using App Bridge. cc @gf3

clayton-shopify:
While we're getting rid of inline scripts, there's another one here:

https://github.com/Shopify/shopify_app/blob/e4dc47d2105adb1df65966afeac09d4518d0a5d1/lib/generators/shopify_app/home_controller/templates/shopify_app_ready_script.html.erb#L2-L6

lib/generators/shopify_app/install/templates/_flash_messages.html.erb
```diff
@@ -1,19 +1,3 @@
...
-
-      <% if flash[:error] %>
-        ShopifyApp.flashError(<%== flash[:error].to_json %>);
-      <% end %>
-                             
-      document.removeEventListener(eventName, flash)
-    });
-  }
-</script>
+  <%= content_tag(:div, nil, id: 'flash', data: { flash: { notice: flash[:notice], error: flash[:error] } }) %>
```
clayton-shopify: Maybe use a more unique id for this to avoid conflicts with the rest of the app? How about `shopify-app-flash`?

Abdulwahaab710:
@clayton-shopify For the following inline JS I was thinking maybe we can move it to a separate file and include a helper method that sets the title by returning a `div` tag with a title in a data attribute.
https://github.com/Shopify/shopify_app/blob/e4dc47d2105adb1df65966afeac09d4518d0a5d1/lib/generators/shopify_app/home_controller/templates/shopify_app_ready_script.html.erb#L2-L6

The helper method can be something like the following:

```ruby
module ShopifyAppHelper
  def shopify_app_bar_title(title = '')
    content_tag(:div, nil, id: 'shopify-app-bar', data: { title: title }, style: 'display: none;')
  end
end
```

And after moving the inline js into a file and the JavaScript will look like the following:

```js
document.addEventListener("DOMContentLoaded", () => {
  title = document.getElementById('shopify-app-bar').dataset.title;
  if (title) {
    ShopifyApp.ready(() => {
      ShopifyApp.Bar.initialize({ title: title });
    });
  }
});
```
The helper method can be called in any view, see the following example:

```rb
<%= shopify_app_bar_title('Home') %>
```

The reason why I am suggesting creating a helper method is that when we will move the file, we will have to pass the title of the page through a div data attribute. In the JavaScript file we will have to check if the div exists and if it does then we can set the bar title. Now if a developer wants to set the bar title they can set the title, one by adding a script tag and set the title like in this [file](https://github.com/Shopify/shopify_app/blob/e4dc47d2105adb1df65966afeac09d4518d0a5d1/lib/generators/shopify_app/home_controller/templates/shopify_app_ready_script.html.erb#L2-L6). The second way would be adding a `div` tag and setting the title in the data attribute.
This doesn't look like a nice syntax IMO.

Any thoughts?
cc: @nwtn & @g-Off

nwtn:
@gf3 Can you or someone from your team weigh in on the above proposal? I don’t really want to put extra effort into solving this title thing if it’s only going to work with EASDK and not App Bridge

gf3:
the solution presented **will continue to work** with App Bridge via the legacy bridge as-is

that being said, this style of server-side javascript rendering relies on window-level globals and is not a pattern we continue to recommend. it's possible to define `AppBridge` on the `window` to make this work in the future, however this continues to perpetuate the patterns of the EASDK which we're trying to move away from

nwtn:
@gf3 sorry, i wasn’t clear. what i meant was: what are our options to get rid of the inline JS but still use app bridge (not legacy bridge)? even if it’s not our preferred pattern, is the best bet to use the CDN-hosted script and go with a pattern similar to what @Abdulwahaab710 proposed?

gf3:
@nwtn i am not intimately familiar with how the shopify_app gem is able to pass assets to the consuming rails application. however, it's recommended to use a modern build system with app bridge (i.e. pulling the package from npm/yarn and integrating it into your app build). if that's not possible, the CDN path is supported and will not be removed (and is intended for scenarios where a build system is unavailable)

nwtn:
Ok, I’d definitely say updating build tooling for this repo is out of scope, so my suggestion is that if we’re editing any of that EASDK JavaScript, it should be updated by:

* using the App Bridge client script instead of the EASDK one (e.g. `<script src="https://unpkg.com/@shopify/app-bridge"></script>`, see https://help.shopify.com/en/api/embedded-apps/app-bridge/getting-started)
* using App Bridge actions instead of EASDK methods, e.g.:

````
var app = createApp({
  apiKey: '12345',
  shopOrigin: shopOrigin,
});

var TitleBar = actions.TitleBar;
var titleBarOptions = {
  title: 'My page title',
};
var myTitleBar = TitleBar.create(app, titleBarOptions);
```

If making these changes is out of scope, then I’d prefer if we just leave the existing EASDK stuff alone.

gf3:
@nwtn one recommendation is if using the `unpkg` CDN i would pin it to a specific version instead of grabbing the latest. e.g.

```html
<script src="https://unpkg.com/@shopify/app-bridge@1.0.7"></script>
```

Abdulwahaab710:
@nwtn I will open another PR that replaces EASDK with App Bridge, since the purpose of this PR is to move all inline JavaScript into JavaScript asset files.

nwtn:
@Abdulwahaab710 is this ok to merge now?

Abdulwahaab710:
@nwtn Yes it is okay to merge this PR :)

PR #697, opened 2019-02-27
Fix infinite loop when Edge has third-party cookies blocked.

flavio-b:
@nwtn, @sergey-alekseev

Another fix on top of #684. I only now had the chance to test how 8.4.2 works on MS Edge (v44), when third-party cookies are blocked, and it still results in an infinite loop. 

Edge does't throw an error when we try to set session storage or cookies, even if third-party cookies are blocked. My solution is to manually check and throw the error if cookies are empty, after we just attempted to set one.

### DIFF
--- a/app/assets/javascripts/shopify_app/storage_access.js
+++ b/app/assets/javascripts/shopify_app/storage_access.js
@@ -36,6 +36,9 @@
     try {
       sessionStorage.setItem('shopify.granted_storage_access', true);
       document.cookie = 'shopify.granted_storage_access=true';
+      if (!document.cookie) {
+        throw 'Cannot set third-party cookie.'
+      }
       this.redirectToAppHome();
     } catch (error) {
       console.warn('Third party cookies may be blocked.', error);


flavio-b:
I just signed the CLA. Please advise if any other step is required.

nwtn:
Thanks @flavio-b this looks great! I’m not sure why the CLA status is still failing, I’ll look into that.

nwtn:
@flavio-b could you try signing the CLA again just in case?

flavio-b:
@nwtn I tried again, but it says "GitHub username has already signed a CLA".

nwtn:
Got it sorted, thanks again @flavio-b!

flavio-b:
Thanks for merging this so quickly! Looking forward to 8.5.2.

sergey-alekseev:
Thanks! I'll forward this to the team. Should a new version 8.5.2 be released?

flavio-b:
@nwtn, 
I hate to bring this topic up again, but it looks like that flash message that tells the user to re-enable third-party cookies, is not showing up anymore. It used to display after the user gets sent back to the Shopify Apps's tab.

I've tested with the latest Chrome and Edge on Windows.

nwtn:
That flash message was being handled on shopify’s side (independent of this gem or app bridge). I’ll take a look to see if there were any changes that might have caused this.

In the meantime, @flavio-b can you post a gif of the behaviour you’re seeing now?

flavio-b:
Thanks! Here's the Gif (Chrome 73.0.3683.86  on macOS Mojave):

error

sergey-alekseev:
@flavio-b looks like there is a problem with your gif.

flavio-b:
@sergey-alekseev , hmm.. it can see it fine from here. If you're talking about the flickering, it's because of the quick redirects that happens when you turn off third-party cookies. The cutoff screen is just how I recorded a segment of the screen. The idea is that an red alert should have displayed at the end, but it doesn't.

nwtn:
The gif works for me, thanks.

I am able to reproduce the issue. I took a look at the code that’s producing the flash and there are no obvious recent changes that should affect this, so I’ll need to do some more exploration. I’m going to open a separate issue for this.

nwtn:
@flavio-b thanks for the heads-up on this. We found the cause and will have some folks work on it early next week.

flavio-b:
@nwtn, thanks for the prompt reply, and for investigating this issue. Much appreciated.

nwtn:
@flavio-b everything should be back to working as expected again. thanks for the heads-up on this!

flavio-b:
@nwtn, thanks for the follow up. I'll do another test later.

PR #703, opened 2019-03-11
Add expiry to cookies to allow persisted cookies in example

Jipsica:
See https://github.com/Shopify/merchant-experience-of-apps/issues/196

In order to optimize performance of apps on mobile, subsequent load time can be reduced if the app uses persisted cookies. This updates the example to include this suggestion.

### DIFF
--- a/example/config/initializers/session_store.rb
+++ b/example/config/initializers/session_store.rb
@@ -1,3 +1,3 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: '_example_session'
+Rails.application.config.session_store(:cookie_store, key: '_example_session', expire_after: 14.days)


Jipsica:
🎩 

Screen Shot 2019-03-11 at 1 17 45 PM

jamiemtdwyer:
A few notes:

- I don't know how many people actually even view the example application for this project, let alone dig through all of the initializers.
- Testing on the latest version of `shopify_app`, this file does not appear to be generated when using the default (or any?) generators.

I'd suggest looking at how [other intiializers](https://github.com/Shopify/shopify_app/blob/master/lib/generators/shopify_app/install/templates/omniauth.rb) are generated, if we want to encourage developers to include this file in their projects.

PR #705, opened 2019-03-12
Fix the request by adding the missing params

### LINKED ISSUE #701: Blank page and javascript error when loading app on iOS
When loading an app using version 8.5.1 of them gem, on iOS 12.1.4, I got a blank screen. I checked with the debugger and it turned out to be a javascript error

alt

A few other people on my team did get the same problem. One solved it just by hitting the back button, then it seemed to redirect/auth properly. Three of us went to our login page and hit the install button again which seemed to refresh our sessions and correct the error (stale cookies maybe?).

I would expect that this wouldn't need to be done normally in this case and re-auth (or whatever the problem was here) should happen automatically?

JoePym:
This provisionally fixes #701. I've tophatted it on a local app, but I'm struggling with how I'd test this (and it makes me super uncomfortable there's no tests).

Open to feedback!

### DIFF
--- a/app/assets/javascripts/shopify_app/partition_cookies.js
+++ b/app/assets/javascripts/shopify_app/partition_cookies.js
@@ -1,7 +1,8 @@
 (function() {
   document.addEventListener("DOMContentLoaded", function() {
-    var storageAccessHelper = new StorageAccessHelper();
+    var redirectTargetElement = document.getElementById("redirection-target");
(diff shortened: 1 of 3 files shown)


ccleung:
does this fix both:
desktop issue https://github.com/Shopify/facebook-ads/issues/3781 and mobile issue https://github.com/Shopify/facebook-ads/issues/3769?

Issue #690, opened 2019-02-15
AfterAuthenticateJob misleading documentation => CAUSED US A LOT OF HEADACHE


nafaabout:
The documentation says: 

> If your app needs to perform specific actions after it is installed ShopifyApp can queue or run a job of your choosing (note that we already provide support for automatically creating Webhooks and Scripttags). To configure the after authenticate job update your initializer as follows:

I used it to send a welcome email to the user after he installs the app about how to better use it. After a couple days we received a flood of complaints, people are receiving too many emails even if **THEY HAVE BEEN USING THE APP FOR MONTHS NOW**. 
The problem was that the documentation was misleading and it turns out that this job is getting run every time a new session is created for the shop, for example when the user closes and reopens his browser.

**I suggest to change it to** 
> If your app needs to perform specific actions after the shop is authenticated successfully (i.e every time a new session is created for the shop), ShopifyApp can queue a job ... bla bla bla.

**And add this to make sure the message is clear**

> If you want to perform that action only once, e.g send a welcome email to the user when he installs the app, you should make sure that this action is idempotent, i.e make sure from your side that this action can only be performed once.

gfscott:
Closing this issue since change was merged in #706

PR #706, opened 2019-03-17
Make clear what AfterAuthenticateJob does (#690)

nafaabout:


README.md
```diff
@@ -364,6 +364,9 @@ We've also provided a generator which creates a skeleton job and updates the ini
 bin/rails g shopify_app:add_after_authenticate_job
 ```
 
+If you want to perform that action only once, e.g send a welcome email to the user when he installs the app, you should make sure that this action is idempotent, meaning that it won't have an impact if run multiple times.
```
alanhill: e.g send a welcome email to the user when ~he installs the app~ they install the app

README.md
```diff
@@ -342,7 +342,7 @@ If `src` responds to `call` its return value will be used as the scripttag's sou
 AfterAuthenticate Job
 ---------------------
 
-If your app needs to perform specific actions after it is installed ShopifyApp can queue or run a job of your choosing (note that we already provide support for automatically creating Webhooks and Scripttags). To configure the after authenticate job update your initializer as follows:
+If your app needs to perform specific actions after the shop is authenticated successfully (i.e every time a new session is created), ShopifyApp can queue or run a job of your choosing (note that we already provide support for automatically creating Webhooks and Scripttags). To configure the after authenticate job update your initializer as follows:
```
markiiikram: Thanks for your pull request @NafaaBout.

As Shopify stores can have multiple users, we suggest "If your app needs to perform specific actions after the **user** is authenticated successfully (i.e every time a new session is created), ShopifyApp can queue..." instead.

README.md
markiiikram: Please update to non-gendered language: "e.g send a welcome email to the user when **they install** the app..." Thanks.

PR #708, opened 2019-03-21
Add a concern ShopifyApp::Authenticated so apps don't have to inherit from AuthenticatedController

JackMc:
Currently, any app that uses `shopify_app` needs to inherit from `ShopifyApp::AuthenticatedController` so those parts of their app receive Shopify authentication. This has the unfortunate consequence that controllers cannot inherit from the application's actually `ApplicationController` and leads to a situation where there's no place for shared functionality of both authenticated and unauthenticated controllers can go.

This PR adds a `Authenticated` concern, which can simply be included in an `AuthenticatedController` defined by the app, which can then inherit from `ApplicationController` and benefit from any shared functionality therein.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+8.6.0
+-----
+
+* Added an `Authenticated` concern to allow gem users to inherit from a custom `AuthenticatedController` instead of
+  `ShopifyApp::AuthenticatedController`
+
 8.5.1
 -----
 
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Table of Contents
 * [**ScripttagsManager**](#scripttagsmanager)
 * [**AfterAuthenticate Job**](#afterauthenticate-job)
 * [**ShopifyApp::SessionRepository**](#shopifyappsessionrepository)
-* [**AuthenticatedController**](#authenticatedcontroller)
+* [**Authenticated**](#authenticated)
 * [**AppProxyVerification**](#appproxyverification)
  * [Recommended Usage](#recommended-usage)
 * [**Troubleshooting**](#troubleshooting)
@@ -397,10 +397,12 @@ ShopifyApp::SessionRepository
 
 If you only run the install generator then by default you will have an in memory store but it **won't work** on multi-server environments including Heroku. If you ran all the generators including the shop_model generator then the `Shop` model itself will be the `SessionRepository`. If you look at the implementation of the generated shop model you'll see that this gem provides a concern for the `SessionRepository`. You can use this concern on any model that responds to `shopify_domain` and `shopify_token`.
 
-AuthenticatedController
------------------------
+Authenticated
+-------------
 
-The engine includes a controller called `ShopifyApp::AuthenticatedController` which inherits from `ActionController::Base`. It adds some before_filters which ensure the user is authenticated and will redirect to the login page if not. It is best practice to have all controllers that belong to the Shopify part of your app inherit from this controller. The HomeController that is generated already inherits from AuthenticatedController.
+The engine provides a `ShopifyApp::Authenticated` concern which should be included in any controller that is intended to be behind Shopify OAuth. It adds `before_action`s to ensure that the user is authenticated and will redirect to the Shopify login page if not. It is best practice to include this concern in a base controller inheriting from your `ApplicationController`, from which all controllers that require Shopify authentication inherit.
+
+For backwards compatibility, the engine still provides a controller called `ShopifyApp::AuthenticatedController` which includes the `ShopifyApp::Authenticated` concern. Note that it inherits directly from `ActionController::Base`, so you will not be able to share functionality between it and your application's `ApplicationController`.
 
 AppProxyVerification
 --------------------
--- a/lib/generators/shopify_app/shopify_app_generator.rb
+++ b/lib/generators/shopify_app/shopify_app_generator.rb
@@ -9,6 +9,7 @@ module ShopifyApp
       def run_all_generators
         generate "shopify_app:install #{@opts.join(' ')}"
         generate "shopify_app:shop_model"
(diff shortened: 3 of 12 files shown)


clayton-shopify:
This seems like a good idea.

Since the `Authenticated` concern is meant to be used by apps, it should probably be documented and tested as well.

JackMc:
Agreed about documentation! I'm not sure where the best place to fix that up is, but I'll find it :) With regards to tests, it doesn't actually implement any functionality and the old `AuthenticatedController` does not have any, but I will try and write some up.

clayton-shopify:
What I had in mind for tests were checks that the `ShopifyApp::Authenticated` concern exists and can be included in a Rails controller, and maybe also that it does a couple of the basic things it's expected to.

JackMc:
I can do that! I'm also going to update the generator and example app to use this pattern.

JackMc:
:tophat:'d this with a new Rails app and it seems to work.

JackMc:
Added a test that the concern exists, can be included in a controller, and includes the associated modules.

lib/generators/shopify_app/shopify_app_generator.rb
```diff
@@ -9,6 +9,7 @@ def initialize(args, *options)
       def run_all_generators
         generate "shopify_app:install #{@opts.join(' ')}"
         generate "shopify_app:shop_model"
+        generate("shopify_app:authenticated_controller")
```
clayton-shopify: I presume the parentheses are here because Rubocop wanted them. May as well add them to the other `generate` lines to match.

clayton-shopify:
The `AuthenticatedController` section of the README will need to be updated, and the `Authenticated` module should be documented there as well.

clayton-shopify:
The change should also be added to the change log, and I guess the version number should be bumped to 8.6.0 since this is adding a feature.

JackMc:
Bumped the gem, added changelog entries, and fixed up the README. Let me know what y'all think!

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+8.6.0
+-----
+
+* Added an `Authenticated` concern to allow gem users to inherit from a custom `AuthenticatedController` instead of
+  Shopify's `AuthenticatedController`
```
clayton-shopify:
```suggestion
  `ShopifyApp::AuthenticatedController`
```

README.md
```diff
@@ -397,10 +397,12 @@ ShopifyApp::SessionRepository
 
 If you only run the install generator then by default you will have an in memory store but it **won't work** on multi-server environments including Heroku. If you ran all the generators including the shop_model generator then the `Shop` model itself will be the `SessionRepository`. If you look at the implementation of the generated shop model you'll see that this gem provides a concern for the `SessionRepository`. You can use this concern on any model that responds to `shopify_domain` and `shopify_token`.
 
-AuthenticatedController
------------------------
+Authenticated
+-------------
+
+The engine provides a `ShopifyApp::Authenticated` concern which should be included in any controller that is intended to be behind Shopify OAuth. It adds `before_action`s to ensure that the user is authenticated and will redirect to the Shopify login page if not. It is best practice to include this concern in a base controller inheriting from your `ApplicationController`, from which all controllers under Shopify authentication inherit.
```
clayton-shopify:
```suggestion
The engine provides a `ShopifyApp::Authenticated` concern which should be included in any controller that is intended to be behind Shopify OAuth. It adds `before_action`s to ensure that the user is authenticated and will redirect to the Shopify login page if not. It is best practice to include this concern in a base controller inheriting from your `ApplicationController`, from which all controllers that require Shopify authentication inherit.
```

JackMc: Done ✅

CHANGELOG.md
JackMc: Done ✅

README.md
```diff
@@ -397,10 +397,12 @@ ShopifyApp::SessionRepository
...
 
-The engine includes a controller called `ShopifyApp::AuthenticatedController` which inherits from `ActionController::Base`. It adds some before_filters which ensure the user is authenticated and will redirect to the login page if not. It is best practice to have all controllers that belong to the Shopify part of your app inherit from this controller. The HomeController that is generated already inherits from AuthenticatedController.
+For backwards compatibility, there exists a `ShopifyApp::AuthenticatedController` which can be inherited from and includes this concern. It inherits directly from `ActionController::Base`, though, and you will not be able to use any shared functionality from your `ApplicationController`.
```
clayton-shopify:
```suggestion
For backwards compatibility, the gem still provides a controller called `ShopifyApp::AuthenticatedController` which includes the `ShopifyApp::Authenticated` concern. Note that it inherits directly from `ActionController::Base`, so you will not be able to use any shared functionality from your application's `ApplicationController`.
```

JackMc: Thanks!!

README.md
```diff
@@ -397,10 +397,12 @@ ShopifyApp::SessionRepository
...
 
-The engine includes a controller called `ShopifyApp::AuthenticatedController` which inherits from `ActionController::Base`. It adds some before_filters which ensure the user is authenticated and will redirect to the login page if not. It is best practice to have all controllers that belong to the Shopify part of your app inherit from this controller. The HomeController that is generated already inherits from AuthenticatedController.
+For backwards compatibility, the gem still provides a controller called `ShopifyApp::AuthenticatedController` which includes the `ShopifyApp::Authenticated` concern. Note that it inherits directly from `ActionController::Base`, so you will not be able to use any shared functionality from your application's `ApplicationController`.
```
tylerball: `s/the gem/the engine/` We seem to be using "engine" to describe this everywhere else in the README, so let's keep that consistent.

For clarity let's rephrase the last sentence: `...so you will not be able to share functionality between it and your application's ApplicationController.`

PR #709, opened 2019-03-22
Restore the original root_url to fix a leaky test.

clayton-shopify:
Recently I observed these test failures: https://travis-ci.org/Shopify/shopify_app/jobs/509660290

The failures can be reproduced locally with `SEED=6112 bundle exec rake test`.

They occur because a test modifies the `root_url` configuration option and fails to reset it at the end of the test.

### DIFF
--- a/test/shopify_app/configuration_test.rb
+++ b/test/shopify_app/configuration_test.rb
@@ -25,11 +25,17 @@ class ConfigurationTest < ActiveSupport::TestCase
   end
 
   test "can set root_url which affects login_url" do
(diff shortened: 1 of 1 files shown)


clayton-shopify:
There's nothing I can do about the first Policial warning; the `begin` block is required prior to Ruby 2.5. I didn't bother with the parenthesis warning since none of the other `assert_equal`s use it but I'm happy to add the parentheses if you prefer.

tiffanytse:
Sounds good then; as long as the it'll let you deploy with policial failing.

Issue #700, opened 2019-02-28
WebhooksManager#current_webhooks raises NoMethodError: undefined method `index_by' for nil:NilClass


rochlefebvre:
Encountered with shopify_app-8.3.2 in a rails 5.2.2/ruby 2.6.0 application.
```
{
  "class": "ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper",
  "wrapped": "ShopifyApp::WebhooksManagerJob",
  # ...
  "error_message": "undefined method `index_by' for nil:NilClass",
  "error_class": "NoMethodError",
  "failed_at": 1551291482.9294562,
  "retry_count": 3,
  "error_backtrace": [
    "/app/vendor/bundle/ruby/2.6.0/gems/shopify_app-8.3.2/lib/shopify_app/managers/webhooks_manager.rb:58:in `current_webhooks'",
    "/app/vendor/bundle/ruby/2.6.0/gems/shopify_app-8.3.2/lib/shopify_app/managers/webhooks_manager.rb:54:in `webhook_exists?'",
    "/app/vendor/bundle/ruby/2.6.0/gems/shopify_app-8.3.2/lib/shopify_app/managers/webhooks_manager.rb:28:in `block in create_webhooks'",
    "/app/vendor/bundle/ruby/2.6.0/gems/shopify_app-8.3.2/lib/shopify_app/managers/webhooks_manager.rb:27:in `each'",
    "/app/vendor/bundle/ruby/2.6.0/gems/shopify_app-8.3.2/lib/shopify_app/managers/webhooks_manager.rb:27:in `create_webhooks'",
    "/app/vendor/bundle/ruby/2.6.0/gems/shopify_app-8.3.2/lib/shopify_app/jobs/webhooks_manager_job.rb:11:in `block in perform'",
    "/app/vendor/bundle/ruby/2.6.0/gems/shopify_api-5.2.0/lib/shopify_api/session.rb:30:in `temp'",
    "/app/vendor/bundle/ruby/2.6.0/gems/shopify_app-8.3.2/lib/shopify_app/jobs/webhooks_manager_job.rb:9:in `perform'",
    "/app/vendor/bundle/ruby/2.6.0/gems/activejob-5.2.2/lib/active_job/execution.rb:39:in `block in perform_now'",
```
The error is raised from this statement:
[...]

PR #720, opened 2019-04-10
Handle Webhook.all raturning `nil` and raising on `index_by`

kmcphillips:
## Problem

Described in #700, it is possible for `.all` on an ActiveResource object to return `nil` and not just an empty array.

This causes the `WebhooksManager` to raise calling `nil#index_by` in this case. 

## Solution

Coerce `nil.to_a` to force `[]`. It's done a few lines above in the same file in the same way.


kmcphillips:
Closes #700

Issue #603, opened 2018-06-19
Using nested route throws routing exception on install


cerveny:
I am using Rails 5.2.0, shopify_app gem 8.2.6 a shopify_api gem 4.11.0. I created bare shopify application. Then I added configuration for nested route as outlined in Nested Route section of Mounting the Engine.

When I start server and visit "localhost:3000/nested" I am redirected to login page at "localhost:3000/nested/login". After I enter development store name and hit install, I am redirected to "localhost:3000/nestedauth/shopify" which throws "No route matches [GET] "/nestedauth/shopify"" and instalation stops. 

![shopify-nested-route-exception](https://user-images.githubusercontent.com/5987984/41576349-37e9903c-7387-11e8-86f5-afcf00a755af.png)

My routes.rb look like this
```
Rails.application.routes.draw do
  scope '/nested' do
    root :to => 'home#index'
  end
  mount ShopifyApp::Engine, at: '/nested'
end
```

My shopify_app.rb
```
ShopifyApp.configure do |config|
  config.root_url = '/nested'
  config.application_name = "My Shopify App"
  config.api_key = ...rest as generated by generator
```
  
My omniauth.rb
```
  Rails.application.config.middleware.use OmniAuth::Builder do
    provider :shopify,
            callback_path: '/nested/auth/shopify/callback',
             ShopifyApp.configuration.api_key ...rest as generated by generator
```

juan267:
I'm experiencing the same problem, any updates on this?

Shine18:
Any update on this?

ragalie:
Are you experiencing the behaviour with 8.4.0? The logic has changed considerably from 8.2.6 and may be resolved at this point.

Shine18:
I added a route '/shopify' in my routes files. and pointed it to the index of app's home controller. That's how i fixed it.

andreas-sotnik-2k19:
I have the same issue on 8.6.0 version...

I think the issue may be caused by `authenticate_in_context` method of `ShopifyApp::SessionsController`:

```
def authenticate_in_context
     redirect_to "#{main_app.root_path}auth/shopify"
end
```

It takes root_path of the app without taking nesting into consideration...


I'm trying to do a local patch like this:

```
ShopifyApp::SessionsController.class_eval do
  def authenticate_in_context
    redirect_to "#{ShopifyApp.configuration.root_url}/auth/shopify"
  end
end
```

This fixes the url, but still there is no route for `/app/auth/shopify`...

[...]

PR #726, opened 2019-04-21
Clarify how nested routes affect Rails routing

davetron5000:
# Problem

If you follow the instructions to set up your app, then decide to use nested routes, things don't work as advertised.  See #603.  You need to also know that the engine doesn't render the Shopify Admin UI so you have to route to that *and* you have to know that going to that url triggers the "add your app to shopify" UI.  When you are totally new to this + shopify, these things are not obvious.

# Solution

Document it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,19 @@ provider :shopify,
   callback_path: '/nested/auth/shopify/callback'
 ```
 
+You may also need to change your `config/routes.rb` to render a view for `/nested`, since this is what will be rendered in the Shopify Admin of any shops that have installed your app.  The engine itself doesn't have a view for this, so you'll need something like this:
+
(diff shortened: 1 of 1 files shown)


katiedavis:
Hi @davetron5000
Thanks for your contribution, we're looking into it. In the meantime, would you mind signing the CLA? https://cla.shopify.com/
Thanks!

davetron5000:
done

davetron5000:
Gonna bump this (promise this is the only time I will).  This is a simple enough doc change so maybe it can merged?  If not, feel free to close

nwtn:
@davetron5000 thanks so much for this. I’ll sort out the rebase and merge this today.

PR #729, opened 2019-04-29
fix after_authneticate_job typo

nwtn:
closes #669


nwtn:
ignoring policial because it’s an existing issue
ignoring CLA because it’s getting triggered by the shopify-admins account, which doesn’t need to sign

Issue #734, opened 2019-05-02
Safari ITP cookie screen UX problems


nwtn:
* [ ] the current screen asking for authentication is not responsive on mobile
* [x] ~the auth screen/ITP flow shouldn’t happen at all in shopify mobile, only browsers~ apps should be able to solve this by upgrading to shopify_app 9.0.2
* [ ] the copy is sometimes not showing up in English (need reproduction steps cc @darrinhenein)
* [ ] the copy explicitly mentions third-party apps, even for Shopify-built apps

nwtn:
cc @adamwillking for the copy

amythibodeau:
@adamwillking Just a note that this is a high priority fix. Our POS channel is getting a lot of attention right now and this experience is pretty jank.

darrinhenein:
I was able to reproduce the German language UI just by opening the channel in Shopify Mobile on my Canadian-localized iPhone X, nothing more specific.

Also want to confirm that there is no workaround to this screen? Can we provide a custom version for POS or specific first-party apps?

ragalie:
I wonder if the user agent has changed for Shopify Mobile in a way that `shopify_app` isn't supporting (or maybe `shopify_app` isn't on the latest version of `browser_sniffer`?). I'm pretty sure we set this up so that this screen never appears in Shopify Mobile since ITP doesn't apply in the `WebView`s it's using. cc @francinen

darrinhenein:
![IMG_1B059530D71E-1](https://user-images.githubusercontent.com/792313/57105834-e1a73900-6cf9-11e9-919d-9640e5be3dd6.jpeg)

iOS 12.2

ragalie:
Ok I'm pretty sure a bump to browser_sniffer will significantly decrease the incidence of this since it'll stop this showing up in Shopify Mobile: https://github.com/Shopify/browser_sniffer/commit/7de9740c2a6516315bd456b3d42e21e350410f4d

I'll do that right now and open a PR.

We'll should also add a caution tape bot that reminds people to update `browser_sniffer` in both core *and* shopify_app when they update it.

ragalie:
OK, I opened that PR. After we force the version upgrade here, each app that uses `shopify_app` needs to upgrade the version of `shopify_app` they're using.

@nwtn I'll get reviews on this PR, can you own cutting a new `shopify_app` release after the PR goes in?

francinen:
Thanks, @ragalie!

[...]

PR #737, opened 2019-05-03
add meta viewport to fix responsive problems in ITP flow

nwtn:
see #734 

i’m going to need some help 🎩 this since i’m not super familiar with what all the states in the flow are supposed to look like.


nwtn:
🎩 on Safari iOS looks ok to me, or at least better than status quo.

PR #738, opened 2019-05-03
 remove “third-party” from ITP copy

nwtn:
This is an MVP fix, English-only, for the copy in the ITP flow.

~(the updated yarn lock is unrelated, but should be committed)~


tiffanytse:
So the yarn updates are in another PR I assume? 

Otherwise LGTM. 👍

nwtn:
> So the yarn updates are in another PR I assume?

Ya I’m not sure which PR this was from, but it looks like maybe `yarn install` wasn’t run on a previous commit/PR that changed the deps.

On second thought, though, I’m going to split that into a separate PR.

gfscott:
Since the English copy change is approved I've sent it to translation. PR can merge once that check passes.

nwtn:
i need to do a release, so i've made translations async

PR #743, opened 2019-05-07
Allow #after_authenticate_job to accept a string class name

kmcphillips:
## Problem

We are encountering the "removed from the module tree but still active" errors in development:

Slack_-_Shopify

This is being caused by referencing the job as a class name. Then when autoload triggers in some changes, it causes that job to be reloaded. But since the `ShopifyApp.configure` block is not done in a `to_prepare` and is only done once, the reference is held and the two versions of the class conflict.

## Solution

Allow for the job to be passed in as a string and constantized. That way the class is looked up by name on each use. This is _slightly_ slower but not consequentially so.

Going forward the generator will output a string, but the literal class is still valid. I have done this approach so that there's no backwards compatiblity problems, and no behaviour difference on existing apps. We simply support both class and string now, with string being preferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,9 +347,11 @@ If your app needs to perform specific actions after the user is authenticated su
 
 ```ruby
 ShopifyApp.configure do |config|
-  config.after_authenticate_job = { job: Shopify::AfterAuthenticateJob }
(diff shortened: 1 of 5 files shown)


nwtn:
Thank you!

Issue #618, opened 2018-08-15
[Docs request] Add concurrent access mode examples (offline & online auth)


MatiasFMolinari:
### Overview
Shopify's [OAuth API](https://help.shopify.com/en/api/getting-started/authentication/oauth/api-access-modes) and [Omniauth gem](https://github.com/Shopify/omniauth-shopify-oauth2) supports both offline and online access modes. We should provide a clear example of how to use them both, concurrently, in our apps. More specifically, how to setup online access for only the `ShopifyApp::AuthenticatedController` while still maintaining offline access for `ShopifyApp::AppProxyVerification` and `ShopifyApp::WebhookVerification`.

### Previous discussions
This has been discussed before, see #580, but there has not been a clear solution moving forward. As of today, there aren't any [public repos in Github implementing this](https://github.com/search?q=OmniAuth+Builder+provider+shopify+per_user_permissions+language%3Aruby&type=Code) either. From what I understood, you can build two Shopify tokens, one for offline and one for online, but I'm not sure how to scope them properly.

### Examples

Here's an example of what I'm talking about:

#### Offline setup (default) <-- This works
AuthenticatedControllers, proxy app, and webhooks work properly, but you can't access individual user info on AuthenticatedControllers.
```ruby
# config/omniauth.rb
Rails.application.config.middleware.use OmniAuth::Builder do
  # Offline access
  provider :shopify,
  ShopifyApp.configuration.api_key,
  ShopifyApp.configuration.secret,
  scope: ShopifyApp.configuration.scope,
  setup: lambda { |env|
    strategy = env['omniauth.strategy']

    shopify_auth_params = strategy.session['shopify.omniauth_params']&.with_indifferent_access
    shop = if shopify_auth_params.present?
      "https://#{shopify_auth_params[:shop]}"
    else
      ''
    end

    strategy.options[:client_options][:site] = shop
  }
end
```

#### Online setup <-- This works
AuthenticatedControllers work properly, but proxy app and webhooks are `Unauthorized`.
```ruby
# config/omniauth.rb
Rails.application.config.middleware.use OmniAuth::Builder do
  # Online access
  provider :shopify,
  ShopifyApp.configuration.api_key,
  ShopifyApp.configuration.secret,
  scope: ShopifyApp.configuration.scope,
  per_user_permissions: true, # <-- this changed
  setup: lambda { |env|
    strategy = env['omniauth.strategy']

[...]

PR #759, opened 2019-06-05
Add ability to enable simultaneous offline & online access modes

joshreeves:
We needed online access in our app while still storing and retrieving an offline access token in the database so that webhooks and background processes would continue to work. This PR solves that problem by optionally (based on the `online_access` configuration value) expanding the authentication process to include 2 passes through the Shopify oauth flow. It is inspired by the discussion here: https://github.com/Shopify/shopify_app/issues/618

Would love feedback on the direction of this, and the liklihood of getting it merged. Thanks!

@EiNSTeiN- @nwtn

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,13 +238,25 @@ To enable per user authentication you need to update the `omniauth.rb` initializ
 provider :shopify,
   ShopifyApp.configuration.api_key,
   ShopifyApp.configuration.secret,
+  scope: ShopifyApp.configuration.scope
+
+provider :shopify,
+  ShopifyApp.configuration.api_key,
+  ShopifyApp.configuration.secret,
   scope: ShopifyApp.configuration.scope,
-  per_user_permissions: true
+  per_user_permissions: true,
+  path_prefix: "/online_auth"
+```
+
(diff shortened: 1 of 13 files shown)


marisveide:
Hi!

Is there any progress on this?
We would need to have the offline token, but at the same time see the Shopify User ID, and its other details. That is needed for security reasons - to be able to show who accessed the app, and did what, and allow app admins to set permissions for other Shopify users who are not admins.

Maris

joshreeves:
Hi @marisveide - I haven't heard anything from the internal team about this being merged. I do see that they recently merged (and then reverted) a PR that implements similar functionality (https://github.com/Shopify/shopify_app/pull/760).

marisveide:
Thank you, @joshreeves !
Then, I guess, we should hold still and monitor those issues - maybe they will fix something and merge then...

What we need is to actually have those both modes accessible in parallel - offline and online.

nickmealey:
Would love to see this get merged in!

tanema:
This was just shipped. You can now use a shop_session_store and a user_session_store in your app config. You can use them concurrently or on their own

PR #760, opened 2019-06-05
Per-user token support

ashchristopher:
**Problem:** The current version of `shopify_app` only supports per-shop tokens even though you can specify the per-user variety of token when configuring `omniauth`.

**Solution:** This change introduces per-user token management to the shopify_app gem. It follows the same pattern as the Shop session management in that session data is stored in the `user` table similar to what we saw with the `shop` table. The change should be backwards compatible as it defaults to Shop sessions.

To 🎩 I tested with the [pos-channel](https://github.com/Shopify/pos-channel) application. I installed the local gem rather than downloading the package.

```
+# gem 'shopify_app', '~> 11.1.0'
+gem 'shopify_app', path: "/Users/ashchristopher/src/github.com/Shopify/shopify_app"
```
I ensured that the pos-channel continues to work as expected with no changes to the current pos-channel code.

To 🎩 the per-user changes, you can install the gem locally (see above) and around with a new `shopify_app` application.

Some things to look for:
- [x] When in per-shop configuration, each new shopify domain gets it's own row in the `shop` table
- [x] When in per-user configuration, each new user gets it's own row in the `user` table
- [x] When in per-user configuration, when logged in as user A and you sign-out and log in as User B, the tokens used are different


lib/shopify_app/session/storage_strategies/user_storage_strategy.rb
```diff
@@ -0,0 +1,26 @@
...
+        user = User.find_or_initialize_by(shopify_user_id: user[:id])
+        user.shopify_token = session.token
+        user.shopify_domain = session.domain
+        user.save!
+        user.id
+      end
+
+      def self.retrieve(id)
+        return unless id
+        if user = User.find_by(shopify_user_id: id)
```
ashchristopher: Is the shopify_user_id unique to Shopify or unique to the shop?

ashchristopher: Same shopify_user_id in multiple shops?

clayton-shopify:
It looks like this lets you choose between either online (per-user) or offline (per-shop) access. It's not uncommon for apps to want both: online when the user is interacting with the app through its web interface, and offline for background work. It would be great if shopify_app could accommodate that use case.

dan-gamble:
Would this allow you, in theory, to know which logged in user is accessing your app?

Could this also mean, on plus, that you could read those users permissions and use them to mask off relevant bits of the app as Shopify does?

lib/shopify_app/session/storage_strategies/shop_storage_strategy.rb
```diff
@@ -0,0 +1,24 @@
+module ShopifyApp
+  module SessionStorage
+    class ShopStorageStrategy
```
tylerball: I think it's worth having specific tests for theses strategy classes.

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -55,10 +55,16 @@ def set_shopify_session
         token: token,
         api_version: ShopifyApp.configuration.api_version
       )
-
-      session[:shopify] = ShopifyApp::SessionRepository.store(session_store)
+      session[:shopify] = ShopifyApp::SessionRepository.store(session_store, user:associated_user)
```
tylerball: space: `user: associated_user`

lib/shopify_app/session/session_storage.rb
```diff
@@ -3,12 +3,13 @@ module SessionStorage
...
 
     included do
-      validates :shopify_domain, presence: true, uniqueness: { case_sensitive: false }
+      validates :shopify_domain, presence: true, uniqueness: { case_sensitive: false },
+        unless: Proc.new {|_| ShopifyApp.configuration.per_user_tokens }
       validates :shopify_token, presence: true
       validates :api_version, presence: true
     end
 
-    def with_shopify_session(&block)
```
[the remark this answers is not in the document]
ashchristopher: Thanks for calling this out @tanema - I overlooked how this was being used. I have reverted back.

lib/shopify_app/session/session_storage.rb
```diff
@@ -18,23 +19,17 @@ def with_shopify_session(&block)
...
-        shop.save!
-        shop.id
+
+      def strategy_klass
+        ShopifyApp.configuration.per_user_tokens ? ShopifyApp::SessionStorage::UserStorageStrategy : ShopifyApp::SessionStorage::ShopStorageStrategy
       end
 
-      def retrieve(id)
-        return unless id
+      def store(auth_session, user:nil)
```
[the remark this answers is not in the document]
ashchristopher: Fixed

lib/shopify_app/session/storage_strategies/shop_storage_strategy.rb
ashchristopher: Hey @tylerball - these models don't exist in the gem - they are generated, and only exist in the subsequent project that is using the `shopify_app` gem. Do you have any suggestions on how we could test?

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -14,7 +14,7 @@ class ShopifyDomainNotFound < StandardError; end
       rescue_from ActiveResource::UnauthorizedAccess, :with => :close_session
     end
 
-    def shopify_session
+    def manage_shopify_session
```
tanema: Why did you change this methods name as well? This isnt really even a management method more of a `ensure_shopify_session` however I think it would be best to leave it as it was

lib/shopify_app/session/session_storage.rb
```diff
@@ -3,7 +3,8 @@ module SessionStorage
     extend ActiveSupport::Concern
 
     included do
-      validates :shopify_domain, presence: true, uniqueness: { case_sensitive: false }
+      validates :shopify_domain, presence: true, uniqueness: { case_sensitive: false },
+        unless: Proc.new {|_| ShopifyApp.configuration.per_user_tokens }
```
tanema: Why do we not need a domain for per user session? How will we identify the domain that the API requests are sent to?

lib/generators/shopify_app/user_model/templates/user.rb
```diff
@@ -0,0 +1,7 @@
+class User < ActiveRecord::Base
+  include ShopifyApp::SessionStorage
```
tanema: Why use a single session storage mixin rather than a different mixin for each Shop or User? Is seems needlessly abstract.

lib/shopify_app/session/storage_strategies/shop_storage_strategy.rb
tanema: Honestly there should have been one already for the shop session in the same way as the InMemoryStore but you should be able to test these strategies in the same way as the InMemoryStore. These models exist in our Gem, and all you need to do is define a TestClass within your tests that gets extended by your concern.

service.yml
```diff
@@ -2,6 +2,6 @@ audience: partner
 classification: library
 org_line: App & Partner Platform
 owners:
-  - Shopify/app-partner-dev-tools-education
+  - Shopify/platform-dev-tools-education
```
tanema: Thanks!

test/controllers/callback_controller_test.rb
```diff
@@ -160,11 +167,16 @@ def mock_shopify_user_omniauth
         :shopify,
         provider: :shopify,
         uid: 'shop.myshopify.com',
-        credentials: { token: '1234' },
-        extra: { associated_user: 'user_object' }
+        credentials: {token: '1234' },
```
tanema: { token: '1234' }

(we recently realized that rubocop is not running on this repo so there may be some style issues 🤕)

lib/shopify_app/session/session_storage.rb
ashchristopher: In per-user session - many users will have the same domain. In per-shop session, the domain is the unique identifier.

lib/shopify_app/controller_concerns/login_protection.rb
ashchristopher: I was ultimately trying to free up the `shop_session` so that we could replace the `@shop_session` with `@shopify_session`. `@shop_session` made sense when `shopify_app` was limited to just `Shop` tokens - but now that we need to support per-user tokens, calling the Shopify session `@shop_session` isn't really correct. 

I think the `shopify_session` function is poorly named - it's really about setting up the `shop_session` context. But agree that `manage_shopify_session` is also poorly named. I can revert and you guys can decide if you would like to rename in the future.

Let me know what you think about `@shop_session`.

This does bring up another naming issue -

ashchristopher:
> As @clayton-shopify asked, this seems like you can only use one or the other, is this going to be the case? You should also probably update the readme to add more to the readme about enabling this and using a user session store instead. Also what the differences are between the two. Is there more documentation or a project brief on the motivation of all of this so that maybe I can get these answers somewhere else?

Yes - at this time, we only need to support one or the other. Omniauth only supports one or the other from an OAuth perspective - so if there is a use-case that would require both to be present at the same time, it falls outside of our scope and should probably go to your team to investigate as the solution will probably be more involved.

 I will update the `README.md` when we are happy with the implementation, and the API's. I wanted to ensure we are good and on the same page with the approach.

We don't have a brief - this is one part of a greater initiative. I will write a one-pager and circulate rather than elaborate on a public repo.

lib/generators/shopify_app/user_model/templates/user.rb
ashchristopher: @tanema This was for backwards compatibility - any existing apps that use the `shopify_app` gem will have had the `shop.rb` file copied to their app. This `Shop` model extends `ShopifyApp::SessionStorage`.

One of my goals was to have a drop-in replacement that would not require users of the library to make changes to their existing code in order to upgrade so as to keep per-user tokens strictly a feature addition.

Happy to revisit this decision though.

lib/shopify_app/session/session_storage.rb
ashchristopher: from @tanema Does the `unless` apply just to the uniqueness, or to the whole validation statement? Perhaps needs to be split into multiple validates.

lib/shopify_app/session/session_storage.rb
```diff
@@ -3,9 +3,12 @@ module SessionStorage
     extend ActiveSupport::Concern
 
     included do
-      validates :shopify_domain, presence: true, uniqueness: { case_sensitive: false }
       validates :shopify_token, presence: true
       validates :api_version, presence: true
+      validates :shopify_domain, presence: true,
+        if: Proc.new {|_| ShopifyApp.configuration.per_user_tokens }
+      validates :shopify_domain, presence: true, uniqueness: { case_sensitive: false },
+        if: Proc.new {|_| !ShopifyApp.configuration.per_user_tokens }
```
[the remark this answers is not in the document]
ashchristopher: One is applied when we are using shop tokens, and the other when using user tokens. They have different validation requirements.

test/controllers/callback_controller_test.rb
```diff
@@ -50,12 +50,19 @@ class CallbackControllerTest < ActionController::TestCase
     end
 
     test '#callback sets up a shopify session with a user for online mode' do
-      mock_shopify_user_omniauth
-
-      get :callback, params: { shop: 'shop' }
-      assert_not_nil session[:shopify]
-      assert_equal 'shop.myshopify.com', session[:shopify_domain]
-      assert_equal 'user_object', session[:shopify_user]
+      begin
```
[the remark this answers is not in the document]
ashchristopher: If the test throws and exception, rather than fails, won't the singleton configuration object still be set to `true` for all subsequent tests?

lib/shopify_app/session/session_storage.rb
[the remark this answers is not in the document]
bradsokol: Why would domain uniqueness matter only when using shop tokens?

test/controllers/callback_controller_test.rb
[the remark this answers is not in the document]
bradsokol: You leave the `ensure` clause in. You write it this way:
```ruby
test '#callback sets up a shopify session with a user for online mode' do
  ...
ensure
  ...
end
```

bradsokol:
I tophatted with an existing app using shop tokens. 👍

lib/shopify_app/session/session_storage.rb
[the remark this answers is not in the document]
tanema: Because the opposite of per-user is per-shop. There would only be one session for each shop. The domain uniqueness is to ensure that there is only ever one per shop

lib/shopify_app/session/storage_strategies/shop_storage_strategy.rb
ashchristopher: Tests have been added.

lib/shopify_app/session/session_storage.rb
ashchristopher: This was split into multiple validations.

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -55,10 +55,16 @@ def set_shopify_session
...
-
-      session[:shopify] = ShopifyApp::SessionRepository.store(session_store)
+      session[:shopify] = ShopifyApp::SessionRepository.store(session_store, user: associated_user)
       session[:shopify_domain] = shop_name
       session[:shopify_user] = associated_user
+
+      if ShopifyApp.configuration.per_user_tokens
+        # Adds the user_session to the session to determine if the logged in user has changed
+        user_session = auth_hash&.extra&.session
+        raise Exception, "Missing user session signature" if user_session.nil?
```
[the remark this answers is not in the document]
ashchristopher: Fixed

lib/shopify_app/configuration.rb
```diff
@@ -15,6 +15,8 @@ class Configuration
     attr_accessor :scripttags
     attr_accessor :after_authenticate_job
     attr_accessor :session_repository
+    attr_accessor :per_user_tokens
```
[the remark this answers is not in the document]
ashchristopher: Done

lib/shopify_app/session/session_storage.rb
```diff
@@ -18,23 +21,17 @@ def with_shopify_session(&block)
...
 
     class_methods do
-      def store(session)
-        shop = find_or_initialize_by(shopify_domain: session.domain)
-        shop.shopify_token = session.token
-        shop.save!
-        shop.id
+
+      def strategy_klass
+        ShopifyApp.configuration.per_user_tokens ? ShopifyApp::SessionStorage::UserStorageStrategy : ShopifyApp::SessionStorage::ShopStorageStrategy
```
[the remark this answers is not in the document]
ashchristopher: Fixed - though curious if it might be more readable to use an `if` statement?

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+11.4.0
+-----
+* Modularizes durable session storage
+* Introduces per-user token support and user session management (beta)
```
bradsokol: What does "beta" mean here?

test/controllers/callback_controller_test.rb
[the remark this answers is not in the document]
bradsokol: We spoke offline about this. The syntax as written is necessary to support older versions of Ruby.

CHANGELOG.md
tanema: I think this means that it requires a beta flag right now

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -27,12 +27,30 @@ def shopify_session
...
+      if ShopifyApp.configuration.per_user_tokens?
+        invalid_session_data = session[:user_session].nil? || params[:session].nil? # session data was not sent/stored correctly
+        sessions_do_not_match = session[:user_session] != params[:session] # current user is different from stored user
+
+        if invalid_session_data || sessions_do_not_match
+          clear_session = true   
+        end
+      end
+
       if shop_session && params[:shop] && params[:shop].is_a?(String) && (shop_session.domain != params[:shop])
```
slucaskim: Could this be in an `else`, as it only applies to the shop based token case?

test/controllers/callback_controller_test.rb
```diff
@@ -145,7 +152,7 @@ class CallbackControllerTest < ActionController::TestCase
       assert_raise NameError do
         get :callback, params: { shop: 'shop' }
       end
-    end
+   end
```
slucaskim:
```suggestion
    end
```

lib/shopify_app/controller_concerns/login_protection.rb
ashchristopher: Line 49?

slucaskim: It's minor, but:
```suggestion
      else
        if shop_session && params[:shop] && params[:shop].is_a?(String) && (shop_session.domain != params[:shop])
```

ashchristopher: Oh - line 49 needs to apply regardless of whether you are per-user or per-shop. What this does is do an additional check for the user session if you are in per-user mode.

README.md
```diff
@@ -12,61 +12,35 @@ Shopify Application Rails engine and generator
```
slucaskim: I reorganized the readme to move all authentication and session related content into one heading called "Authentication" and then I added new content around user-based token strategy in there.

slucaskim: @tanema @tylerball Are you guys ok with this reorganization change here?

lib/shopify_app/configuration.rb
```diff
@@ -15,6 +15,9 @@ class Configuration
...
+    alias_method :per_user_tokens?, :per_user_tokens
+    attr_accessor :scope
```
slucaskim: @ashchristopher what was this added for?

README.md
```diff
@@ -238,21 +201,79 @@ ShopifyApp.configure do |config|
...
+
+#### Migrating from shop-based to user-based token strategy
+After running the generator, ensure that configuration settings are successfully changed:
 
 ```ruby
+# In the `omniauth.rb` initializer:
 provider :shopify,
-  ShopifyApp.configuration.api_key,
-  ShopifyApp.configuration.secret,
-  scope: ShopifyApp.configuration.scope,
```
tylerball: Why were these lines removed? I think it's wise to still show what's required here.

README.md
tylerball: Looks good! Thanks for cleaning up the TOC

README.md
slucaskim: Kk I added them back.

lib/shopify_app/configuration.rb
ashchristopher: It can be removed - I think it may be due to fixing merge conflicts. It is defined on line 11, so let's get rid of it (assuming you are referring to `attr_accessor :scope`)

PR #762, opened 2019-06-10
Redirect to login for non-GET requests

joshreeves:
We had situations where app users would need to re-authenticate upon submitting non-GET requests. This change ensures that users are redirected back to the route they were on when they initiated the non-GET request.

@nwtn


joshreeves:
@katiedavis - rebased with Shopify_app/master

joshreeves:
@katiedavis - Any chance of getting this merged?

katiedavis:
@joshreeves done! Thanks so much!

PR #764, opened 2019-06-12
[wip] Stores the per-user token in the session

ashchristopher:


ashchristopher:
Did a quick review with @ragalie, and although this approach is simpler, it introduces a potential security issue if the developer using this gem doesn't have secure sessions.

Issue #600, opened 2018-06-18
Webhooks throwing error when deserializing admin_graphql_api_id


jeffse:
We started to receive webhook data including the new `admin_graphql_api_id` attribute today. This has a value of something akin to `gid://shopify/Customer/61476333152`.

The default WebhooksController is serializing the entire hash into the job request. When ActiveJob tries to deserialize the object, it's looking at these gid's and trying to deserialize them into ActiveRecord objects. Since these objects don't exist on our system, we get an `ActiveJob::DeserializationError`.

I think I can work around this by creating a custom WebhooksController that deletes all of the `admin_graphql_api_id` attributes, but I'd like to see a more long-term fix in shopify_app.

EiNSTeiN-:
Which webhook event type is this happening for?

edavis10:
I saw it on themes/publish:

ActiveJob::DeserializationError: Error while trying to deserialize arguments: uninitialized constant Theme

I ended up adding a fake `Theme` class that did nothing but if this happens in more places I'll have to be more heavy-handed with it.

EiNSTeiN-:
Thanks. Looks like an unintended side effect caused by a change that was released this morning. I've passed this information on to the responsible team and they will look into it.

Beans0063:
+1

adrianthedev:
I fixed it like this

https://stackoverflow.com/questions/50914718/shopify-api-error-while-trying-to-deserialize-arguments

swalkinshaw:
@adrianthedev I've been investigating a fix for this and a custom locator was my first thought as well however it doesn't fix the problem for everyone.

Do you have a class defined in your application which matches up the resource of the webhook you're receiving? ie: if its a webhook from `products/create`, do you have a local `Product` class?

Another problem is here: https://github.com/rails/globalid/blob/5bfe23a8ab3ab49c9050bc3ad0668f5c01492cf6/lib/global_id/locator.rb#L17

`gid.model_class` gets called even if you have a custom locator and will cause an exception if that constant doesn't exist.

adrianthedev:
yes. I have a class `Product`, but it's different for `Order`. It's namespaced. It started to fail on my end as well. Will update in a sec.

adrianthedev:
the quickest quickfix if to have dummy (empty) models for everything you expect to receive in webhooks until this is fixed properly from shopify.

swalkinshaw:
[...]

PR #771, opened 2019-06-25
Require rails > 5.2.1

chrisbutcher:
Follow-up to https://github.com/Shopify/shopify_app/issues/600
Closes https://github.com/Shopify/shopify/issues/205174

See the CHANGELOG for details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+9.1.0
+-----
+
+* Bump required Rails version to `> 5.2.1` to ensure `5.2.1.1` or greater is used. This ensures two things:
+  * Apps are not vulnerable to [CVE-2018-16476](https://nvd.nist.gov/vuln/detail/CVE-2018-16476)
+  * Webhook payloads, from Shopify for API version 2019-07, which are processed in ActiveJob background jobs (the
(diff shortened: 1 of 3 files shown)


nwtn:
We can’t bump to 9.1 because (I believe) there are breaking changes on master that haven’t been released. If this can wait, we can merge this to master without the version bump; if you need us to do a release right away we’ll have to go to 10.0

chrisbutcher:
I can do the changelog / version.rb changes in a separate commit per https://github.com/Shopify/shopify_api/blob/master/RELEASING if you prefer.

chrisbutcher:
> If this can wait

Thanks Dave! This needs to ship before the 2019-07 API version stabilizes on July 1st, so I think it depends on if we think those other changes will make it in sooner or later.

nwtn:
@chrisbutcher ya, if you could do a separate PR for the 10.0 release, we can merge this now and make sure the release goes out this week.

chrisbutcher:
Thanks @nwtn ! To be clear, I'll merge only the functional changes to `shopify_app.gemspec` here in this PR, and open 2nd PR to to bump to 10.0 in version.rb and start updating the Changelog.

If that sounds good, can I have a 👍 ?

PR #772, opened 2019-06-25
Packaging for release 10.0.0

chrisbutcher:
Closes https://github.com/Shopify/shopify_app/pull/772

Follow-up to https://github.com/Shopify/shopify_app/pull/771

Unblocks https://github.com/Shopify/shopify/pull/203904

👋 @nwtn Are all other undeployed changes in master, now targeting 10.0.0 ready for release? Any other release notes to add?

### DIFF
--- a/lib/shopify_app/version.rb
+++ b/lib/shopify_app/version.rb
@@ -1,3 +1,3 @@
 module ShopifyApp
-  VERSION = '9.0.4'.freeze
+  VERSION = '10.0.0'.freeze
 end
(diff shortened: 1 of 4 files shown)


nwtn:
I believe everything else is in master, but we’ll need to add more release notes. I’ll push a commit to this PR to include those.

lib/shopify_app/version.rb
```diff
@@ -1,3 +1,3 @@
 module ShopifyApp
-  VERSION = '9.0.4'.freeze
+  VERSION = '10.0.0'.freeze
```
nwtn: can you make sure this is changed in package.json too, per the [release docs](https://github.com/Shopify/shopify_app/blob/master/docs/Releasing.md)

PR #798, opened 2019-10-11
Upgrade version of omniauth-shopify-oauth2 gem

theundeadmonk:
The latest version of the `omniauth-shopify-oauth2` gem offers support for user_sessions being serialized.

This PR bumps the version of the gem.


ashchristopher:
🎆

PR #802, opened 2019-10-15
adds config host for ngrok url

katiedavis:
Adds config.hosts [required by Rails 6](https://www.fngtps.com/2019/rails6-blocked-host/).
[Fixes issue brought up in Shopify-App-CLI](https://github.com/Shopify/shopify-app-cli/issues/323)

### DIFF
--- a/lib/generators/shopify_app/install/install_generator.rb
+++ b/lib/generators/shopify_app/install/install_generator.rb
@@ -57,6 +57,14 @@ module ShopifyApp
         route "mount ShopifyApp::Engine, at: '/'"
       end
 
+      def insert_hosts_into_development_config
+        inject_into_file(
+          'config/environments/development.rb',
+          "  config.hosts << /\\h+.ngrok.io/\n",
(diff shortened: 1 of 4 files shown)


lib/generators/shopify_app/install/install_generator.rb
```diff
@@ -57,6 +57,14 @@ def mount_engine
         route "mount ShopifyApp::Engine, at: '/'"
       end
 
+      def insert_hosts_into_development_config
+        inject_into_file(
+          'config/environments/development.rb',
+          "  config.hosts << /\\h+.ngrok.io/\n",
```
tanema: We can insert a regex into hosts? I did not know this!

katiedavis:
> Although the regex is a cool feature, I think we should put in something like `ENV['APP_HOST']` so that we can specify it with the `.env` file and so that when they deploy they can just change their env file.

After discussion with @tanema we've decided to keep the regex because not all users will have a host listed in their `env` file and this config is only used in development.

PR #808, opened 2019-10-29
Rails 6 assets compatibility

tanema:
### Rails 5
- JS assets stay where they are with assets manifests still existing so that old javascript_include_tags will still work
- javascript_include_tags will still work with an added rails6 check

### Rails 6
- Javascript files are copied to `app/javascript/shopify_app`
- An index.js file is created to include what we need in the application.js pack with `required('shopify_app')`

### TODO still
- ShopifyApp::Engine precompile
- ~Check for webpacker~
- ~Add an attribute to force Sprockets~

### DIFF
--- a/shopify_app.gemspec
+++ b/shopify_app.gemspec
@@ -14,6 +14,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('rails', '> 5.2.1')
   s.add_runtime_dependency('shopify_api', '~> 8.0')
   s.add_runtime_dependency('omniauth-shopify-oauth2', '~> 2.2.0')
+  s.add_runtime_dependency('dotenv-rails', '~> 2.7.5')
 
   s.add_development_dependency('rake')
   s.add_development_dependency('byebug')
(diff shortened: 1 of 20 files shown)


tanema:
Okay so it seems like I can no longer require JS from a gem and that it needs to come from npm. This seems really silly for the tiny amount of JS that this gem needs. I think I will end up copying all JS to the assets directory (within a shopify_app directory)

tanema:
I finally got a rails 6 generated app working and installed on generated

ziaulrehman40:
I used these changes manually(and partially) and it worked, please release this compatibility as soon as possible. Thanks.

shopify_app.gemspec
```diff
@@ -14,6 +14,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('rails', '> 5.2.1')
   s.add_runtime_dependency('shopify_api', '~> 8.0')
   s.add_runtime_dependency('omniauth-shopify-oauth2', '~> 2.2.0')
+  s.add_runtime_dependency('dotenv-rails', '~> 2.7.5')
```
smathieu: Why was this added a runtime dependency? This forces all apps to add dotenv, which not really needed.

PR #820, opened 2019-11-15
App Extensions Generators 

kipply:
This pr adds generators for building app extensions (though only endpoints at the moment, no related graphql queries or jobs). 

Specifically, this PR currently supports Marketing Activities, but is (hopfully) designed to support other extensions such as Shopify POS and Kit Skills (more info at https://help.shopify.com/en/api/embedded-apps/app-extensions)

I'll put on some comments on other design decisions that were made, though some general goals were; 
 - Make sure the setup works for other app extensions 
 - Things that are optional to a minimal usage flow do not exist

### DIFF
--- a/lib/generators/shopify_app/add_marketing_activity_extension/add_marketing_activity_extension_generator.rb
+++ b/lib/generators/shopify_app/add_marketing_activity_extension/add_marketing_activity_extension_generator.rb
@@ -1,0 +1,39 @@
+require 'rails/generators/base'
+
+module ShopifyApp
+  module Generators
+    class AddMarketingActivityExtensionGenerator < Rails::Generators::Base
+      source_root File.expand_path('../templates', __FILE__)
+
+      def generate_app_extension
+        template "marketing_activities_controller.rb", "app/controllers/marketing_activities_controller.rb"
+        generate_routes
+      end
+
+      private
+
+      def generate_routes
+        inject_into_file(
+          'config/routes.rb',
+          optimize_indentation(routes, 2),
+          after: "root :to => 'home#index'\n"
+        )
+      end
+
+      def routes
+        <<~EOS
+
+          resource :marketing_activities, only: [:create, :update] do
+            patch :resume
+            patch :pause
+            patch :delete
+            post :republish
+            post :preload_form_data
+            post :preview
+            post :errors
+          end
+        EOS
+      end
+    end
+  end
+end
--- a/lib/generators/shopify_app/add_marketing_activity_extension/templates/marketing_activities_controller.rb
+++ b/lib/generators/shopify_app/add_marketing_activity_extension/templates/marketing_activities_controller.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+
+class MarketingActivitiesController < ExtensionVerificationController
+  def preload_form_data
+    preload_data = {
+      "form_data": {
+        "budget": {
+          "currency": "USD",
+        }
+      }
+    }
+    render(json: preload_data, status: :ok)
+  end
+
+  def update
+    render(json: {}, status: :accepted)
+  end
+
+  def pause
+    render(json: {}, status: :accepted)
+  end
+
+  def resume
+    render(json: {}, status: :accepted)
+  end
+
+  def delete
+    render(json: {}, status: :accepted)
+  end
+
+  def preview
+    placeholder_img = "https://cdn.shopify.com/s/files/1/0533/2089/files/placeholder-images-image_small.png"
+    preview_response = {
+      "desktop": {
+        "preview_url": placeholder_img,
+        "content_type": "text/html",
+        "width": 360,
+        "height": 200
+      },
+      "mobile": {
+        "preview_url": placeholder_img,
+        "content_type": "text/html",
+        "width": 360,
+        "height": 200
+      }
+    }
+    render(json: preview_response, status: :ok)
+  end
+
+  def create
+    render(json: {}, status: :ok)
+  end
+
+  def republish
+    render(json: {}, status: :accepted)
+  end
+
+  def errors
+    request_id = params[:request_id]
+    message = params[:message]
+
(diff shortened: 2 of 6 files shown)


lib/generators/shopify_app/hmac_controller/templates/hmac_verification_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  private
+
+  def verify_request
+    hmac_header = request.headers['HTTP_X_SHOPIFY_HMAC_SHA256']
+    request_body = request.body.read
+    secret = ShopifyApp.configuration.secret
+    digest = OpenSSL::Digest.new('sha256')
+
+    calculated_hmac = Base64.strict_encode64(OpenSSL::HMAC.digest(digest, secret, request_body))
```
atesgoral: Can you rename this to `expected_hmac`?

lib/generators/shopify_app/hmac_controller/templates/hmac_verification_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    head(:unauthorized) unless calculated_hmac == hmac_header
```
EiNSTeiN-: use time-constant comparison please
```suggestion
    head(:unauthorized) unless ActiveSupport::SecurityUtils.secure_compare(calculated_hmac, hmac_header)
```

lib/generators/shopify_app/add_extension/add_extension_generator.rb
```diff
@@ -0,0 +1,65 @@
...
+
+module ShopifyApp
+  module Generators
+    class AddExtensionGenerator < Rails::Generators::Base
+      source_root File.expand_path('../templates', __FILE__)
+
+      class_option :type, type: :string, aliases: "-t", required: true
+
+      def generate_app_extension
+        return unless extension_type_valid?
```
kipply: The design of this generator in general is a tad design-breaking since really "add_extension" translates to "add_some_controllers_and_jobs_and_maybe_some_graphql_stuff" and currently every controller generator gets it's own folder. 

I chose not to follow that since the hmac generation and routes injection code will stay the same across extensions and probably some more code to follow. 

I'm also ok with just naming this `add_marketing_activities_extension` and not handling any cases of multiple extensions until we get there.

lib/generators/shopify_app/add_extension/templates/marketing_activities_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    render(json: {}, status: :accepted)
+  end
+
+  def errors
+    request_id = params[:request_id]
+    message = params[:message]
+
+    Rails.logger.info("[Marketing Activity App Error Feedback] Request id: #{request_id}, message: #{message}")
+
+    render(json: {}, status: :ok)
```
kipply: having all this handling is not necessary for any kind of usage flow, but should be included as it's how app developers will be able to debug their responses.

lib/generators/shopify_app/add_extension/add_extension_generator.rb
kipply: gonna go with this^

tanema: renaming would be good as we can just make a parent coordinator task that calls many children tasks.

kipply: Would the way it is now be fine?

lib/generators/shopify_app/add_marketing_activity_extension/add_marketing_activity_extension_generator.rb
```diff
@@ -0,0 +1,40 @@
...
+        inject_into_file(
+          'config/routes.rb',
+          optimize_indentation(routes, 2),
+          after: "root :to => 'home#index'\n"
+        )
+      end
+
+      def routes
+        <<~EOS
+          \nresource :marketing_activities, only: [:create, :update] do
```
tanema: why do you have a newline at the beginning of this rather than just putting a blank line in  your template?

lib/generators/shopify_app/add_marketing_activity_extension/templates/marketing_activities_controller.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+class MarketingActivitiesController < HmacVerificationController
```
tanema: Should this template be commented on what each action does so that the developer can have contextual information?

lib/generators/shopify_app/hmac_controller/templates/hmac_verification_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class HmacVerificationController < ApplicationController
```
tanema: Can this controller just be part of the Gem rather than generated? It doesn't seem like something that the user customizes, just something that the use extends. If we are clever about this as well it can be reused for our webhook controller as well.

lib/generators/shopify_app/add_marketing_activity_extension/templates/marketing_activities_controller.rb
kipply: Decided against, since we don't want to replace / duplicate documentation. Endpoints in this controller should make a fair bit of sense to developers who have for example, tried using a marketing activity. 

Developers should be referencing https://help.shopify.com/en/api/embedded-apps/app-extensions/shopify-admin/marketing-activities/endpoints 

Let me know if you want more context / if you think that some commenting should be included since you understand the philosophy for these tools more!

lib/generators/shopify_app/hmac_controller/templates/hmac_verification_controller.rb
kipply: I hid it from the namespace so that we can call it like a generator from other generators, though it won't come up in help and whatnot.

tanema:
It looks like you need to rebase to resolve your conflicts

PR #822, opened 2019-11-18
FIX: Adding ngrok to config.hosts in environment/development breaks i…

rcapozzi-shopify:
A recent commit added an ngrok entry to `Rails.application.config.hosts`. This seems to work in Rails 6, but it breaks in Rails 5. I understand the argument for adding ngrok, but there are other proxy services one could use. None of those alternatives are included. A security minded person might argue that this gem should not be doing such an update at all, but rather it should be an exercise left to the user.

The video referenced here:  https://help.shopify.com/en/api/tutorials/build-a-shopify-app-with-ruby-and-sinatra, must be using Rails 5.x.

### DIFF
--- a/lib/generators/shopify_app/install/install_generator.rb
+++ b/lib/generators/shopify_app/install/install_generator.rb
@@ -67,7 +67,7 @@ module ShopifyApp
       def insert_hosts_into_development_config
         inject_into_file(
           'config/environments/development.rb',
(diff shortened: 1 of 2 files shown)


tanema:
Your fix looks good but it seems like you need to update the tests for your new condition

rcapozzi-shopify:
I'm canning this request and making another one with the fixed test.

PR #835, opened 2019-12-09
Remove dotenv from dependencies

smathieu:
Nothing in the gem uses this dependency. It should be up to each app how they manage their own environment variables.


smathieu:
It's not backward compatible, therefore needs a minor version bump. e.g. If an app uses dotenv, but does not explicitly have it in its gemfile, it would break it. (unlikely, but still)

Issue #812, opened 2019-10-31
Rails 6.0.0 deprecation for the Shop model initializer


resistorsoftware:
The deprecation notice is long, and the recommended reading on the why is not too enlightening. Is there a quick fix for this? From the initializer shopify_app.rb

     config.session_repository = Shop

DEPRECATION WARNING: Initialization autoloaded the constant Shop.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload Shop, for example,
the expected changes won't be reflected in that stale Class object.

This autoloaded constant has been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
 (called from <main> at /Users/dlazar/Documents/Workspace/Apps/shinesty/config/environment.rb:5)
DEPRECATION WARNING: Initialization autoloaded the constant Shop.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload Shop, for example,
the expected changes won't be reflected in that stale Class object.

This autoloaded constant has been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.

tanema:
Yeah if you want to get rid of the deprecation notice, the best thing to do would be to require your shop model file directly so that you load it before that line in your config

resistorsoftware:
Sorry I don't follow. What does it mean to require a model before an initializer references it? 

I come from Sinatra world, so the quick and dirty isms of Rails are not something I grok as much as you do.

resistorsoftware:
Is that as simple as just saying 

     require 'shop'

in the initializer?

tanema:
Yeah what I was thinking `require 'shop'` however that might end up autoloading other things, so it might not get rid of your warning.

I will make a note to possibly run the config block after app initialization so that this no longer happen.

resistorsoftware:
Super! It seems like I have no patience for some of these matters. I read and re-read that document 10x and still came away with no real plan as to how to ensure my initializers are clean and re-loadable.

axlekb:
This appears to be a duplicate of https://github.com/Shopify/shopify_app/issues/757

resistorsoftware:
I would be happy to see it condensed down to one closed issue... no one in the crowd seems to enamoured to just fix it. Me included.

[...]

Issue #757, opened 2019-05-30
Rails 6.0.0.rc1 deprecation for initialization autoloading of session_repository


axlekb:
Hi folks,

Running `rails 6.0.0.rc1` and noticed a deprecation with `shopify_app.rb` initializer.

This appears to be the offending line which is created by `rails generate shopify_app:shop_model`
```
  config.session_repository = Shop
```

```
DEPRECATION WARNING: Initialization autoloaded the constant Shop.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload Shop, for example,
the expected changes won't be reflected in that stale Class object.

This autoloaded constant has been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
 (called from <top (required)> at <...>/config/environment.rb:7)
```

nwtn:
thanks for the heads-up!

shawnfrench:
A workaround for now is to remove code calling `session_repository=` and instead set the session repository storage class directly within a `to_prepare` callback (avoids autoloading the constant, but runs once at application startup and every time the code is reloaded):
```
ShopifyApp.configure do |config|
  ...
end

Rails.application.config.to_prepare do
  ShopifyApp::SessionRepository.storage = ShopifyApp::InMemorySessionStore
end
```

axlekb:
So I can confirm this appears to work when the `Rails.configuration.cache_classes` is `true`.  So it works on my production environment, however I get errors in my test and development environment that don't cache classes.

```
#<ShopifyApp::SessionRepository::ConfigurationError: ShopifySessionRepository.storage is not configured!>
/.../gems/shopify_app-11.4.0/lib/shopify_app/session/session_repository.rb:23:in `storage'
/.../gems/shopify_app-11.4.0/lib/shopify_app/session/session_repository.rb:15:in `retrieve'
/.../gems/shopify_app-11.4.0/lib/shopify_app/controller_concerns/login_protection.rb:31:in `shop_session'
/.../gems/shopify_app-11.4.0/lib/shopify_app/controller_concerns/login_protection.rb:35:in `login_again_if_different_shop'
```

```ruby
# lib/shopify_app/configuration.rb
    def session_repository=(klass)
      if Rails.configuration.cache_classes
        ShopifyApp::SessionRepository.storage = klass
      else
        ActiveSupport::Reloader.to_prepare do
[...]

PR #841, opened 2019-12-18
Use to_prepare to prevent autoload of constants in initializers

shawnfrench:
As reported in https://github.com/Shopify/shopify_app/issues/757 and https://github.com/Shopify/shopify_app/issues/812, currently our generated initializer running with rails 6 will give this warning:

```
DEPRECATION WARNING: Initialization autoloaded the constant Shop.

Being able to do this is deprecated. Autoloading during initialization is going

Reloading does not reboot the application, and therefore code executed during

This autoloaded constant has been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
```

As [documented](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html) we cannot load constants in initializers, and as detailed above, this is currently a problem with config like `ShopifyApp` which allows/expects a class constant to be set for `session_repository` like:
```
  ...
  config.session_repository = Shop
end
```

The easiest way to fix/avoid this is to wrap `ShopifyApp.configure` blocks in a `Rails.application.config.to_prepare` block. 

From:
```
  ...
end
```

To:
```
Rails.application.config.to_prepare do
    ...
  end
end
```

**IMPORTANT**: This PR fixes our generated code, but for those with an existing rails 6 + `shopify_app` app initializer, they will need to make this change manually in their initializer to remove the rails 6 warning.

**NOTE**: If you have other initializers referencing the `ShopifyApp.config` that will not work (eg. omniauth) and you should instead pull those shared values from `ENV` or some shared config/constant.

Closes https://github.com/Shopify/shopify_app/issues/757
Closes https://github.com/Shopify/shopify_app/issues/812

### DIFF
--- a/lib/generators/shopify_app/install/templates/shopify_app.rb
+++ b/lib/generators/shopify_app/install/templates/shopify_app.rb
@@ -1,15 +1,17 @@
-ShopifyApp.configure do |config|
-  config.application_name = "<%= @application_name %>"
-  config.api_key = ENV['SHOPIFY_API_KEY']
-  config.secret = ENV['SHOPIFY_API_SECRET']
-  config.old_secret = "<%= @old_secret %>"
-  config.scope = "<%= @scope %>" # Consult this page for more scope options:
-                                 # https://help.shopify.com/en/api/getting-started/authentication/oauth/scopes
-  config.embedded_app = <%= embedded_app? %>
-  config.after_authenticate_job = false
-  config.api_version = "<%= @api_version %>"
-  config.session_repository = ShopifyApp::InMemorySessionStore
-end
+Rails.application.config.to_prepare do
+  ShopifyApp.configure do |config|
+    config.application_name = "<%= @application_name %>"
+    config.api_key = ENV['SHOPIFY_API_KEY']
+    config.secret = ENV['SHOPIFY_API_SECRET']
+    config.old_secret = "<%= @old_secret %>"
+    config.scope = "<%= @scope %>" # Consult this page for more scope options:
+                                   # https://help.shopify.com/en/api/getting-started/authentication/oauth/scopes
+    config.embedded_app = <%= embedded_app? %>
+    config.after_authenticate_job = false
+    config.api_version = "<%= @api_version %>"
+    config.session_repository = ShopifyApp::InMemorySessionStore
+  end
 
-# ShopifyApp::Utils.fetch_known_api_versions                        # Uncomment to fetch known api versions from shopify servers on boot
-# ShopifyAPI::ApiVersion.version_lookup_mode = :raise_on_unknown    # Uncomment to raise an error if attempting to use an api version that was not previously known
+  # ShopifyApp::Utils.fetch_known_api_versions                        # Uncomment to fetch known api versions from shopify servers on boot
+  # ShopifyAPI::ApiVersion.version_lookup_mode = :raise_on_unknown    # Uncomment to raise an error if attempting to use an api version that was not previously known
+end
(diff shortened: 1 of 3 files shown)


shawnfrench:
Scratch that @tanema - there is a problem I just found while testing. 

Our `session_repository=` method already (conditionally) uses the `ActiveSupport::Reloader.to_prepare` which means the block is basically queued in a nest, and never actually runs.
https://github.com/Shopify/shopify_app/blob/87836e490b86da66e36150fe9c8806d9b2700b97/lib/shopify_app/configuration.rb#L49-L57

We can either:
1) remove the `to_prepare` logic from our method and have people wrap the entire `ShopifyApp.configure` block as this PR suggests. Those relying on the cache_classes/preparation code might run into loading issues.

2) perhaps a better approach is to remove this odd setter (it doesn't even set the attribute...) and deprecate then remove `session_repository=` and have folks set `ShopifyApp::SessionRepository.storage` directly in their initializer, wrapping that code in a `.to_prepare` themselves. 

If we go with 2) our `shopify_app` initializer would then look like:
```
  ...
end

Rails.application.config.to_prepare do
  ShopifyApp::SessionRepository.storage = ShopifyApp::InMemorySessionStore
end
```

Thoughts or better ideas @Shopify/platform-dev-tools-education?

tanema:
How about a 3rd option

We remove the to_prepare block in the session_repository method and add it to the Shopify.configure method.

```
  def self.configure
    ActiveSupport::Reloader.to_prepare do
      yield configuration
    end
  end
```

This way it will automatically fix any existing code and operate how your current code works. It is backward compatible and gets rid of the deprecation warning.

shawnfrench:
Yup that works as long as people aren't already wrapping in `to_prepare` :+1: Not sure what the best practice is here... but I'm good with option 3.

tanema:
You said if it is already wrapped in a to_prepare that the block never runs right? I wonder if there is a way to check if we are already within a to_prepare block and throw and error.  If not, we will just update the version number to indicate a breaking change.

tanema:
I took a look around and it seems like ActiveSupport::Reloader uses ActiveSupport::Callbacks and it does not seem like there is any functionality to check if a callback is in progress https://github.com/rails/rails/blob/master/activesupport/lib/active_support/callbacks.rb

I think the 3rd option would be best still, we can 

- change the version number. 
- Document it in the changelog 
- Check if the session store is undefined after startup and throw an error adding a note that "you might have nested your config in a to_prepare block"
- *maybe* add a comment to the initializer template saying that the configure block already handles to_prepare

rafaelfranca:
Let me try to understand, why that block needs to run in a `to_prepare` now? This means that it will now run once on every request in development now, but it used to be executed only once before.

I can see why `session_repository=` needed to do that.

Maybe what you want is `after_initialize`?

There is also another option. Just remove the constant from the config and instead accept a string with the name of the constant. That would remove the need for the `to_prepare` and `afeter_initialize` all together.

tanema:
Setting it as a string is a good possibility I just worry about validating it. We cannot validate that the class is available and has the correct interface without loading the model. Although I suppose we could do that validation in a to_prepare block in a similar way we are doing right now with Rails.configuration.cache_classes

shawnfrench:
Hmmm that's a great point @rafaelfranca - we already do [accept a string](https://github.com/Shopify/shopify_app/blob/master/lib/shopify_app/session/session_repository.rb#L30) and validate the class shape in `ShopifyApp::SessionRepository.storage=` so we can just remove the logic in the `configuration.session_repository=` method and let the storage class deal with the string. PR'd here: https://github.com/Shopify/shopify_app/pull/848

shawnfrench:
Replaced by https://github.com/Shopify/shopify_app/pull/848

PR #848, opened 2020-01-07
Use a string when setting session_repository

shawnfrench:
As reported in https://github.com/Shopify/shopify_app/issues/757 and https://github.com/Shopify/shopify_app/issues/812, currently our generated initializer running with rails 6 will give this warning:

```
DEPRECATION WARNING: Initialization autoloaded the constant Shop.

Being able to do this is deprecated. Autoloading during initialization is going

Reloading does not reboot the application, and therefore code executed during

This autoloaded constant has been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
```

We can avoid issues with the autoloader all-together by just using a string to configure the `session_repository` (and `ShopifyApp::SessionRepository.storage`). 

PR also fixes the setter to actually store the value in `@session_repository`.

These changes are backwards compatible, but anyone using Rails 6 should change their initializer to configure `session_repository` with a string to avoid the deprecation and eventual load error. Replaces https://github.com/Shopify/shopify_app/pull/841.

### DIFF
--- a/lib/shopify_app/configuration.rb
+++ b/lib/shopify_app/configuration.rb
@@ -50,13 +50,8 @@ module ShopifyApp
     end
 
     def session_repository=(klass)
-      if Rails.configuration.cache_classes
-        ShopifyApp::SessionRepository.storage = klass
-      else
-        ActiveSupport::Reloader.to_prepare do
-          ShopifyApp::SessionRepository.storage = klass
-        end
-      end
+      @session_repository = klass
+      ShopifyApp::SessionRepository.storage = klass
     end
 
     def has_webhooks?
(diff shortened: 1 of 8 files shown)


lib/shopify_app/configuration.rb
```diff
@@ -50,13 +50,8 @@ def login_url
...
 
     def session_repository=(klass)
-      if Rails.configuration.cache_classes
-        ShopifyApp::SessionRepository.storage = klass
-      else
-        ActiveSupport::Reloader.to_prepare do
-          ShopifyApp::SessionRepository.storage = klass
-        end
-      end
+      @session_repository = klass
```
rafaelfranca: `attr_accessor` should be changed to `attr_reader` it is probably raising a warning as well already because you the method is being redefined.

PR #849, opened 2020-01-08
Revert per user token

slucaskim:
Revert https://github.com/Shopify/shopify_app/pull/760/

We found a problem with the session login when we deployed our app with 11.5.0, a problem that was not encountered locally when testing. Rolling this change back as a patch 11.5.1


isaacbowen:
I'm experimenting with this right now - what was the issue you ran into?

tanema:
We are running some tests on one of our internal apps and it was throwing errors in production rather than just in development. We will definitely be adding back this functionality once we have smoothed out the bumps :D

slucaskim:
The issue that triggered this revert was unrelated to the changes in this original PR, but rather another change that was not reverted. Details in this [fix PR](https://github.com/Shopify/shopify_app/pull/855)

PR #851, opened 2020-01-11
Add SameSite=None and secure to App Session

jenshenny:
### Background

In Chrome 80, the default value of the `SameSite` cookie property will switch from `None` to `Lax` ([more details here](https://docs.google.com/document/d/1d2OTcEHQaRvHaJumPuKyu0pVPre_p_GU5hp_S-Ul_7w/edit)). This means that all iframe'd and cross-origin non-GET requests will not send cookies, meaning that it will break  embedded app functionality (you can try enabling the [SameSite None flag](chrome://flags/#same-site-by-default-cookies) and accessing an embedded app in your admin)

Additionally, `SameSite=None` (which we do not use today) is unfortunately interpreted incorrectly by [specific browsers](https://www.chromium.org/updates/same-site/incompatible-clients), so this PR handles such situations as well.

### Changes
Added a middleware that sets all app sessions to `SameSite=None` and `secure` if the browser is compatible and the app is an embedded app.

Decided to add a middleware as setting the session option to `same_site: :none` isn't applied to set-cookie headers for `shopify_app` assets and `/auth/shopify` (omniauth)

### Tophatting
Tophatted with a new shopify app (where `_example_session` has secure and SameSite None.

image

 Also tophatted with one of our first party embedded apps Exchange (though I removed the `ShopifyApp.configuration.embedded_app == true` check to test as Exchange runs both as an embedded app and non embedded app). Also tophatted with the user agent `Mozilla/5.0 (iPad; CPU OS 11_3 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/11.0 Tablet/15E148 Safari/604.1`.

If the session already has  `secure` and `SameSite` set, a `secure` and `SameSite` will still be appended to the end of the set-cookie header. I personally don't think it's a huge problem as the browser ignores the previous `secure` and `SameSite` in the header, but would love to get more opinions on it.

### DIFF
--- a/lib/shopify_app/configuration.rb
+++ b/lib/shopify_app/configuration.rb
@@ -34,6 +34,9 @@ module ShopifyApp
     # allow namespacing webhook jobs
     attr_accessor :webhook_jobs_namespace
 
+    # allow enabling of same site none on cookies
+    attr_accessor :enable_same_site_none
+
     def initialize
       @root_url = '/'
       @myshopify_domain = 'myshopify.com'
@@ -57,6 +60,10 @@ module ShopifyApp
 
     def has_scripttags?
       scripttags.present?
+    end
+
+    def enable_same_site_none
+      @enable_same_site_none.nil? ? embedded_app? : @enable_same_site_none
     end
   end
 
--- a/lib/shopify_app/engine.rb
+++ b/lib/shopify_app/engine.rb
@@ -12,5 +12,9 @@ module ShopifyApp
         storage_access.svg
       ]
     end
+
+    initializer "shopify_app.middleware" do |app|
+      app.config.middleware.insert_before(ActionDispatch::Cookies, ShopifyApp::SameSiteCookieMiddleware)
+    end
   end
 end
--- a/lib/shopify_app/middleware/same_site_cookie_middleware.rb
+++ b/lib/shopify_app/middleware/same_site_cookie_middleware.rb
@@ -10,7 +10,7 @@ module ShopifyApp
       user_agent = env['HTTP_USER_AGENT']
 
       if headers && headers['Set-Cookie'] && !SameSiteCookieMiddleware.same_site_none_incompatible?(user_agent) &&
-        ShopifyApp.configuration.embedded_app == true
+          ShopifyApp.configuration.enable_same_site_none
 
         cookies = headers['Set-Cookie'].split("\n").compact
 
--- a/lib/shopify_app/middleware/same_site_cookie_middleware.rb
+++ b/lib/shopify_app/middleware/same_site_cookie_middleware.rb
@@ -1,0 +1,58 @@
+module ShopifyApp
+  class SameSiteCookieMiddleware
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      _status, headers, _body = @app.call(env)
+    ensure
+      user_agent = env['HTTP_USER_AGENT']
+
+      if headers && headers['Set-Cookie'] && !SameSiteCookieMiddleware.same_site_none_incompatible?(user_agent) &&
+        ShopifyApp.configuration.embedded_app == true
+
+        cookies = headers['Set-Cookie'].split("\n").compact
+
+        cookies.each do |cookie|
+          headers['Set-Cookie'] = headers['Set-Cookie'].gsub("#{cookie}", "#{cookie}; secure; SameSite=None")
+        end
+      end
+    end
+
+    def self.same_site_none_incompatible?(user_agent)
+      sniffer = BrowserSniffer.new(user_agent)
+
+      webkit_same_site_bug?(sniffer) || drops_unrecognized_same_site_cookies?(sniffer)
+    rescue
+      true
+    end
+
+    def self.webkit_same_site_bug?(sniffer)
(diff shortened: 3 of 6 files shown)


tanema:
Hey @jenshenny Is this work needed after the additional work being added to Rack and Rails?

jenshenny:
Nope! This work can do without the bump to Rack 2.1.0 and the additional work in Rails. Originally, I was trying to change the session option of `same_site` to `:none` in the app which then would need the bump in Rack, but there were many unusual areas where the `Set-Cookie` header didn't apply the session option change. I figured it would be much simpler to manually change the `Set-Cookie` header myself.

lib/shopify_app/engine.rb
```diff
@@ -12,5 +12,9 @@ class Engine < Rails::Engine
         storage_access.svg
       ]
     end
+
+    initializer "shopify_app.middleware" do |app|
+      app.config.middleware.insert_before(ActionDispatch::Cookies, ShopifyApp::SameSiteCookieMiddleware)
```
rafaelfranca: I'd allow people to configure this. Some apps could be using versions of Rails and rack that already does this by default, so we should not try to do it again.

lib/shopify_app/middleware/same_site_cookie_middleware.rb
```diff
@@ -0,0 +1,58 @@
...
+    ensure
+      user_agent = env['HTTP_USER_AGENT']
+
+      if headers && headers['Set-Cookie'] && !SameSiteCookieMiddleware.same_site_none_incompatible?(user_agent) &&
+        ShopifyApp.configuration.embedded_app == true
+
+        cookies = headers['Set-Cookie'].split("\n").compact
+
+        cookies.each do |cookie|
+          headers['Set-Cookie'] = headers['Set-Cookie'].gsub("#{cookie}", "#{cookie}; secure; SameSite=None")
```
JackMc: It might be good to check here if the existing `cookie` value contains the phrase `; SameSite` as a quick hack to ensure we don't double up or trounce on intentionally secured (i.e. `SameSite=(lax|strict)`) cookies.

lib/shopify_app/middleware/same_site_cookie_middleware.rb
```diff
@@ -0,0 +1,58 @@
...
+      @app = app
+    end
+
+    def call(env)
+      _status, headers, _body = @app.call(env)
+    ensure
+      user_agent = env['HTTP_USER_AGENT']
+
+      if headers && headers['Set-Cookie'] && !SameSiteCookieMiddleware.same_site_none_incompatible?(user_agent) &&
+        ShopifyApp.configuration.embedded_app == true
```
JackMc: From your PR description:

> Also tophatted with one of our first party embedded apps Exchange (though I removed the ShopifyApp.configuration.embedded_app == true check to test as Exchange runs both as an embedded app and non embedded app).

To me, this is a smell that `ShopifyApp.configuration.embedded_app == true` might not be the right configuration option to rely on. Would it make sense to add a configuration option that is on by default if the application is embedded, but can be turned on by developers even if their application is not embedded? This would also allow more configurability for embedded app authors who may want to handle this in their own app.

lib/shopify_app/configuration.rb
```diff
@@ -70,5 +73,8 @@ def self.configuration=(config)
 
   def self.configure
     yield configuration
+    if !ShopifyApp.configuration.embedded_app? || @disable_same_site_cookie_middleware
```
JackMc: This expresses the same thing, I think

```suggestion
    unless ShopifyApp.configuration.embedded_app?
```

JackMc: (As discussed in person) It seems like this logic could be simpler if the option were phrased positively, i.e. `set_same_site_none`. This would also hide that we're using a middleware from our public interface in case we want to change it later (which we probably will).

We could then write it as:

```
if @set_cookies_same_site_none.nil?
   @set_cookies_same_site_none = ShopifyApp.configuration.embedded_app?
```

lib/shopify_app/middleware/same_site_cookie_middleware.rb
jenshenny: Whoops, forgot this comment - just added a check for that

lib/shopify_app/configuration.rb
```diff
@@ -70,5 +73,9 @@ def self.configuration=(config)
 
   def self.configure
     yield configuration
+
+    if configuration.enable_same_site_none.nil?
```
tanema: This should probably just be in an accessor on the instance.

netwire88:
Is this affecting redirects to `confirm_recurring_application_charge` URL? Currently, it works if the redirect is handled in a non-embedded app browser window, but in the embedded app browser window, redirects don't work.

tanema:
@netwire88 Please open an issue.

RichardBlair:
@jenshenny  just wanted to thank you for this PR. It's well written and the tests do a good job of documenting what I needed to do. Our app (Parcelify.co) has been around a while, and as such it was less obvious for me were to hook things up. Your PR pointed me to the solution. 👏

Issue #850, opened 2020-01-10
Investigate / update install instructions to include adding app from partner dashboard.


tiffanytse:
Instructions for adding the app to the store from the partners dashboard, install link doesn't work if this is not done first.

cc: @ShayneP @stevehaase

stevehaase:
When following the tutorial, the final step is to open the URL and enter the myshopify domain to add the app. That process results in a 400 - Oauth error invalid_request page with the message 

> Oauth error invalid_request: Could not find Shopify API application with api_key

This happens whether installing the app via an install link created in the partner dash, or through the login screen from the app's UI.

Jipsica:
@stevehaase A few questions about the app and how you installed it:

> enter the myshopify domain to add the app. 

Was this a custom app, or public app? In either case, did you try to install on your dev shop using the UI or with a url pasted into the browser?

If you tried to paste the url into your browser in either case, you'll see `Oauth error invalid_link: This app can't be installed using this link. Contact the app developer for more information.`

Also, we're not quite sure how you saw the error: `Oauth error invalid_request: Could not find Shopify API application with api_key`. Can you tell us the steps you took to get this error?

oehlschl:
Sorry if this is unrelated, but this issue is the one place I've ever seen the error message we just received, which was the `Oauth error invalid_link` error while working with a partner to install our custom app.

> Was this a custom app, or public app.... If you tried to paste the url into your browser in either case, you'll see Oauth error invalid_link....

Does this mean that custom apps can no longer be installed with OAuth links as described here? https://help.shopify.com/en/api/getting-started/authentication/custom-authentication#create-a-custom-app-and-generate-an-installation-link

Again, sorry if this is unrelated; just trying to get more information on the error we're experiencing.

jessipa:
@oehlschl The help guide you linked is the most up to date way on how custom apps can be installed on stores. 

Note that you must use the link provided when you hit `Generate link`. If you try to install using a link constructed following Step 2 [here](https://help.shopify.com/en/api/getting-started/authentication/oauth#step-2-ask-for-permission), installation will not work.

oehlschl:
Thanks for the prompt response @jessipa! I'll double check our installation steps now.

oehlschl:
Just for completeness, I did confirm that we went through the prescribed "Generate Link" flow but still received the invalid_link error above.

I understand that at some point we'll just need to reach out to Shopify support.

Anyway, thanks again for your answer!

[...]

PR #854, opened 2020-01-15
Quickstart - Update instructions for installing an app

Jipsica:
Addresses https://github.com/Shopify/shopify_app/issues/850

We updated the instructions for step 7 in the quickstart.md file. With the release of public apps, we now restrict the usage of copying and pasting installation urls as a way to install apps.

You must now go through the UI to install these apps on dev stores for testing.

cc @lyninx since we paired on this.


tanema:
@Jipsica I know this is probably not part of your work but why is this? Can we still install the app using the regular OAuth initiation flow? If not, this is a huge detriment to the app-cli. Why was this decision made?

Jipsica:
@tanema We spoke with @tylerball about this so he should have more context on that decision, let me know if you need more info after that.

tanema:
Okay well we need to have more conversation because if we are not telling people to go to the main page of their app, fill in the store name and then initiate authentication that way, then we need to rewrite more than just this readme.

tanema:
Okay I talked to about this and it seems like we allow the normal OAuth flow as long as the store is a non-transferrable store belonging to the same org that owns the app. Developers are not required to click this button to install a development app. Instead of adding instructions that make them go back to their partners dashboard you should leave the instructions to visit their app. It will open the normal form for the user to enter their domain to start the OAuth process. Instead of adding the image I think you should be adding a notice of how restrictive the installation is and a description of the precicse conditions you need for installing an app on a development store.

lyninx:
@tanema the reason we wanted to direct developer to the "install on a development store" UI is because that's the main way for them to make a dev store non-transferable in the first place. i do agree that we can keep the existing flow, but we'll need to be pretty clear about how they can actually acquire a non transferable store.

tanema:
> but we'll need to be pretty clear about how they can actually acquire a non transferable store.

Yup agreed with this exact point.

kevinhughes27:
> belonging to the same org that owns the app 

this might also trip people up. Will there be a specific error for this case? Does this affect applications built with other frameworks or just ShopifyApp?

Jipsica:
Thanks for the great feedback, everyone!

@tanema If I understand your point above, would it be better if we added additional information explaining how the developer can create a transfer-disabled store?

@kevinhughes27 I think this spreadsheet explaining the [different installation cases](https://docs.google.com/spreadsheets/d/1rv5dkahiJDFjRhGL1FdADVqDXzsYfW0KnBsPwhiyyAA/edit#gid=0) might answer your concerns. Let me know if you have questions.

tanema:
@Jipsica Yes, that would be much better yes. All the information about this change would be excellent because I know this will trip people up in the future.

Also I dont think Kevin has access to that spreadsheet because it looks like he doesnt work at Shopify anymore

Jipsica:
I've looped in the app infrastructure team (@shutingc) to make sure I capture the correct information for the store-transfer disabled shops. I will get back to this once I meet with them.

kevinhughes27:
I can actually view that spreadsheet.

It might be worth considering the experience of an app developer working on a new app. It seems pretty possible (and I've done some consulting like this) where I am working on an app created by another partner but using my test store (aka different org). Given these new install policies how am I supposed to figure out why the app is not installing in this case? Is the same org part really necessary as long as the shop is a dev shop?

shutingc:
"Different org installation is not allowed" will be displayed during the app installation. @kevinhughes27 The case you mentioned is quite interesting. Do you mind explain more why app developers with different org need to work together?

RichardBlair:
Just jumping in because I've actually run into the exact use case that @kevinhughes27 has mentioned. Within the app ecosystem it's really common to hire someone and not give them access to the partner dashboard. Even with the permissions in the partner dash (viewing rev, for instance) companies are very hesitant to hand over access. As a result, when hired by one of these app companies you usually signup for your own partner account to create dev stores there. Forcing dev stores to be in the same org as the app they are developing on forces a decision on the partner.

A more appropriate approach might be to restrict apps in development to dev stores. If the store is published and the app is in development, you could expire the api keys instead (or something along these lines). This would allow any partner to install any app that's not published on a dev store.

Jipsica:
Thanks for the feedback. We've just released the first iteration for the changes on installation policies, so we'll definitely consider these improvements for the future (i.e. installing apps on dev stores not within the same organization). 

For this PR though, I think we'll stick with providing clear instructions on how the app developer can successfully install an app for testing purposes, and any implications that could have. For example, what a store-transfer disabled store is and the limitations associated with it.

docs/Quickstart.md
```diff
@@ -72,4 +72,10 @@ $ heroku run rake db:migrate
 
 7. Install the App!
 -------------------
-`https://<name>.herokuapp.com/`
+ Ensure you have a dev store created, if you don't already have one, [create one](https://help.shopify.com/en/api/getting-started/making-your-first-request#create-a-development-store).
```
Jipsica: Expand dev store into development store, and link to an external doc cc @ow

Jipsica:
```suggestion
 Ensure you have a store-transfer disabled shop for this next step. If you don't already have one, [create one](https://help.shopify.com/en/api/getting-started/making-your-first-request#create-a-development-store).
```

docs/Quickstart.md
```diff
@@ -72,4 +72,10 @@ $ heroku run rake db:migrate
...
+
+##### Important Note: Installing the app on a development store in the following step will convert the store to be [transfer-disabled](https://help.shopify.com/en/api/guides/store-transfers#transfer-disabled-stores). This is an irreversible change.
+
+Install the app using your Partners Dashboard in the Apps area.
+
+![Installing an App on a dev store using partners dashboard](/docs/install-on-dev-shop.png)
```
Jipsica: Add the other image (card) on how to test this app

Jipsica:
This is ready for another review. We've clarified the `install the app` step to include more information about transfer-disabled shops. The external helps doc it points too will be improved by the App Infra team (@ow )

ow:
I've just dropped in a few other suggestions and clarifications around the new Shopify Partner app types and store transfers in here to help make it clearer for people starting to build with Shopify for the first time. I'll also look at making some expansions on this in the future!

Jipsica:
@tanema let me know if you have more comments. Thanks!

ow:
Thanks for everyone's help on this! We'll be adding more clarity to these docs shortly, as well. 🚢

Issue #857, opened 2020-01-16
v11.7 generates `ArgumentError (wrong number of arguments (given 2, expected 1)):`


netwire88:
I just updated to v11.7 of the gem, but my existing `Shop` object that implements 
`def self.store(session)` 
is now failing with error `ArgumentError (wrong number of arguments (given 2, expected 1)):`. This wasn't happening with v 11.3.0. Anyone else experiencing this?
I've also tried changing it to ` def self.store(session, *args)` and that didn't work.

I was referring to https://github.com/Shopify/shopify_app/blob/master/lib/shopify_app/session/storage_strategies/shop_storage_strategy.rb

tanema:
Yes sorry, the new signature is `self.store(auth_session, *args)`  We added user authentication along side of shop authentication and we needed to make those changes 

https://github.com/Shopify/shopify_app/pull/856

@slucaskim Can you Please update the README.md to update the session#store method signature as well as a note for the change in the new version please?

slucaskim:
@netwire88 To clarify, are you overriding the `self.store` method in your own `Shop` model?

netwire88:
Thanks, the method worked after changing to `self.store(auth_session, *args)` and restarting the server. However, after I sign into the account and Shopify API is redirecting, I get this in the browser console. Are others experiencing the same issue?

```
Failed to find a valid digest in the 'integrity' attribute for resource 'https://myservername.com/assets/shopify_app/request_storage_access-5a556c57b51c57f00d9d6a2ba007f3172739bab34cc2f647a93e792348e8798d.js' with computed SHA-256 integrity '0lJjndtl8qUdAHc8U0DNkSIqNg4Nx2/QjfsZIycjNoU='. The resource has been blocked.
```

tanema:
That is part of https://github.com/Shopify/shopify_app/issues/814 It seems like there is an issue with Sprockets 4, We suggest that you pin Sprockets to 3.9 until those issues are fixed

tanema:
@slucaskim It sounds like that is what he is doing yes

netwire88:
Thanks

PR #858, opened 2020-01-16
User-token documentation updates

slucaskim:
Addresses https://github.com/Shopify/shopify_app/issues/857


slucaskim:
> This looks good. Should we add anything about using a UserSession repository and receiving the user object with store(session, user: user) ?

I think the overriding use case is not that common, so I don't think it's necessary?

tanema:
@slucaskim It is a lot more common than you would think. Please checkout the session_storage model in the BuyButton repo to get an idea of common usage

PR #863, opened 2020-01-17
Small fix for samesite cookie middleware

tanema:
This was causing Oauth to no longer work on shopify_app.


tanema:
Hey all I am going to play around with this some more. It doesnt make sense why this fixes my OAuth issues since the gsub should not be removing the newlines.

I can confirm that:
- Removing the middleware fixes my oauth
- Adding the newline fixes my oauth
- Also just setting the whole header fixes it. I am not sure the actual root cause and I will dive into today

tanema:
Honestly, I feel crazy. It was all broken on Friday and now with multiple tests, the OAuth process works just fine. I will close this now then 🤷‍♂

PR #871, opened 2020-01-21
migrate client scripts to App Bridge

### LINKED ISSUE #723: Update shopify_app gem to use App Bridge actions
[Original issue](https://github.com/Shopify/app-bridge/issues/692) by @nwtn:

**Is your feature request related to a problem? Please describe.**
Right now shopify_app is using EASDK methods, and in its OAuth thingy it’s actually directly sending a postMessage

**Describe the solution you'd like**
In general, switching to App Bridge best actions and best practices around those

The OAuth thing is interesting. I think the reason we’re doing a raw postMessage is for perf — having to download the whole client just to dispatch a Redirect is pretty huge overhead, especially on a page we want to immediately get the user away from. I know doing postMessage directly is no bueno, so maybe we need to find a less fragile but still performant way to do this. 

**Describe alternatives you've considered**
Leaving it alone? It still works as is, because the postMessage signature is still valid for legacy bridge, so everything still works.

**Additional context**
I’m trying to write up instructions/docs for creating 1P embedded apps and I’d like to get this resolved within ~1 week to include in this guide.

tylerball:
This PR converts all generated JS to use App Bridge.

## Questions
- I'm using an unversioned unpkg script tag to load in the library. The only used actions are `TitleBar` and `Toast` create and I don't anticipate the types for these actions changing a lot, but I can switch to a versioned import if we feel that's better.

cc @Shopify/app-bridge 

closes #723

### DIFF
--- a/lib/generators/shopify_app/install/templates/flash_messages.js
+++ b/lib/generators/shopify_app/install/templates/flash_messages.js
@@ -4,12 +4,21 @@ if (!document.documentElement.hasAttribute("data-turbolinks-preview")) {
   document.addEventListener(eventName, function flash() {
     var flashData = JSON.parse(document.getElementById('shopify-app-flash').dataset.flash);
 
+    var Toast = window['app-bridge'].actions.Toast;
+
     if (flashData.notice) {
-      ShopifyApp.flashNotice(flashData.notice);
+      Toast.create(app, {
+        message: flashData.notice,
+        duration: 5000,
+      }).dispatch(Toast.Action.SHOW);
     }
 
     if (flashData.error) {
-      ShopifyApp.flashError(flashData.error);
+      Toast.create(app, {
+        message: flashData.error,
+        duration: 5000,
+        isError: true,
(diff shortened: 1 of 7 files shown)


lib/generators/shopify_app/install/templates/flash_messages.js
```diff
@@ -4,12 +4,21 @@ if (!document.documentElement.hasAttribute("data-turbolinks-preview")) {
   document.addEventListener(eventName, function flash() {
     var flashData = JSON.parse(document.getElementById('shopify-app-flash').dataset.flash);
```
vividviolet: Can we remove all instances of `flash` and rename to `toast` for consistency?

tylerball: Yup, I'll update where it makes sense. Some of these will stay named `flash` just because that's the idiom [Rails uses](https://api.rubyonrails.org/classes/ActionDispatch/Flash.html)

PR #873, opened 2020-01-24
Fix return_to address in Safari after enable_cookies/granted_storage_access

joshreeves:
When a user requests a deep link within the app, and is using a recent version of Safari, the `return_to` address would be lost in the chain of redirects happening for the `enable_cookies` and `granted_storage_access` steps. This PR solves the issue by converting `session[:return_to]` to `params[:return_to]` (where necessary), and ensuring that JS redirects respect the presence of `params[:return_to]` when present.

@tylerball @tanema

### DIFF
--- a/app/assets/javascripts/shopify_app/storage_access.js
+++ b/app/assets/javascripts/shopify_app/storage_access.js
@@ -28,8 +28,8 @@
     window.parent.location.href = this.redirectData.myshopifyUrl + '/admin/apps';
   }
 
-  StorageAccessHelper.prototype.redirectToAppHome = function() {
-    window.location.href = this.redirectData.appHomeUrl;
+  StorageAccessHelper.prototype.redirectToAppTargetUrl = function() {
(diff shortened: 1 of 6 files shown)


joshreeves:
Hey @tanema, thanks for the feedback. I'm not sure what you're referring to?

tanema:
Sorry 😓 too much lingo. I meant that you fully tested this with your browser, with safari, chrome, FF, ect. I will do this as well but it is good to know

joshreeves:
@tanema - Ah, no worries :) Yep! Tested in Chrome 79, Safari 13, Firefox 72, Mobile Safari 12 (using Shopify Admin), and Mobile Safari 12 Native Webview (within Shopify POS). All working as expected.

tanema:
I am just getting people to test this out a bit here and then I hope to merge it soon. Thanks again for your work!

tanema:
We tested it out and it looks great. This will be in the next version release

joshreeves:
Awesome, thanks @tanema !

Issue #883, opened 2020-02-05
Per User tokens being used for Webhooks Callback Controller


tanema:
Currently, after the release of the per_user_authentication, there is an issue with callbacks/webhooks where the user token is currently used. This will work for a time but the user tokens have a short expiry.

To fix this we need to 
- Modify our installation process to get both tokens
- Store both shop and user tokens
- Look up shop token in callbacks controller using shop domain
- [Maybe] Add an `app/uninstall` webhook required so that we can track shop session to require re-auth
- [Maybe] Add a way to check if shop token is still valid so that we can skip shop auth, so that shop auth is not repeated every time the app is used

@slucaskim

slucaskim:
For knowing if user token or shop token came back from the omniauth callback, we can check for presence of [`associated_user`](https://github.com/Shopify/shopify_app/blob/master/app/controllers/shopify_app/callback_controller.rb#L37) because that would only be present in case of user token strategy

Issue #569, opened 2018-03-30
Per-user authentication: Pass associated_user Auth Info to ShopifyAPI::Session


sshaw:
[`associated_user`](https://github.com/Shopify/shopify_app/blob/76c23425b9dcb1eb0cd6918fde1ef32bb05310e5/app/controllers/shopify_app/sessions_controller.rb#L57) should be [passed to `ShopifyAPI::Session.new`](https://github.com/Shopify/shopify_app/blob/76c23425b9dcb1eb0cd6918fde1ef32bb05310e5/app/controllers/shopify_app/sessions_controller.rb#L50) as [its 3rd argument](http://www.rubydoc.info/gems/shopify_api/ShopifyAPI%2FSession:initialize). This way, one can persist email, locale, etc... when storing a session when using `per_user_permissions`. 

This is what `ShopifyAPI::Session.new`'s 3rd argument is for.

PR #887, opened 2020-02-05
Support for simultaneous user & shop tokens

slucaskim:
Addresses: https://github.com/Shopify/shopify_app/issues/883
### Overview
We want to get early feedback on an attempt to support both user and shop tokens. Mostly, this involved breaking out the `SessionStorage` class into User and Shop specific classes, and explicit `store` and `retrieve` methods for the two types of session storages.

### Next steps
- [x] Change callback controller to validate the shop token (with a small authenticated call to core), and refresh it if needed
- [x] If shop_token is deemed valid, store the user token in the user session repository.
- [x] This way, by the time webhooks are installed in`install_webhooks` we can safely and explicitly use the shop token from the shop session repository.
- [x] Remove `per_user_tokens` configuration variable, and just rely on the presence of `user_session_repository` configuration variable to switch into this new strategy of supporting both Shop and User tokens.
- [x] This gem [should _not_ support](https://github.com/Shopify/shopify_app/pull/887/files#diff-deaed2b262ec885f4e36de05621e41eaR39) a workflow of managing ONLY user tokens, it should either be shop tokens only, or shop tokens and user tokens

### Testing instructions
To test shop token only setup
```
# initializers/shopify_app.rb
config.shop_session_repository = Shop

# models/Shop.rb
include ShopifyApp::ShopSessionStorage
```

To test shop & user token setup, we need to some "next steps" outlined above first. Once they're in place we can test by doing the following.
```
# initializers/shopify_app.rb
config.shop_session_repository = Shop
config.user_session_repository = User
# remove per_user_tokens setting

# models/Shop.rb
include ShopifyApp::ShopSessionStorage

# models/User.rb
include ShopifyApp::UserSessionStorage
```
@tanema In the meantime before next steps are complete, you can test the function of requesting both  by manually toggling the `per_user_permissions` option in the oauth initializer. Or you can also pull https://github.com/Shopify/pos-channel/pull/1747 to try an app that has automatic switching based on the `shop.installed?` flag. The goal would be to port over that sort of auto-switching (based on checking shop token's validity with small API call) into this PR.

### Breaking changes overview
_If your app is using shop based tokens_
 - For classes that include `ShopifyApp::Authenticated` concern, `shopify_session` around action is renamed to `activate_ shopify_session`
 - For classes that include `ShopifyApp::LoginProtection`, `shop_session` is renamed to `current_shopify_session`
 - `session[:shopify]` is no longer set; if your app depends on that, use `session[:shop_id]` instead
 - In your model that stores the shop token, replace `include ShopifyApp::SessionStorage` with `include ShopifyApp::ShopSessionStorage`

_If your app is using user based tokens_
 - For classes that include `ShopifyApp::Authenticated` concern, `shopify_session` around action is renamed to `activate_ shopify_session`
 - For classes that include `ShopifyApp::LoginProtection`, `shop_session` is renamed to `current_shopify_session`
 - `session[:shopify]` is no longer set; if your app depends on that, use `session[:user_id]` instead
 - In your model that stores the user token, replace `include ShopifyApp::SessionStorage` with `include ShopifyApp::UserSessionStorage`
 - In `initializers/omniauth.rb`, ensure following configuration is set:
```
strategy.options[:per_user_permissions] = strategy.session[:user_tokens]
```
 - In `initializers/shopify_app.rb`, ensure following configuration is set (note both shop _and_ user must be set):
```
  config.shop_session_repository = {SHOP_MODEL_CLASS_NAME}
  config.user_session_repository = {USER_MODEL_CLASS_NAME}
```


tanema:
I am just going to link all of these here so that we can start building up the scope of this and keep track of who wants it (because I am trying to clean up issues :P )
- https://github.com/Shopify/shopify_app/issues/883
- https://github.com/Shopify/shopify_app/issues/569
- https://github.com/Shopify/shopify_app/issues/618

tanema:
Okay I tophatted this and everything works great. I have a user and a shop in my database. My only concern is that shop_session is no longer just for a shop and I wonder if we should rename it as `shopify_session` and then add a method called `shop_session` that calls the other but with a deprecation warning. This will be a breaking release but shop_session is integral to how people integrate their apps with this library.

It also seems like I did not need to manually set `per_user_permissions` because you added the check for if the user session existed. But definitely note these things. We need to add a migration guide to this work. Change SessionStorage to ShopSessionStorage, both in config and in the includes, then also that change for per_user_permissions.

slucaskim:
@tanema 
> ...with a deprecation warning
While I agree we should separate `shop_session` and `user_session` into two different concepts, I don't think we should deprecate the `shop_session` because that is fine in the case of just using shop tokens for authentication.

app/controllers/concerns/shopify_app/authenticated.rb
```diff
@@ -9,7 +9,7 @@ module Authenticated
       include ShopifyApp::LoginProtection
       include ShopifyApp::EmbeddedApp
       before_action :login_again_if_different_user_or_shop
-      around_action :shopify_session
+      around_action :activate_shopify_session
```
tanema: I see what you are doing here however, people write their own integrations with this action. People wrap certain parts of their code in shopify_session. And since you are not removing the method named this, this could cause some serious confusion

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -8,6 +8,11 @@ class CallbackController < ActionController::Base
     def callback
       if auth_hash
         login_shop
+
+        if session[:shop_id].blank? && ShopifyApp::SessionRepository.user_storage.present?
```
tanema: What is this catching? Is this meant for the user authentication?

slucaskim: Just pushed another change after some testing/refinements that should be clearer to understand. Yes, this line is meant to kick off the 2nd oauth flow in the case user tokens are wanted.

app/controllers/concerns/shopify_app/authenticated.rb
slucaskim: You're right, so I'll change `shopify_session` to `current_shopify_session` so that it's a more obvious breaking change.

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -55,24 +60,23 @@ def set_shopify_session
...
+      session[:user_session] = auth_hash&.extra&.session
     end
 
     def install_webhooks
       return unless ShopifyApp.configuration.has_webhooks?
 
       WebhooksManager.queue(
         shop_name,
-        token,
+        shop_session&.token || user_session.token,
```
vinistock: Should we use the safe navigator for the user token as well? I believe the session can be `nil` too.

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -82,7 +86,7 @@ def install_scripttags
 
       ScripttagsManager.queue(
         shop_name,
-        token,
+        shop_session&.token || user_session.token,
```
vinistock: Should we use the safe navigator for the user token as well? I believe the session can be `nil` too.

app/controllers/shopify_app/callback_controller.rb
slucaskim: We should never be in a situation where we enqueue the job with a `nil` token as it will definitely break the job. I would want that to raise here before enqueing the job.

README.md
```diff
@@ -225,32 +223,35 @@ $ rails generate shopify_app:shop_model
 This will generate a shop model which will be the storage for the tokens necessary for authentication.
 
 #### User-based token storage
-A more granular control over level of access per user on an app might be necessary, to which the shop-based token strategy is not sufficient. Shopify supports a user-based token storage strategy where a unique token to each user can be managed.
+A more granular control over level of access per user on an app might be necessary, to which the shop-based token strategy is not sufficient. Shopify supports a user-based token storage strategy where a unique token to each user can be managed. Shop tokens must still be maintained so that background jobs can make use of them when necessary.
```
tanema: *must*? Maybe this should be `Shop tokens must still be maintained if you are running background jobs so that you can make use of them when necessary`

test/app_templates/config/initializers/shopify_app.rb
```diff
@@ -4,5 +4,6 @@
   config.secret = ENV['SHOPIFY_API_SECRET']
   config.scope = 'read_orders, read_products'
   config.embedded_app = true
-  config.session_repository = 'ShopifyApp::InMemorySessionStore'
+  config.shop_session_repository = 'ShopifyApp::InMemoryShopSessionStore'
+  config.user_session_repository = 'ShopifyApp::InMemoryUserSessionStore'
```
tanema: Should we deafult to making them use a user_session or allow them to add it later? We could add another generator to add user session functionality then generate the User model and add this line?

slucaskim: It is already like as you say, this line is to setup the InMemory storage, so that tests can run. There's a separate User generator that will set this to an ActiveRecord model

tanema: ohhhhh thanks yeah I thought this was the template

lib/shopify_app/session/session_repository.rb
```diff
@@ -3,31 +3,56 @@ class SessionRepository
...
+      def store_user_session(session, user)
+        user_storage.store(session, user)
+      end
+
+      def shop_storage
+        load_shop_storage || raise(ConfigurationError.new("ShopifySessionRepository.shop_storage is not configured!"))
+      end
+
+      def user_storage
+        load_user_storage
```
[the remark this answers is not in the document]
tanema: absolutely not, User Session storage is not required and a change like that would immediately break every app that uses this gem and required them all to implement a User Session Storage. As well as the fact that all of the handling in the PR checks if the user session storage is nil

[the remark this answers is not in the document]
rezaansyed: Ah I see. I was assuming that `user_storage` method would only be called whenever `User` was configured and so if that happened, it would be an error. Thanks for clarifying!

ragalie:
Hi all! Curious what the timing is looking like for getting this in. Anything I can do to help?

tanema:
Honestly, I have just been dragging my feet because this is a big change that is likely to cause a spike in support. @slucaskim can you rebase this? I can merge it and add the migration guide to the README tomorrow.

tanema:
Hi!, I have been sick this week and am swamped the rest of the week. If you want this out this week, can you add my migration notes to the README in a more readable way and merge? Otherwise, I do not think I will get this until next week

slucaskim:
@tanema Please hold off merging for now, @rezaansyed has found a potential bug which we're investigating.

slucaskim:
> @tanema Please hold off merging for now, @rezaansyed has found a potential bug which we're investigating.

@tanema, @rezaansyed has fixed the error and we confirmed that things are good now. This is ready to merge 👍

PR #892, opened 2020-02-07
Remove rescue in SameSite middleware

jenshenny:
The rescue true in `same_site_none_incompatible` isn't the best way to catch errors. This should catch all the `NoMethodErrors` that should be expected if the BrowserSniffer can't parse some user agents when they are custom.


jenshenny:
@tanema Don't know if I understand what you mean completely. Once I removed the rescue true and tested a weird user agent (ex `Custom User Agent`), the browser name checks were breaking because the browser name was `nil` as BrowserSniffer couldn't detect the browser name in the user agent

jenshenny:
Closing as I'm implementing this functionality into [browser_sniffer](https://github.com/Shopify/browser_sniffer/pull/25/files) itself

PR #895, opened 2020-02-11
Revert "Fix return_to address in Safari after enable_cookies and grated_storage_access"

hurracrane:
This reverts commit 301da901461b7fac1c3e0f8108da0ffb9cd61dfb. Introduced in https://github.com/Shopify/shopify_app/pull/873.

Login protection redirects a user to login, needing to request storage access. Due to [this line](https://github.com/Shopify/shopify_app/blob/master/app/controllers/shopify_app/sessions_controller.rb#L141) and the `return_so` session value, we are never redirected to the granted session storage page, which results in an infinite loop of redirecting back to user login requesting storage access.

### DIFF
--- a/app/assets/javascripts/shopify_app/storage_access.js
+++ b/app/assets/javascripts/shopify_app/storage_access.js
@@ -28,8 +28,8 @@
     window.parent.location.href = this.redirectData.myshopifyUrl + '/admin/apps';
   }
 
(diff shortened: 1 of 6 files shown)


tanema:
cc @joshreeves

hurracrane:
Will create a revert of this revert so we can address both issues.

Issue #870, opened 2020-01-21
Error: CSRF detected


Shine18:
I am redirecting the user to install the app directly if they have removed the app from store and accessing any page directly 
with this code

[http://prntscr.com/qqxqs7](http://prntscr.com/qqxqs7)

but omniauth throws error about csrf vulnerability. 
[http://prntscr.com/qqxrcy](http://prntscr.com/qqxrcy)

Any help/hint would be appreciated.

BoomDig:
We, too, started to see the same error in our logs when we upgraded our Rails 5 app from shopify_app 11.3.2 to 11.7.0 (in a test environment with a private app) and installed the app in a shop. A relevant snippet from our logs:

INFO -- omniauth: (shopify) Setup endpoint detected, running now.
INFO -- omniauth: (shopify) Callback phase initiated.
ERROR -- omniauth: (shopify) Authentication failure! csrf_detected: OmniAuth::Strategies::OAuth2::CallbackError, csrf_detected | CSRF detected

And like some others reported recently when upgrading to version 11.6 or 11.7 of this gem, the error you get when installing an app comes and goes and is not repeatable. Hoping Shopify can investigate this and offer a solution before Feb 4!

Shine18:
Yes, the issue is critical. I'm developing a client's app, and timeline is delayed because of the issue.

tanema:
I will try to look at this today.

tanema:
Can you please give me steps to reproduce? Is it the same issue as https://github.com/Shopify/shopify_app/issues/861 ?

BoomDig:
Our app (created initially with the shopify_app generator) is on Rails 5.2 using Ruby 2.6.5. When we put the app on a test server and add it as a private app to a test shop we get the csrf_detected error after seeing Shopify's permission screen and clicking on the button to allow installation. The app does get installed, though, despite the error. We've applied the work-around described in the discussion to issue 589: https://github.com/Shopify/shopify_app/issues/589 (see vfonic's "acceptable solution"), which appears to be working. But we didn't run into this issue before upgrading from shopify_app 11.3.2 to 11.7.0.

tanema:
Thanks for pointing me to the issue, that really helps with historic issues. I will see what I can do.

overallduka:
@Shine18 Maybe the error is related to existent cookies, because I had had the same, but it goes away when I clear my cookie, maybe you should try.

[...]

PR #898, opened 2020-02-11
Inserting SameSite middleware higher in stack

katiedavis:
Pairprogramming with @tanema

### What we learned
We learned a couple things while working on this.
If the user does not have a tunnel
- They will receive a csrf error on the main login form
- The omniauth session wont be stored because it wont be on https

### What we did
We had to do a few things to make sessions more stable
- appropriately cased the `Secure` attribute on the cookie
- Rewrote the middleware to return the status, headers, body to conform to middleware interface
- Moved the middleware higher in the middleware order. We found examples in the wild also editing cookies where the middleware was inserted higher so we copied that and it worked better.

@jenshenny  I know you are implementing this middleware across a bunch of different applications. Do these changes seem to fix any issues you are seeing?

### DIFF
--- a/lib/shopify_app/engine.rb
+++ b/lib/shopify_app/engine.rb
@@ -14,7 +14,7 @@ module ShopifyApp
     end
 
     initializer "shopify_app.middleware" do |app|
-      app.config.middleware.insert_after(ActionDispatch::Static, ShopifyApp::SameSiteCookieMiddleware)
+      app.config.middleware.insert_after(::Rack::Runtime, ShopifyApp::SameSiteCookieMiddleware)
     end
   end
 end
--- a/lib/shopify_app/engine.rb
+++ b/lib/shopify_app/engine.rb
@@ -14,7 +14,7 @@ module ShopifyApp
     end
 
     initializer "shopify_app.middleware" do |app|
-      app.config.middleware.insert_before(ActionDispatch::Cookies, ShopifyApp::SameSiteCookieMiddleware)
+      app.config.middleware.insert_after(ActionDispatch::Static, ShopifyApp::SameSiteCookieMiddleware)
     end
   end
 end
--- a/lib/shopify_app/middleware/same_site_cookie_middleware.rb
+++ b/lib/shopify_app/middleware/same_site_cookie_middleware.rb
@@ -5,8 +5,7 @@ module ShopifyApp
     end
 
     def call(env)
-      _status, headers, _body = @app.call(env)
-    ensure
+      status, headers, body = @app.call(env)
       user_agent = env['HTTP_USER_AGENT']
 
       if headers && headers['Set-Cookie'] && !SameSiteCookieMiddleware.same_site_none_incompatible?(user_agent) &&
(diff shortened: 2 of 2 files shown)


sshaw:
After seeing this problem myself (#870) I don't think this will fix it. 

I haven't done a deep-dive (that's why you guys get paid the big bucks... 🤔) but once you modify the cookie you're _probably_ invalidating the `hmac` param provided to `/auth/shopify/callback`. 

And if that wasn't invalidating the HMAC you probably [still want to `ensure`](https://github.com/Shopify/shopify_app/pull/898/files#diff-89520877523f34645e272383ae801163L9) the appropriate cookie attributes are set otherwise the error page will not render.

sshaw:
> I haven't done a deep-dive (that's why you guys get paid the big bucks... 🤔)

Hrmmmmm 🆗, I just noticed it's 2020-02-12 (I usually only go by month) and am starting to get a wee bit concerned here so I did a `super`-lightweight-dive:

What's the reasoning for inserting the middleware _after_  `ActionDispatch::Static`?

Since an invalid HMAC appears to be the issue, the middleware should be inserted _before_ the HMAC is generated, which, translated to Rails API calls is:

```rb
Rails.application.config.middleware.insert_after(OmniAuth::Builder, ShopifyApp::SameSiteCookieMiddleware)
``` 

This seems to fix things for me. But do note that I'm using a custom version of shopify_app to avoid breakages like these... 😈 😆

katiedavis:
Hi @sshaw, unfortunately setting the middleware that early in the process causes the cookies to not be set with the SameSite and Secure attributes which causes a loop in rendering as it can't load in the iframe. I'm not sure about the specific set up of you app.

tanema:
@sshaw First, thanks for the feedback, it is helpful.

> you probably still want to ensure the appropriate cookie attributes are set otherwise the error page will not render.

That ensure runs even if there is an error, but if we have page error, the cookie will probably not work or be set either. However it is not a bad idea to leave it in there.

> What's the reasoning for inserting the middleware after ActionDispatch::Static

The simple reasoning is that we have looked at other middleware that is editing cookies and they have done it in a similar place and when we moved the middleware to that location our apps more consistently authenticated. However, I have a suspicion that we could actually just insert if before the cookie middleware instead of after the cookie middleware as it was originally. We have not narrowed down the exact issue yet with the cookies but have noticed that in its current place, the secure and samesite settings don't always make it to the browser. Checking the headers just curling our login path displays that issue

Currently, I don't have a reason to think it is an issue with HMACs, can you explain your thinking in that regards?

Again thanks for your input.

sshaw:
> Currently, I don't have a reason to think it is an issue with HMACs, can you explain your thinking in that regards?

Well keep in mind I'm not working off the same stack but the reason I think this to be the case is because if I keep the middleware in but comment out the part that `#gsub`s the cookie it works.

I do not know OAuth2 internals but it does ensure that the request has not been tampered with (well, obviously 😄) and how this is determined _could_ be based of a "signed cookie" or similar whose signature no longer matches since the value has been `#gsub`'d. Of course I'm assuming a bit here but keeping the middleware but commenting out the cookie `#gsub`ing does not cause the CSRF error for me.

tanema:
Okay I took tips from the people who develop the same_site_cookie gem where to place it in the middleware as well as made it check for secure and samesite so that we don't duplicate any of those cookie settings. This is consistently working for me, and setting the attributes appropriately. Apparently ActionDispatch::Static is not always added but Rack::Runtime always is so it is a more stable place to place the middleware.

sshaw:
@tanema have you tried this running locally via the "Test app on development store" option for an app that's not embedded?

PR #837, opened 2019-12-12
remove dotenv-rails from generator

tylerball:
#835 remove dotenv but it is still in the generator. I will re-release 11.4.0 with this fix.

PR #910, opened 2020-02-26
adds note about dot-env gem

katiedavis:
During OAuth, the shopify_app initializer template [expects ENV variables](https://github.com/Shopify/shopify_app/blob/master/lib/generators/shopify_app/install/templates/shopify_app.rb#L3). 

This adds a note to read the env vars.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,9 @@ SHOPIFY_API_KEY=your api key
 SHOPIFY_API_SECRET=your api secret
 ```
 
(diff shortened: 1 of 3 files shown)


kevinhughes27:
`.env` is only one of the ways to manage env vars, this was actually recently removed from this gem to allow devs to make their own choice:

https://github.com/Shopify/shopify_app/pull/835
https://github.com/Shopify/shopify_app/pull/837

katiedavis:
> `.env` is only one of the ways to manage env vars, this was actually recently removed from this gem to allow devs to make their own choice:
> 
> #835
> #837

Noted! I updated the readme instead

Issue #911, opened 2020-02-27
Upgrading shopify_app from 11.1.0 to 11.6.0 is corrupting session


knagode:
When trying to upgrade gem I noticed that Rails' session stopped working for other part of the app if visited trough localhost:3000 (or in test environment). 

Everything works via ngrok but it is annoying because e.g. ActiveAdmin doesn't work anymore on localhost:3000. 

I used initialiser to fix session_store on localhost:3000 and test env:

```
Rails.application.config.session_store :cookie_store, key: '_session', same_site: :strict
```
However this breaks the app on production or when using ngrok. 

Any idea how to approach this problem properly?

tanema:
Yes, you now need to always use a tunnel. Because of the SameSite policy required by Chrome, the app needs to be served over https so that we can add the secure attribute to cookies. The secure attribute is required when setting SameSite=none.  The session is not corrupt, it will just not set the cookie because you are serving over http but the secure attribute is on the cookie. When you set same_site: :strict, you are overriding our middleware so that is why it fixes it but then it will not work in production for your users with updated Chrome

knagode:
I did that because my rspec feature tests started failing (any tests which were using session somehow - apparently because test environment doesn't use SSL)
Also because e.g. ActiveAdmin stopped working if visited via localhost:3000. 

It seems like adding `secure` attribute to cookies should only happen if request is using SSL. Otherwise it is affecting parts of the app which are irrelevant to Shopify. That way we could use the app on both http://localhost:3000 (parts which are not relevant to shopify) and https://xyz.ngrok.io. Or is there a better solution for this?

knagode:
https://github.com/Shopify/shopify_app/pull/912

I played with it and it seems like this would spare many headaches :)

PR #912, opened 2020-02-27
Do not modify cookie policy on non SSL requests

knagode:
Changing cookie policy for all cookies (not only those which are relevant to Shopify App) can easily brake other functionalities of someone's app. 

E.g.: We use this gem but our Rails app does other things as well - e.g.: we have Active Admin. Developing on http://localhost:3000 became impossible because of this middleware. 

I had similar problems with rspec tests which were testing session related things (outside ShopifyApp scope). Since feature tests are running without SSL - all those tests broke. 

It seems like many problems would be solved if we would only be fixing cookies on https requests.

Relevant issue: https://github.com/Shopify/shopify_app/issues/911

### DIFF
--- a/lib/shopify_app/middleware/same_site_cookie_middleware.rb
+++ b/lib/shopify_app/middleware/same_site_cookie_middleware.rb
@@ -10,13 +10,10 @@ module ShopifyApp
       status, headers, body = @app.call(env)
       user_agent = env['HTTP_USER_AGENT']
 
-      @request = Rack::Request.new(env) 
-      is_https = @request.env['HTTPS'] == 'on' || @request.env['HTTP_X_SSL_REQUEST'] == 'on'
-
       if headers && headers['Set-Cookie'] &&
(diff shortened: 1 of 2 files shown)


lib/shopify_app/middleware/same_site_cookie_middleware.rb
```diff
@@ -10,9 +10,13 @@ def call(env)
       status, headers, body = @app.call(env)
       user_agent = env['HTTP_USER_AGENT']
 
+      @request = Rack::Request.new(env) 
+      is_https = @request.env['HTTPS'] == 'on' || @request.env['HTTP_X_SSL_REQUEST'] == 'on'
```
tanema: You can check this by simply doing `Rack::Request.new(env).ssl?`

knagode: +1

knagode:
@tanema Since there were not middleware tests I added one to test exactly this feature: Check whether cookies are altered only on SSL.

PR #913, opened 2020-02-28
Return to relative paths only

Abdulwahaab710:
## Problem

The shopify_app is currently vulnerable to open redirect. Open redirect is vulnerability that allows an attacker to redirect a user to an arbitrary domain of the attackers choosing. To read more about open redirect please check this [link](https://portswigger.net/kb/issues/00500100_open-redirection-reflected) or [video](https://www.youtube.com/watch?v=4Jk_I-cw4WE)

### Setting up Shopify_app

1. First clone the repo

2. Create a new Ruby on Rails project
  ```bash
  rails new test-project
  ```

3. Add the shopify_app gem into your Gemfile and make it point to the local shopify_app
  ```ruby
  gem 'shopify_app', path: 'THE GEM LOCAL PATH'
  ```

4. Install the gem by running:
  ```bash
  bundle install
  ```

5. Generate a shopify_app using rails generator
  ```bash
  rails generate shopify_app --api_key <YOUR API KEY>  --secret <YOUR SECRET>
  ```

6. Run the application
  ```bash
  rails s
  ```

7. Install this application on your development store.

### Reproducing the open redirect vulnerability

Let's first checkout the master branch to reproduce the issue.

After Installing your application on your Shopify store, go to `https://<YOUR_APPLICATION_DOMAIN>/login?return_to=https://google.com&shop=<YOUR_SHOP>.myshopify.com`. You should be redirect to google.com if you are already logged in and if not just go through the OAuth flow and then you should redirected to google.com.

screencast_2020-03-02_17-37-16

### :tophat:ing the change
Now switch to this branch `fix-open-redirect` and try to visit the previous url `https://<YOUR_APPLICATION_DOMAIN>/login?return_to=https://google.com&shop=<YOUR_SHOP>.myshopify.com`. You should be redirect to the root of your application.

screencast_2020-03-02_17-34-05

### DIFF
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -139,7 +139,7 @@ module ShopifyApp
       query_params = {}
       query_params[:shop] = sanitized_params[:shop] if params[:shop].present?
 
-      return_to = session[:return_to] || params[:return_to]
+      return_to = RedirectSafely.make_safe(session[:return_to] || params[:return_to], nil)
 
       if return_to.present? && return_to_param_required?
         query_params[:return_to] = return_to
--- a/app/controllers/shopify_app/sessions_controller.rb
+++ b/app/controllers/shopify_app/sessions_controller.rb
@@ -125,7 +125,7 @@ module ShopifyApp
     end
 
     def copy_return_to_param_to_session
-      session[:return_to] = params[:return_to] if params[:return_to]
+      session[:return_to] = RedirectSafely.make_safe(params[:return_to], '/') if params[:return_to]
     end
 
     def render_invalid_shop_error
--- a/lib/shopify_app.rb
+++ b/lib/shopify_app.rb
@@ -4,6 +4,7 @@ require 'shopify_app/version'
 # deps
 require 'shopify_api'
 require 'omniauth-shopify-oauth2'
+require 'redirect_safely'
 
 module ShopifyApp
   def self.rails6?
--- a/shopify_app.gemspec
+++ b/shopify_app.gemspec
@@ -18,6 +18,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('shopify_api', '~> 9.0.2')
   s.add_runtime_dependency('omniauth-shopify-oauth2', '~> 2.2.2')
   s.add_runtime_dependency('jwt', '~> 2.2.1')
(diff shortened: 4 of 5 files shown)


tanema:
What is your reason for this? This changes the functionality in a way I cannot predict how it will affect all the apps using this gem.

Abdulwahaab710:
👋 Hey @tanema, I have updated the PR description. 

For your comment about:
>This changes the functionality in a way I cannot predict how it will affect all the apps using this gem.

This shouldn't effect any application that redirects back to it self. 
Let's say that we have a user who visited and got redirected to a route on your application, but they weren't authenticated, so first thing you want to do is authenticate the user and the redirect them back to where they were. The usage for `return_to` is to **return the user back to your application**. The `return_to` path is always relative to your application url.

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -116,6 +116,21 @@ def login_url_params(top_level:)
       query_params
     end
 
+    def return_to_relative_path
+      relative_path(session[:return_to]) || relative_path(params[:return_to])
+    end
+
+    def relative_path(url)
```
JackMc: This logic seems like it would do unexpected things. For example if I set the URI to `https://myotherdomain.com/login` it'll get rewritten to `/login`. An alternative would be to check if the parsed URI is relative, and to set the path to `/` if not. The only snag here would be the situation where it's equal to the request host (and therefore safe) but still fully qualified. The code would look something like:

```ruby
uri = URI.parse(url)
if uri.relative?
  uri.to_s
else
  '/'
end
```

fresh-eggs:
@Abdulwahaab710 is it possible to make use of [Safe Redirect](https://github.com/Shopify/safe_redirect) instead of trying to roll our own solution ?

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -139,7 +139,7 @@ def login_url_params(top_level:)
       query_params = {}
       query_params[:shop] = sanitized_params[:shop] if params[:shop].present?
 
-      return_to = session[:return_to] || params[:return_to]
+      return_to = RedirectSafely.make_safe((session[:return_to] ? session[:return_to] : params[:return_to]), nil)
```
JackMc, line 142: Could this be changed to `session[:return_to] || params[:return_to]`

PR #915, opened 2020-03-02
Migrate redirect calls to App Bridge

vividviolet:
Fixes https://github.com/Shopify/app-bridge/issues/1410

### Tophat

1. In a project consuming Shopify App Gem, open the gemfile
2. Change the gem path for "shopify_app" to point to the local directory of the Shopify_app repo. Ex. 
> gem "shopify_app", path: "../shopify_app"
3. Ensure that you can build and run the project

### DIFF
--- a/app/assets/javascripts/shopify_app/app_bridge_redirect.js
+++ b/app/assets/javascripts/shopify_app/app_bridge_redirect.js
@@ -1,18 +1,18 @@
-(function() {
+(function (window) {
   function appBridgeRedirect(url) {
     var AppBridge = window['app-bridge'];
     var createApp = AppBridge.default;
     var Redirect = AppBridge.actions.Redirect;
     var app = createApp({
       apiKey: window.apiKey,
-      shopOrigin: window.shopOrigin.replace(/^https:\/\//, ''),
+      shopOrigin: window.shopOrigin,
     });
-  
+
     var normalizedLink = document.createElement('a');
     normalizedLink.href = url;
-  
+
     Redirect.create(app).dispatch(Redirect.Action.REMOTE, normalizedLink.href);
   }
 
-  this.appBridgeRedirect = appBridgeRedirect;
+  window.appBridgeRedirect = appBridgeRedirect;
 })(window);
--- a/app/assets/javascripts/shopify_app/app_bridge_redirect.js
+++ b/app/assets/javascripts/shopify_app/app_bridge_redirect.js
@@ -1,0 +1,18 @@
+(function() {
+  function appBridgeRedirect(url) {
+    var AppBridge = window['app-bridge'];
+    var createApp = AppBridge.default;
+    var Redirect = AppBridge.actions.Redirect;
+    var app = createApp({
+      apiKey: window.apiKey,
+      shopOrigin: window.shopOrigin.replace(/^https:\/\//, ''),
+    });
+  
+    var normalizedLink = document.createElement('a');
+    normalizedLink.href = url;
+  
+    Redirect.create(app).dispatch(Redirect.Action.REMOTE, normalizedLink.href);
+  }
+
+  this.appBridgeRedirect = appBridgeRedirect;
+})(window);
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/shopify_app/storage_access.js
```diff
@@ -1,3 +1,4 @@
+//= require ./app_bridge_redirect.js
```
katiedavis: Do we need this?

katiedavis:
@vividviolet The code LGTM, I was wondering if you can add some top hatting instructions? What we should look for? Any trip wires etc

app/assets/javascripts/shopify_app/storage_access.js
vividviolet: Yes, this is the shared helper used to call App Bridge to perform the redirect

vividviolet:
@katiedavis I'm not sure what the typical top hat process is like for this repo. I've asked @416serg to test it with Google Shopping by consuming the Gem

416serg:
Unfortunately our app is behind a few versions so I won't be able to top hat until we get our dependencies updated, which require some code refactoring on the back end, should be done by EOW though at which point I'll 🎩

vividviolet:
Related to: https://github.com/Shopify/app-bridge/issues/1410

app/assets/javascripts/shopify_app/app_bridge_redirect.js
```diff
@@ -0,0 +1,18 @@
+(function() {
+  function appBridgeRedirect(url) {
+    var AppBridge = window['app-bridge'];
+    var createApp = AppBridge.default;
+    var Redirect = AppBridge.actions.Redirect;
+    var app = createApp({
+      apiKey: window.apiKey,
+      shopOrigin: window.shopOrigin.replace(/^https:\/\//, ''),
+    });
```
greatestape, line 9: It looks like we initialize AppBridge here: https://github.com/Shopify/shopify_app/blob/master/lib/generators/shopify_app/install/templates/shopify_app.js

Maybe we can just rely on `window.app` here?

app/views/shopify_app/shared/redirect.html.erb
```diff
@@ -1,10 +1,15 @@
...
+<html lang="<%= I18n.locale %>">>
 <head>
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   <base target="_top">
   <title>Redirecting…</title>
+  <script src="https://unpkg.com/@shopify/app-bridge@^1"></script>
+  <script>
+    window.apiKey = "<%= ShopifyApp.configuration.api_key %>";
+    window.shopOrigin = "https://<%= current_shopify_domain %>";
```
greatestape, line 11: If we stop setting this here, we can stop removing it on the other side.

app/views/shopify_app/sessions/request_storage_access.html.erb
```diff
@@ -5,6 +5,11 @@
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   <base target="_top">
   <title>Redirecting…</title>
+  <script src="https://unpkg.com/@shopify/app-bridge@^1"></script>
+  <script>
+    window.apiKey = "<%= ShopifyApp.configuration.api_key %>";
+    window.shopOrigin = "https://<%= current_shopify_domain %>";
```
greatestape, line 11: If the only thing consuming this is App Bridge, and it doesn't want the `https://`, maybe we should drop it

app/assets/javascripts/shopify_app/app_bridge_redirect.js
greatestape: Ah, I think I'm wrong and that js isn't getting loaded, and also, it depends on the api key and shop origin params being set in a different place. It looks like the lib has two different patterns for doing this right now. I don't know if now is the right time to fix it, but it's tempting.

henrytao-me: Using `script` tag in the `head` is more reliable I think as it doesn't have to wait until DOMLoaded.

app/views/shopify_app/shared/redirect.html.erb
henrytao-me: I prefer to keep `https://` for consistency. The app gem is doing it in other pages.

henrytao-me:
Just rebase again as it is out of sync with master.

PR #916, opened 2020-03-02
Adding documentation about CallbackController and how to upgrade to 8.4.0 or newer

### LINKED ISSUE #754: Update the changelog around the new callback_controller in Shopify App gem
I was bumping the `shopify_app` gem in my app from 8.2.6 to 8.6.1 and I was running into errors because of a missing callback action in our `sessions_controller` that extended from `ShopifyApp::SessionsController`. It turns out that we took out the callback action and its helper methods in `ShopifyApp::SessionsController` and put it in `ShopifyApp::CallbackController`.

To prevent other devs from running into this issue, we should update the changelog/docs for the version that introduced this breaking change.

gbossh:
Add missing documentation, including a slightly more expanded explanation of what happens when the callback is handled and a link to the changelog from the readme.

This closes #754.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ Authentication
 
 ### Callback
 
-Upon completing the authentication flow Shopify calls the app at the `callback_path` mentioned before. If the app needs to do some extra work it can define and configure the route to a custom callback controller, inheriting from `ShopifyApp::CallbackController` and hook into or override any of the defined helper methods. The default callback controller provides already the following behaviour:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -213,6 +213,16 @@ end
 Authentication
 --------------
 
+### Callback
+
+Upon completing the authentication flow Shopify calls the app at the `callback_path` mentioned before. If the app needs to do some extra work it can define and configure the route to a custom callback controller, inheriting from `ShopifyApp::CallbackController` and hook into or override any of the defined helper methods. The default callback controller provides already the following behaviour:
```
katiedavis:
```suggestion
Upon completing the authentication flow Shopify calls the app at the `callback_path` mentioned before. If the app needs to do some extra work it can define and configure the route to a custom callback controller, inheriting from `ShopifyApp::CallbackController` and hook into or override any of the defined helper methods. The default callback controller already provides the following behaviour:
```

PR #920, opened 2020-03-04
Validate jwt shop and user

rezaansyed:
### What are you trying to accomplish?

Allow apps to be able to authenticate themselves via JWT a `Authorization` headers from incoming requests.

### What approach did you choose and why?

We extended the functionality of the `ShopSessionStorage` and `UserSessionStorage` classes such that the session can be retrieved via the `shopify_domain` and `shopify_user_id` respectively. The `Authorization` header we expect will be a JWT where the payload will include the following identifying information to be used in the session:
- `dest`: shopify domain of the shop
- `sub`: shopify user id

Thus, the JWT class allows us to interface with the incoming header and retrieve the `shopify_domain` and `shopify_user_id` needed to make Shop and User session tokens respectively.

To ensure that this does not have any effect on any consumers of `shopify_app` we have added a configuration flag `allow_jwt_authentication` which is defaulted to false. Even when set to true and no JWT header is passed in, we fallback to cookie based session retrieval.

### What should reviewers focus on?

- Session retrieval logic in `ShopSessionStorage` and `UserSessionStorage` classes
- Refactor in SessionRepository
- JWT class used to parse JWT header and retrieve `shopify_domain` and `shopify_app` logic

### The impact of these changes

- No immediate impact on consumers until the session identifier flow via JWT is complete across dependent systems
- With a complete working flow, consumers can only enable feature by switching `allow_jwt_authentication` to true

### Tophatting instructions
Please refer to this [doc](https://docs.google.com/document/d/1W2ncTaQYucpzXVOZbJYDXgma1e7GwMPh8y6SVb69wCQ/edit#heading=h.53xsfnwuz3ta) for steps to tophatting this change.

## Before you deploy

- [x] I [tophatted](https://development.shopify.io/engineering/developing_at_Shopify/write-code/tophatting) or tested this change.
- [x] This PR is [safe to rollback](https://development.shopify.io/engineering/developersToolbox/ship/safe_to_merge#Signs_your_PR_is_safe_to_rollback).


lib/generators/shopify_app/install/templates/shopify_app.rb
```diff
@@ -9,6 +9,7 @@
   config.after_authenticate_job = false
   config.api_version = "<%= @api_version %>"
   config.shop_session_repository = 'ShopifyApp::InMemoryShopSessionStore'
+  config.allow_jwt_authentication = false
```
ragalie: I'd probably keep it out of the template for now since we don't want people to be using it yet. We can add it in to the template when we're ready for third-party rollout.

lib/shopify_app.rb
```diff
@@ -44,10 +44,12 @@ def self.use_webpacker?
   require 'shopify_app/middleware/same_site_cookie_middleware'
 
   # session
+  require 'shopify_app/session/jwt'
   require 'shopify_app/session/session_storage'
   require 'shopify_app/session/shop_session_storage'
   require 'shopify_app/session/user_session_storage'
   require 'shopify_app/session/session_repository'
+  require 'shopify_app/session/null_user_session_store'
```
ragalie: Alphabetize the `session_store` requires.

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -27,20 +27,36 @@ def activate_shopify_session
     end
 
     def current_shopify_session
-      if session[:user_id].present?
-        @current_shopify_session ||= user_session
-      else
-        @current_shopify_session ||= shop_session
+      @current_shopify_session ||= begin
+        user_session_by_jwt || user_session_by_cookie || shop_session_by_jwt || shop_session_by_cookie
```
ragalie: We might as well simplify this to `user_session || shop_session`

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -27,20 +27,36 @@ def activate_shopify_session
...
       end
     end
 
     def user_session
-      return if session[:user_id].blank?
+      user_session_by_jwt || user_session_by_cookie
+    end
+
+    def user_session_by_jwt
+      return unless ShopifyApp.configuration.allow_jwt_authentication && jwt_shopify_user_id
```
ragalie: I'd split these into two lines:

```
return unless ShopifyApp.configuration.allow_jwt_authentication
return unless jwt_shopify_user_id
```

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -62,6 +78,18 @@ def login_again_if_different_user_or_shop
 
     protected
 
+    def jwt_shopify_domain
+      @jwt_shopify_domain ||= JWT.new(jwt).shopify_domain if jwt
```
ragalie: I think it's easier to parse of the `if jwt` is on a separate line, e.g.

```
return unless jwt
...
```

lib/shopify_app/session/in_memory_session_store.rb
```diff
@@ -12,6 +12,10 @@ def self.store(session, *args)
       id
     end
 
+    def self.retrieve_by_jwt(payload)
```
ragalie: This looks vestigial.

Also does this class make sense / is it used anywhere anymore?

ragalie: Oh nvm I see that we subclass from it. Either way, I don't think it should have `retrieve_by_jwt` method anymore.

lib/shopify_app/session/in_memory_shop_session_store.rb
```diff
@@ -1,4 +1,7 @@
 module ShopifyApp
   class InMemoryShopSessionStore < InMemorySessionStore
+    def self.retrieve_by_shopify_domain(shopify_domain)
```
ragalie: This implementation doesn't seem right. The implementation in `InMemorySessionStore` stores by a random ID, but this expects to look up by `shopify_domain`. You may need to implement your own `self.store` here to do what you're expecting.

lib/shopify_app/session/in_memory_user_session_store.rb
```diff
@@ -1,4 +1,7 @@
 module ShopifyApp
   class InMemoryUserSessionStore < InMemorySessionStore
+    def self.retrieve_by_shopify_user_id(user_id)
```
ragalie: Same issue here: https://github.com/Shopify/shopify_app/pull/920/files#r392201432

lib/shopify_app/session/jwt.rb
```diff
@@ -0,0 +1,46 @@
...
+      @payload
+    end
+
+    def set_payload
+      parse_token_data(ShopifyApp.configuration.secret)
+      parse_token_data(ShopifyApp.configuration.old_secret) if !@payload && ShopifyApp.configuration.old_secret
+    end
+
+    def parse_token_data(secret)
+      @payload, _ = ::JWT.decode(@token, secret, true, { algorithm: 'HS256' })
```
ragalie: I would move the `@payload =` setter into `set_payload` instead of in this method. This method (by its name) should be focused on parsing and the `set_payload` method can actually do the work of setting the instance variable.

test/shopify_app/controller_concerns/login_protection_test.rb
```diff
@@ -63,6 +65,69 @@ class LoginProtectionTest < ActionController::TestCase
     end
   end
 
+  test "#current_shopify_session retrieves user session using jwt" do
+    ShopifyApp.configuration.allow_jwt_authentication = true
```
ragalie: Do you want to reset this to `false` after each test runs? You could do this in an `after` block.

test/shopify_app/controller_concerns/login_protection_test.rb
```diff
@@ -63,6 +65,69 @@ class LoginProtectionTest < ActionController::TestCase
...
+    jwt = JWT.encode(payload, nil, 'none')
+    jwt_mock = Struct.new(:shopify_user_id).new(payload['sub'])
+    ShopifyApp::JWT.stubs(:new).with(jwt).returns(jwt_mock)
+
+    expected_session = ShopifyAPI::Session.new(
+      domain: domain,
+      token: token,
+      api_version: '2020-01',
+    )
+
```
ragalie: Looks like an extra newline here

test/shopify_app/session/jwt_test.rb
```diff
@@ -0,0 +1,141 @@
+require 'test_helper'
+
+module ShopifyApp
+  class JWTTest < ActiveSupport::TestCase
+
```
ragalie: extra newline

test/shopify_app/session/jwt_test.rb
```diff
@@ -0,0 +1,141 @@
...
+    TEST_USER_ID = 'test-user'
+
+    setup do
+      ShopifyApp.configuration.api_key = 'api_key'
+      ShopifyApp.configuration.secret = 'secret'
+      ShopifyApp.configuration.old_secret = 'old_secret'
+      ShopifyApp.configuration.myshopify_domain = 'myshopify.io'
+    end
+
+    test "#payload returns the jwt payload" do
```
ragalie: These test names look wrong now.

test/shopify_app/session/null_user_session_store_test.rb
```diff
@@ -0,0 +1,19 @@
+require 'test_helper'
+
+module ShopifyApp
+  class NullUserSessionStoreTest < ActiveSupport::TestCase
+    test '.retrieve returns nil' do
+      assert_nil NullUserSessionStore.retrieve('payload')
+    end
+
+    test '.store returns nil' do
```
ragalie: Looks like it actually raises

test/shopify_app/session/session_repository_test.rb
```diff
@@ -0,0 +1,162 @@
+require 'test_helper'
+
+module ShopifyApp
+  class SessionRepositoryTest < ActiveSupport::TestCase
+
```
ragalie: extra newline

test/shopify_app/session/session_repository_test.rb
```diff
@@ -0,0 +1,162 @@
...
+      assert_raises(ArgumentError) { SessionRepository.user_storage = storage.new }
+    end
+
+    test ".user_storage= does not raise ArgumentError if input is nil" do
+      assert_nothing_raised { SessionRepository.user_storage = nil }
+    end
+
+    test ".user_storage= clears storage setting if input is nil" do
+      SessionRepository.user_storage = InMemoryUserSessionStore
+      assert_not_nil SessionRepository.user_storage
```
ragalie: It'll never be `nil` so we should probably `assert_kind_of InMemoryUserSessionStore.class, SessionRepository.user_storage`

test/shopify_app/session/session_repository_test.rb
```diff
@@ -0,0 +1,162 @@
...
+      assert_raises(ArgumentError) { SessionRepository.shop_storage = storage.new }
+    end
+
+    test ".shop_storage= does not raise ArgumentError if input is nil" do
+      assert_nothing_raised { SessionRepository.shop_storage = nil }
+    end
+
+    test ".shop_storage= clears storage setting if input is nil" do
+      SessionRepository.shop_storage = InMemoryShopSessionStore
+      assert_not_nil SessionRepository.shop_storage
```
ragalie: Same as here: https://github.com/Shopify/shopify_app/pull/920/files#r392206271

test/shopify_app/session/session_repository_test.rb
```diff
@@ -0,0 +1,162 @@
...
+      SessionRepository.shop_storage = nil
+      assert_raises(SessionRepository::ConfigurationError) { SessionRepository.shop_storage }
+    end
+
+    test '.shop_storage accepts a String as argument' do
+      SessionRepository.shop_storage = 'ShopifyApp::InMemoryShopSessionStore'
+      assert_kind_of InMemoryShopSessionStore.class, SessionRepository.shop_storage
+    end
+
+    test '.retrieve_user_session_by_jwt retrieves a user session by JWT' do
```
ragalie: Fix test name

test/shopify_app/session/shop_session_storage_test.rb
```diff
@@ -20,7 +20,28 @@ class ShopSessionStorageTest < ActiveSupport::TestCase
       assert_equal TEST_SHOPIFY_TOKEN, session.token
     end
 
-    test "tests that session store can store shop session records" do
+    test ".retrieve_by_shopify_domain can retrieve shop session records by JWT" do
```
ragalie: Fix test name

test/shopify_app/session/shop_session_storage_test.rb
```diff
@@ -20,7 +20,28 @@ class ShopSessionStorageTest < ActiveSupport::TestCase
...
+      )
+      ShopMockSessionStore.stubs(:find_by).with(shopify_domain: TEST_SHOPIFY_DOMAIN).returns(instance)
+
+      expected_session = ShopifyAPI::Session.new(
+        domain: instance.shopify_domain,
+        token: instance.shopify_token,
+        api_version: instance.api_version,
+      )
+
+      shopify_domain =  TEST_SHOPIFY_DOMAIN
```
ragalie: extra space

test/shopify_app/session/user_session_storage_test.rb
```diff
@@ -23,6 +23,28 @@ class UserSessionStorageTest < ActiveSupport::TestCase
       assert_equal TEST_SHOPIFY_USER_TOKEN, session.token
     end
 
+    test "tests that session store can retrieve user session records by JWT" do
```
ragalie: fix test name

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -27,20 +27,38 @@ def activate_shopify_session
...
 
     def user_session
-      return if session[:user_id].blank?
+      user_session_by_jwt || user_session_by_cookie
+    end
+
+    def user_session_by_jwt
+      return unless ShopifyApp.configuration.allow_jwt_authentication
+      return unless jwt_shopify_user_id
+      ShopifyApp::SessionRepository.retrieve_user_session_by_shopify_user_id(jwt_shopify_user_id)
```
greatestape: Why do we call a different method here than we do on line 47 if it's a user id in both cases?

rezaansyed: So line 47 leverages a user id that is generated by `shopify_app` but `shopify_user_id` is different where it is known across all of shopify for the particular user.

tanema:
Can I get some tophatting instructions?

lib/shopify_app/session/session_repository.rb
```diff
@@ -2,21 +2,23 @@ module ShopifyApp
...
     class ConfigurationError < StandardError; end
 
+    EXPECTED_METHODS = [:store, :retrieve].freeze
+    EXPECTED_SHOP_STORAGE_METHODS = (EXPECTED_METHODS + [:retrieve_by_shopify_domain]).freeze
+    EXPECTED_USER_STORAGE_METHODS = (EXPECTED_METHODS + [:retrieve_by_shopify_user_id]).freeze
+
     class << self
       def shop_storage=(storage)
         @shop_storage = storage
+        return unless storage
```
emmack: Does `@shop_storage` need to be set to `nil` if `storage` is `nil`? If not, maybe swap lines 11 and 12.

lib/shopify_app/session/session_repository.rb
```diff
@@ -2,21 +2,23 @@ module ShopifyApp
...
 
-        unless storage.nil? || self.shop_storage.respond_to?(:store) && self.shop_storage.respond_to?(:retrieve)
-          raise ArgumentError, "shop storage must respond to :store and :retrieve"
-        end
+        raise ArgumentError, "shop storage does not have expected interface" unless expected_shop_interface?(shop_storage)
       end
 
       def user_storage=(storage)
         @user_storage = storage
+        return unless storage
```
emmack: Same as above.

rezaansyed: We are using a null object pattern to support nil cases and if @user_storage is reassigned to nil, then it wouldn't actually set the value of that. The setter should ideally set the logic to whatever is passed in and handling of nil cases are handled through the pattern.

lib/shopify_app/session/session_repository.rb
```diff
@@ -2,21 +2,23 @@ module ShopifyApp
...
+
     class << self
       def shop_storage=(storage)
         @shop_storage = storage
+        return unless storage
 
-        unless storage.nil? || self.shop_storage.respond_to?(:store) && self.shop_storage.respond_to?(:retrieve)
-          raise ArgumentError, "shop storage must respond to :store and :retrieve"
-        end
+        raise ArgumentError, "shop storage does not have expected interface" unless expected_shop_interface?(shop_storage)
```
ragalie: Re: Tim's comment about changing the interface; it does occur to me that this changes the interface (we now require an extra method). Maybe `expected_shop_interface?` (and the user one) should only expect the new `retrieve_` methods if the `allow_jwt_authentication` option is active.

tanema: @ragalie Actually I am about to ship https://github.com/Shopify/shopify_app/pull/925 which will remove that

The problem is that it loads the session constants at startup and this breaks zeitworks reloading functionality, thus raises a deprecation warning

ragalie: :+1:

ragalie: @rezaansyed we should wait to ship until #925 ships in this case and then rebase ours onto theirs.

lib/shopify_app/session/session_repository.rb
```diff
@@ -2,21 +2,23 @@ module ShopifyApp
   class SessionRepository
     class ConfigurationError < StandardError; end
 
+    EXPECTED_METHODS = [:store, :retrieve].freeze
+    EXPECTED_SHOP_STORAGE_METHODS = (EXPECTED_METHODS + [:retrieve_by_shopify_domain]).freeze
+    EXPECTED_USER_STORAGE_METHODS = (EXPECTED_METHODS + [:retrieve_by_shopify_user_id]).freeze
```
tanema: Yeah sorry as I just mentioned, I guess I blanked on this and forgot my other work but this will all have to be removed. I am removing it in https://github.com/Shopify/shopify_app/pull/925 as it conflicts with reloading

lib/shopify_app/session/session_repository.rb
rezaansyed: Sounds good. I'll wait for that PR to ship. 👍

lib/shopify_app/session/session_repository.rb
rezaansyed: @tanema just updated to account for your changes. Could you give it a quick look? I tophatted the changes.

PR #923, opened 2020-03-10
Split out return_to session setting to own method

slucaskim:
Hello, I would like to the function of setting the `return_to` session variable split out to it's own function so that I can call it from our app in contexts where we can't call the `redirect_to_login` method (JSON api call).

### DIFF
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -66,17 +66,21 @@ module ShopifyApp
       if request.xhr?
         head :unauthorized
       else
-        if request.get?
-          path = request.path
-          query = sanitized_params.to_query
-        else
-          referer = URI(request.referer || "/")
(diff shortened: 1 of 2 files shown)


tanema:
It would be best to have 1 or 2 tests to make sure this method returns what you expect it does

slucaskim:
> It would be best to have 1 or 2 tests to make sure this method returns what you expect it does

I was counting on these existing tests to cover for this behaviour (like integration tests) https://github.com/Shopify/shopify_app/blob/master/test/shopify_app/controller_concerns/login_protection_test.rb#L246-L268 Do you think that's enough @tanema

tanema:
If it is called separately then I think it is a good idea to have at least a couple tests to ensure that it works as expected and there was not some sort of state setup in the original calling method that you missed.

slucaskim:
I'm going to implement this in our app, on second thought I don't think it's correct at the gem level

PR #925, opened 2020-03-16
Fix for constantization issue in session store

### LINKED ISSUE #908: DEPRECATION WARNING: Initialization autoloaded the constant Shop.
Rails 6 introduced a new code loader `zeitwerk`. Zeitwerk recommends _not_ loading any constants (such as any class/module from `/app` directory) in `config/initializers`. [Rails Guides: Autoloading and Reloading Constants (Zeitwerk Mode)](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html)

ShopifyApp doesn't rely on anything from main app's `/app` directory, apart from one thing: checking whether `session_repository` implements `#store` and `#retrieve` methods. (
https://github.com/Shopify/shopify_app/blob/master/lib/shopify_app/session/session_repository.rb#L9-L11)

I believe the above check could be removed and the code could be simplified like this:

```diff
module ShopifyApp
  class SessionRepository
-    class ConfigurationError < StandardError; end
-
-    class << self
-      def storage=(storage)
-        @storage = storage
-
-        unless storage.nil? || self.storage.respond_to?(:store) && self.storage.respond_to?(:retrieve)
-          raise ArgumentError, "storage must respond to :store and :retrieve"
-        end
-      end
-
      def retrieve(id)
-        storage.retrieve(id)
+        ShopifyApp.configuration.session_repository.retrieve(id)
      end

      def store(session, *args)
-        storage.store(session, *args)
+        ShopifyApp.configuration.session_repository.store(session, *args)
      end
-
-      def storage
-        load_storage || raise(ConfigurationError.new("ShopifySessionRepository.storage is not configured!"))
-      end
-
-      private
-
-      def load_storage
-        return unless @storage
-        @storage.respond_to?(:safe_constantize) ? @storage.safe_constantize : @storage
-      end
-    end
  end
end
```

```diff
# lib/shopify_app/configuration.rb
module ShopifyApp
  class Configuration
    # ...

    def session_repository=(klass)
      @session_repository = klass
-      ShopifyApp::SessionRepository.storage = klass
    end
+
+    def session_repository
+      @session_repository.respond_to?(:constantize) ? @session_repository.constantize : @session_repository
+    end

    # ...
  end
end
```

The `SessionRepository` class is kept only for backwards compatibility, but it's actually just a proxy for calling `#session_repository` on `ShopifyApp.configuration`. `ShopifyApp.configuration` then _lazily_ calls `constantize` on its `@session_repository` instance variable (if necessary).

We could further optimize this by memoizing the result of `@session_repository.constantize`:

```rb
# lib/shopify_app/configuration.rb
module ShopifyApp
  class Configuration
    # ...

    def session_repository=(klass)
      @session_repository = klass
    end

    def session_repository
      @session_repository_class ||= @session_repository.respond_to?(:constantize) ? @session_repository.constantize : @session_repository
    end

    # ...
  end
end
```

Repro steps:

1. Ensure you have rails gem >= 6 installed
2. 
```sh
rails new shopping_zeitwerks
cd shopping_zeitwerks
echo "gem 'shopify_app'" >> Gemfile
bundle install
rails generate shopify_app
bundle exec rake
```

### Expected

No deprecation warnings shown

### Actual

> DEPRECATION WARNING: Initialization autoloaded the constant Shop.
>
> Being able to do this is deprecated. Autoloading during initialization is going
>
> Reloading does not reboot the application, and therefore code executed during
>
> This autoloaded constant has been unloaded.
>
> Please, check the "Autoloading and Reloading Constants" guide for solutions.
 (called from <main> at /Users/viktor/Developer/Ruby/shopping_zeitwerks/config/environment.rb:5)
Migrations are pending. To resolve this issue, run:
>
>        rails db:migrate RAILS_ENV=test
>

Here's a repo where each commit is one repro step from step 2 above:
https://github.com/vfonic/shopping_zeitwerks

I've also created a WIP PR where we can work on fixing this issue: #909

tanema:
fix #908 

The issue is that shopify_app loads the session repo class at configuration time to check it's interface. This however breaks reloading and logs a deprecation warning `DEPRECATION WARNING: Initialization autoloaded the constant Shop`

This will now only load the ShopStore or UserStore when they are needed.

@vfonic

### DIFF
--- a/lib/shopify_app/session/session_repository.rb
+++ b/lib/shopify_app/session/session_repository.rb
@@ -5,18 +5,10 @@ module ShopifyApp
     class << self
       def shop_storage=(storage)
         @shop_storage = storage
-
-        unless storage.nil? || self.shop_storage.respond_to?(:store) && self.shop_storage.respond_to?(:retrieve)
-          raise ArgumentError, "shop storage must respond to :store and :retrieve"
-        end
       end
 
       def user_storage=(storage)
         @user_storage = storage
-
-        unless storage.nil? || self.user_storage.respond_to?(:store) && self.user_storage.respond_to?(:retrieve)
-          raise ArgumentError, "user storage must respond to :store and :retrieve"
-        end
       end
 
       def retrieve_shop_session(id)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+13.0.1
+------
+* fix for deprecation warning while loading session storage at startup
+
 13.0.0
 ------
 + #887 Added concurrent user and shop session support (online/offline)
--- a/lib/shopify_app/configuration.rb
+++ b/lib/shopify_app/configuration.rb
@@ -14,8 +14,6 @@ module ShopifyApp
     attr_accessor :webhooks
     attr_accessor :scripttags
     attr_accessor :after_authenticate_job
-    attr_reader :shop_session_repository
-    attr_reader :user_session_repository
     attr_accessor :api_version
 
     # customise urls
@@ -51,13 +49,19 @@ module ShopifyApp
     end
 
     def user_session_repository=(klass)
-      @user_session_repository = klass
       ShopifyApp::SessionRepository.user_storage = klass
     end
 
+    def user_session_repository
+      ShopifyApp::SessionRepository.user_storage
+    end
+
     def shop_session_repository=(klass)
-      @shop_session_repository = klass
       ShopifyApp::SessionRepository.shop_storage = klass
+    end
+
+    def shop_session_repository
+      ShopifyApp::SessionRepository.shop_storage
     end
 
     def has_webhooks?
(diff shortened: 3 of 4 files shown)


lib/shopify_app/session/session_repository.rb
```diff
@@ -5,18 +5,10 @@ class ConfigurationError < StandardError; end
...
 
       def user_storage=(storage)
         @user_storage = storage
-
-        unless storage.nil? || self.user_storage.respond_to?(:store) && self.user_storage.respond_to?(:retrieve)
-          raise ArgumentError, "user storage must respond to :store and :retrieve"
-        end
       end
 
       def retrieve_shop_session(id)
```
vfonic: Kinda related to these changes, but also not. Seems like `.user_storage` is not raising `ConfigurationError` like `.shop_storage` does. I think these two methods should be consistent.

Maybe you could squeeze in that change?

tanema: The reason this is not done is that the app can run without the user session storage however is a lot less functional (especially in the absence of the user session) without shop storage.

vfonic: Ah so there's a case where someone would try to fetch `user_storage` even when it's not been set? I thought that might be an erroneous behavior.

Thanks for clarification!

PR #943, opened 2020-03-31
Update README.md - Migrating to 13.0.0 anchor

baconck:
Fix anchor on "Migrating to 13.0.0".
Should be `#migrating-to-1300`, instead of `#migrating-to-13`


tanema:
Thanks!

PR #947, opened 2020-04-08
Update changes to session

alisherry:
Updating 13.0.0 migration docs to highlight changes in `session[:shopify]`


tanema:
Thanks for the addition!

PR #948, opened 2020-04-08
Enabled rubocop

### LINKED ISSUE #813: Enable Rubocop
I think that since policial was dismantled, we have not enabled linting on this repo for a while. We need to get our code up to appropriate standards.

adfaris:
resolves #813 

add a rubocop gem to the project and run rubocop with the autocorrect option on and the before and after results are shown below. 

_Before:_
```127 files inspected, 382 offenses detected```

_After:_
```127 files inspected, 0 offenses detected```

Screen Shot 2020-04-10 at 11 03 48 AM

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,2 +1,7 @@
 inherit_gem:
     rubocop-shopify: rubocop.yml
+
+AllCops:
+  Exclude:
+    - 'lib/generators/**/*'
+
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,10 +1,13 @@
-inherit_from:
-  - https://shopify.github.io/ruby-style-guide/rubocop.yml
+inherit_gem:
+    rubocop-shopify: rubocop.yml
 
-LineLength:
-  Exclude:
-    - test/**/*
+# inherit_from:
+#   - https://shopify.github.io/ruby-style-guide/rubocop.yml
 
-Metrics/ClassLength:
-  Exclude:
-    - test/**/*
+# LineLength:
+#   Exclude:
+#     - test/**/*
+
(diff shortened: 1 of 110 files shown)

### CODE REVIEW

File: .rubocop.yml [hunk 1]
```diff
@@ -1,10 +1,17 @@
...
-  - https://shopify.github.io/ruby-style-guide/rubocop.yml
+inherit_gem:
+    rubocop-shopify: rubocop.yml
 
-LineLength:
+AllCops:
+  TargetRubyVersion: 2.4
   Exclude:
-    - test/**/*
+    - 'lib/generators/**/*'
```
vfonic, line 7: Hey @adfaris!

Sorry for coming late to the game.

Why are we excluding `'lib/generators/**/*'` here?

adfaris: Rubocop doesn't play with the template files with erb in them. https://github.com/Shopify/shopify_app/blob/f95aac7fc02412e0f6d57bcdf425cb2df1414644/lib/generators/shopify_app/add_webhook/templates/webhook_job.rb#L1

Perhaps we can be more specific here.

vfonic: Ah I see!

Yeah, maybe `'lib/generators/**/templates/*'` or similar? I haven't tried this out.

vfonic: Actually I think the real issue here is that those are not `.rb` files.

Those are `.rb.tt` files, I believe. Take a look at how rails calls them:
https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/controller/templates/controller.rb.tt

/cc @tanema

adfaris: I made these changes in this PR: 

https://github.com/Shopify/shopify_app/pull/959

PR #950, opened 2020-04-11
A webhook_verification_helper to run from your app

### LINKED ISSUE #621: Test helper for WebhookVerification
Including `ShopifyApp::WebhookVerification` in the webhook controller is recommended on the shopify_app readme. However, it is quite difficult to test it as it requires knowledge about how the authentication works. Even finding the source of the authentication error was difficult because the error was hidden under the request layer.

A test helper would make things a lot easier. I also searched through other repos in `Shopify/` which use `WebhookVerification` and I noticed that none of them have tests. In one case someone actually tried to test it but skipped the test entirely.
image

The test helper would essentially have to set the hmac header as shown here:
https://github.com/Shopify/shopify_app/blob/c7508ce341b1d54ae7a6fc9f514c4ab836f24e22/test/shopify_app/controller_concerns/webhook_verification_test.rb#L31

cc @Shopify/riskeng @erickadbay

adfaris:
resolves #621

### DIFF
--- a/lib/shopify_app/test_helpers/webhook_verification_helper.rb
+++ b/lib/shopify_app/test_helpers/webhook_verification_helper.rb
@@ -1,0 +1,17 @@
+module ShopifyApp
+  module TestHelpers
+    module WebhookVerificationHelper
+      def authorized_webhook_verification_headers!
+        params = { foo: 'anything' }
+        digest = OpenSSL::Digest.new('sha256')
+        secret = ShopifyApp.configuration.secret
+        valid_hmac = Base64.encode64(OpenSSL::HMAC.digest(digest, secret, params.to_query)).strip
+        @request.headers['HTTP_X_SHOPIFY_HMAC_SHA256'] = valid_hmac
+      end
+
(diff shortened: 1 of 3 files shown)


lib/shopify_app/test_helpers/webhook_verification_helper.rb
```diff
@@ -0,0 +1,17 @@
+module ShopifyApp
+  module TestHelpers
+    module WebhookVerificationHelper
+      def authorized_webhook_verification_headers!
+        params = { foo: 'anything' }
```
tanema, line 5: This should probably be a parameter passed in, in case they want to change the payload for any reason

adfaris:
> Thanks, this looks great. I just have one comment.

@tanema, the method now takes params hash as a parameter. Thanks for the feedback.

tanema:
Thanks so much!

tanema:
I will merge this soon. I am just waiting for a time where I want to do a release so that the README is not out of sync with the up to date version

PR #951, opened 2020-04-13
Implement JWT callback endpoint

ragalie:
## Goals

When we receive an OAuth callback with a JWT in the headers, we want to do two things:
1. Validate that the omniauth data matches the JWT data.
2. Persist the token information to the session store.

## Implementation

1. It's a bit of a pain to have two separate callback paths in omniauth. There is a callback_path option, and it does take a block, but I'm a little unclear about how it knows how to start the normal OAuth flow if you use a block for this option. I elected to incorporate the JWT flow into the existing callback path in a backwards-compatible fashion.
2. Omniauth tries to check the state param as part of the callback action and throws an error if the param doesn't exist or doesn't match the param in the session. We won't have a param in the session in this context. So we need to conditionally disable the state check, based on whether we have a valid JWT. This is safe, I believe, because the goal of the state check is to avoid CSRF, and we can trust the JWT as evidence that the user is doing this from the Shopify context.
  * To make this work, I moved the JWT logic into a Rack middleware. This has two benefits: (1) it means that other Rack middleware (such as omniauth) can make decisions based on the JWT data, and (2) it allows any Rack application to decode our JWT, even if they're not using Rails or the `shopify_app` controller stuff.

### DIFF
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -6,15 +6,22 @@ module ShopifyApp
     include ShopifyApp::LoginProtection
 
     def callback
-      if jwt_shopify_domain || jwt_shopify_user_id
-        jwt_callback
-        return
+      unless auth_hash
+        return respond_with_error
       end
 
-      if auth_hash
-        login_shop
+      if jwt_request? && !valid_jwt_auth?
+        return respond_with_error
+      end
 
-        if ShopifyApp::SessionRepository.user_storage.present? && user_session.blank?
+      if jwt_request?
+        set_shopify_session
+        head(:ok)
+      else
+        reset_session_options
+        set_shopify_session
+
+        if redirect_for_user_token?
           return redirect_to(login_url_with_optional_shop)
         end
 
@@ -23,30 +30,30 @@ module ShopifyApp
         perform_after_authenticate_job
 
         redirect_to(return_address)
+      end
+    end
+
+    private
+
+    def respond_with_error
+      if jwt_request?
+        head(:unauthorized)
       else
         flash[:error] = I18n.t('could_not_log_in')
         redirect_to(login_url_with_optional_shop)
       end
     end
 
-    private
+    def redirect_for_user_token?
+      ShopifyApp::SessionRepository.user_storage.present? && user_session.blank?
+    end
 
-    def jwt_callback
-      if valid_jwt_auth?
-        set_shopify_session
-        head(:ok)
-      else
-        head(:unauthorized)
-      end
+    def jwt_request?
+      jwt_shopify_domain || jwt_shopify_user_id
     end
 
     def valid_jwt_auth?
       auth_hash && jwt_shopify_domain == shop_name && jwt_shopify_user_id == associated_user_id
-    end
-
-    def login_shop
-      reset_session_options
-      set_shopify_session
     end
 
     def auth_hash
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -33,7 +33,7 @@ module ShopifyApp
 
     def jwt_callback
       if valid_jwt_auth?
-        create_user_session_from_jwt_callback
+        set_shopify_session
         head(:ok)
       else
         head(:unauthorized)
@@ -42,16 +42,6 @@ module ShopifyApp
 
     def valid_jwt_auth?
       auth_hash && jwt_shopify_domain == shop_name && jwt_shopify_user_id == associated_user_id
-    end
-
-    def create_user_session_from_jwt_callback
-      session = ShopifyAPI::Session.new(
-        domain: shop_name,
-        token: token,
-        api_version: ShopifyApp.configuration.api_version
-      )
-
-      ShopifyApp::SessionRepository.store_user_session(session, associated_user)
     end
 
     def login_shop
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -24,6 +24,22 @@ module ShopifyApp
       end
     end
 
+    def jwt_callback
+      if auth_hash && jwt_shopify_domain == shop_name && jwt_shopify_user_id == associated_user_id
+        session_store = ShopifyAPI::Session.new(
+          domain: shop_name,
+          token: token,
+          api_version: ShopifyApp.configuration.api_version
+        )
+
+        ShopifyApp::SessionRepository.store_user_session(session_store, associated_user)
+
+        head(:ok)
+      else
+        head(:unauthorized)
+      end
+    end
+
     private
 
     def login_shop
@@ -43,6 +59,10 @@ module ShopifyApp
       return unless auth_hash['extra'].present?
 
       auth_hash['extra']['associated_user']
+    end
+
+    def associated_user_id
+      associated_user && associated_user['id']
     end
 
     def token
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -6,6 +6,11 @@ module ShopifyApp
     include ShopifyApp::LoginProtection
 
     def callback
+      if jwt_shopify_domain || jwt_shopify_user_id
+        jwt_callback
+        return
+      end
+
       if auth_hash
         login_shop
 
@@ -24,6 +29,8 @@ module ShopifyApp
       end
     end
 
+    private
+
     def jwt_callback
       if valid_jwt_auth?
         create_user_session_from_jwt_callback
@@ -32,8 +39,6 @@ module ShopifyApp
         head(:unauthorized)
       end
     end
-
-    private
 
     def valid_jwt_auth?
       auth_hash && jwt_shopify_domain == shop_name && jwt_shopify_user_id == associated_user_id
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -25,15 +25,8 @@ module ShopifyApp
     end
 
     def jwt_callback
-      if auth_hash && jwt_shopify_domain == shop_name && jwt_shopify_user_id == associated_user_id
-        session_store = ShopifyAPI::Session.new(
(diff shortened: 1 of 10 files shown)


theundeadmonk:
Is there a reason we prefer not to override the old callback endpoint to handle both cases based on whether `enable_jwt_auth` is set in the config?

Would that be easier since we then wouldn't have to keep track of both urls.

ragalie:
I don't think we can do that because the existing callback endpoint has to run for the initial install (to get the shop token), even if `allow_jwt_authentication` is true.

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -24,8 +24,31 @@ def callback
       end
     end
 
+    def jwt_callback
+      if valid_jwt_auth?
+        create_user_session_from_jwt_callback
+        head(:ok)
```
theundeadmonk, line 30: Should we use head(:no_content) instead?

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -24,8 +24,31 @@ def callback
       end
     end
 
+    def jwt_callback
```
rezaansyed, line 27: Since this is only for user tokens, should we call this `user_callback`?

ragalie:
Ok, so there are some issues with the current approach :) Here's what I've discovered:
1. It's a bit of a pain to have two separate callback paths in omniauth. There _is_ a `callback_path` option, and it does take a block, but I'm a little unclear about how it knows how to start the normal OAuth flow if you use a block for this option. I'll probably switch to using `/callback` for both (as Adi originally suggested) since I'll be fighting less with omniauth.
2. Omniauth tries to check the `state` param as part of the callback action and throws an error if the param doesn't exist or doesn't match the param in the session. We won't have a param in the session in this context. So we need to conditionally disable the state check, based on whether we have a valid JWT. This is safe, I believe, because the goal of the state check is to avoid CSRF, and we can trust the JWT as evidence that the user is doing this from the Shopify context. cc @leleabhinav 
3. We'll need to update the logic in the omniauth gem to support three settings for what kind of tokens are expected: only shop, only user, shop or user. Right now it supports either shop or user, and we use a hacky session thing to work around this, and that isn't viable for our use case. Since it's quite common to use both shop and user tokens, we should just natively support this.

I'll work on this later today and we can see how it looks and if we like it.

ragalie:
Tophatting instructions:
1. Follow the general docs to update your app: https://docs.google.com/document/d/1nc6wdQcJ4JLACjN7oHrHTMNVv_pfWVw2QMqiUsIqmBo/edit#
2. Modify the `setup` block slightly by adding a byebug at the top and setting `per_user_permissions` to `true`.
3. Go to `https://whatever.ngrok.io/login` and start OAuth.
4. You'll hit the byebug once, continue.
5. You'll hit it again, and you'll see the callback URL is the path that it's stopped on. Kill the server.
6. Paste the URL into Postman (or curl), then get a valid JWT.
7. You can get a valid JWT by `dev c` in core, instantiating the `AccessAndAuth::Admin::SessionTokenGenerator` class with the right arguments for your shop/app and running `generate`.
8. You have <1 minute from when you generate to paste the JWT into the `Authorization` header (e.g. `Authorization: Bearer token-value`) and then send the request.
9. When you make the request you'll hit your byebug again. This time continue.
10. You'll get a 200 back and the token will have been inserted into the DB.

You can test permutations of this (e.g. no JWT, expired JWT, JWT for wrong shop, an invalid code, etc.) and you should get 4xx/5xxs.

app/controllers/shopify_app/callback_controller.rb
ragalie: I'd rather stick with OK because I think we're likely to add some content later, and I don't want apps to break when the status code switches from 204 => 200.

app/controllers/shopify_app/callback_controller.rb
ragalie: I think it's more tightly coupled to the `jwt` aspect than the `user` aspect, since the other callback endpoint can also handle user tokens.

theundeadmonk:
LGTM. 
I was wondering if we could perhaps reject the request in the middleware itself if the JWT is invalid. Not sure if we have the Shop or the User info there, but we could check for things like expiry, etc. Just a thought.

ragalie:
I'm reluctant to do that because some apps might use other bearer tokens in parts of their app. I think rejecting the request is more of a controller concern.

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -6,6 +6,11 @@ class CallbackController < ActionController::Base
     include ShopifyApp::LoginProtection
 
     def callback
+      if jwt_shopify_domain || jwt_shopify_user_id
+        jwt_callback
+        return
+      end
+
```
greatestape, line 13: Should this be `if`/`elsif`/`else` blocks rather than having the early return?

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -6,6 +6,11 @@ class CallbackController < ActionController::Base
...
       if auth_hash
         login_shop
```
greatestape, line 16: Should this block of code be extracted out for single-level-of-abstraction reasons? It looks like we're doing that with `jwt_callback` above, and it seems nice.

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -26,6 +31,29 @@ def callback
...
+      else
+        head(:unauthorized)
+      end
+    end
+
+    def valid_jwt_auth?
+      auth_hash && jwt_shopify_domain == shop_name && jwt_shopify_user_id == associated_user_id
+    end
+
+    def create_user_session_from_jwt_callback
```
greatestape, line 47: The relationship between this and the existing `set_shopify_session` is a little confusing to me. If we need the behaviours to be different, I wonder if we should factor out some of the common elements (like the actual `ShopifyAPI::Session.new` call) to highlight why these need to be different. Or, if they don't need to have different logic, can we just use `set_shopify_session` and not have this method?

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -45,6 +73,10 @@ def associated_user
       auth_hash['extra']['associated_user']
     end
 
+    def associated_user_id
+      associated_user && associated_user['id']
+    end
```
greatestape, line 78: This seems to be on the edge of superfluous, but doing safe navigation with nil/hash objects seems sort of gnarly. Here's what I found:

```ruby
associated_user&.[]('id')
```

If that feels like readable Ruby, you could put that in at line 43 and not need this method, but I dunno...
```suggestion
    end
```
```suggestion
    end
```
```suggestion
    end
```

lib/shopify_app/middleware/jwt_middleware.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    private
+
+    def call_next(env)
+      @app.call(env)
+    end
+
+    def authorization_header?(env)
+      env['HTTP_AUTHORIZATION']
+    end
```
greatestape, line 27: Should this be explicitly returning true/false? (or drop the `?` from the method name?)

lib/shopify_app/middleware/jwt_middleware.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def extract_token(env)
+      match = env['HTTP_AUTHORIZATION'].match(TOKEN_REGEX)
```
greatestape, line 30: If you changed `authorization_header?` to `authorization_header`, you could re-use it here...

Also, you could use that gnarly "one weird trick" I mentioned above to do this in one line:

```ruby
      authorization_header.match(TOKEN_REGEX)&.[](1)
```

(from the category of bad ideas that are so bad they're good?)

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -26,6 +31,29 @@ def callback
...
+      if valid_jwt_auth?
+        create_user_session_from_jwt_callback
+        head(:ok)
+      else
+        head(:unauthorized)
+      end
+    end
+
+    def valid_jwt_auth?
+      auth_hash && jwt_shopify_domain == shop_name && jwt_shopify_user_id == associated_user_id
```
greatestape, line 44: It doesn't look like we log anywhere else in the old callback code, but I still wonder if our future selves will be happier if we log why the jwt was invalid (or really, why the callback failed)

app/controllers/shopify_app/callback_controller.rb
ragalie: It's clever but not idiomatic :)

app/controllers/shopify_app/callback_controller.rb
ragalie: Both the JWT middleware and the ominauth middleware have pretty decent logging, so I think we can rely on those two sources to give enough debugging information (and supplement later on when/if we discover particular cases that aren't covered well).

app/controllers/shopify_app/callback_controller.rb
ragalie: I tried cleaning this up for clarity, take a look and see what you think!

lib/shopify_app/middleware/jwt_middleware.rb
```diff
@@ -0,0 +1,41 @@
...
+      env['HTTP_AUTHORIZATION']
+    end
+
+    def extract_token(env)
+      match = authorization_header(env).match(TOKEN_REGEX)
+      match && match[1]
+    end
+
+    def set_env_variables(token, env)
+      jwt = ShopifyApp::JWT.new(token)
```
leleabhinav, line 35: We should also validate `iat`, `nbf` and `exp`.

ragalie:
@leleabhinav are you OK with where this ended up?

leleabhinav:
> @leleabhinav are you OK with where this ended up?
👍 🚢

ragalie:
@Shopify/platform-dev-tools-education I'm ready for a review on this when you have a chance :) Thanks!

ragalie:
For sure, I'm thinking we could pretty easily make that move if/when we think it would be useful.

PR #952, opened 2020-04-13
Webhook jobs default behavior

### LINKED ISSUE #722: Default Behavior for Webhook Jobs
Hi,  [the current web hook template](https://github.com/Shopify/shopify_app/blob/dd363607a4f6c44f085fd44e6af01b58f2457213/lib/generators/shopify_app/add_webhook/templates/webhook_job.rb) uses `find_by` without checking the return value. It can return `nil` if the shop is not found. 

Since calling `find_by!` will raise an exception that likely results in a retry, how about changing it to the following: 

```rb
def perform(shop_domain:, webhook:)
  shop = Shop.find_by(shopify_domain: shop_domain)

  # If the shop cannot be found log the error and reutrn so that it's not retried
  if shop.nil?
    logger.error("#{self.class} failed: cannot find shop with domain '#{shop_domain}'")
    return
  end

  shop.with_shopify_session do
  end
end
```

Thoughts?

~(Seems that `ActiveJob::Base` should have/add a `logger` method...)~

adfaris:
resolves #722 

This PR implements the logic suggested by @sshaw. If a shop doesn't exist, will exit with an error message.

### DIFF
--- a/lib/generators/shopify_app/add_webhook/templates/webhook_job.rb
+++ b/lib/generators/shopify_app/add_webhook/templates/webhook_job.rb
@@ -2,6 +2,11 @@ class <%= @job_class_name %> < ActiveJob::Base
   def perform(shop_domain:, webhook:)
     shop = Shop.find_by(shopify_domain: shop_domain)
 
+    if shop.nil?
+      logger.error("#{self.class} failed: cannot find shop with domain '#{shop_domain}'")
+      return
+    end
+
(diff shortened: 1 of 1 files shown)


tanema:
If this is ready I can merge it.

adfaris:
> If this is ready I can merge it.

@tanema, it is ready you can merge it.

PR #958, opened 2020-04-15
Add shop parameter when redirecting after callback

theundeadmonk:
The callback controller redirects to the root url after the app has been installed.
Currently, the root url depends on the cookie to fetch the shop and load the app.

In a world without 3rd party cookies, the app has no way of knowing which shop the request is for. By having the Callback Controller add a Shop parameter to the root url, we can load the app and fetch the JWT for the particular shop.

### DIFF
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -205,11 +205,16 @@ module ShopifyApp
     end
 
     def return_address
+      return base_return_address unless ShopifyApp.configuration.allow_jwt_authentication
+      return_address_with_params(shop: current_shopify_domain)
+    end
+
+    def base_return_address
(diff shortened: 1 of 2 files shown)


ragalie:
I'm not sure I understand why this is necessary. When the app is redirected back to the root_url, then it will load the JS, the JS will request the JWT, and then the app will use the JWT to know which shop it is. Why do we need to add the shop param?

theundeadmonk:
We cannot create the App Bridge client without the shop url. Without the client, we cannot request the JWT.

ragalie:
Ok this makes sense to me conceptually. I'd maybe only do this when `allow_jwt_authentication` is on since it's not necessary for non-JWT and so it's easier to disentangle when/if we separate the OAuth flow more for JWT clients.

PR #959, opened 2020-04-16
Rubocop fixes

adfaris:
This is a continuation of PR #948.

Removed generator template exception from rubocop.yml file and rename files `webhook_job.rb` to `webhook_job.rb.tt` to both files that have erb inside the code. Both the test and rubocop are passing.

Screen Shot 2020-04-16 at 3 26 25 PM

cc @vfonic @tanema


lib/generators/shopify_app/shop_model/shop_model_generator.rb
```diff
@@ -30,9 +30,13 @@ def rails_migration_version
         Rails.version.match(/\d\.\d/)[0]
       end
 
-      # for generating a timestamp when using `create_migration`
-      def self.next_migration_number(dir)
```
vfonic, line 34: These methods, in this form, were not private. Can you please double check that this generator works as expected now?

lib/shopify_app/test_helpers/webhook_verification_helper.rb
```diff
@@ -8,7 +9,7 @@ def authorized_webhook_verification_headers!(params = {})
         @request.headers['HTTP_X_SHOPIFY_HMAC_SHA256'] = valid_hmac
       end
 
-      def unauthorized_webhook_verification_headers!(params = {})
+      def unauthorized_webhook_verification_headers!
```
vfonic, line 12: I can't find where is this method called from?

Maybe it sometimes passes `params` hash and sometimes it doesn't. In case it "sometimes does", this method will now raise error.

Can you please double check that this works as expected?

/cc @tanema how/where is this method called from?

adfaris: These methods were introduced in https://github.com/Shopify/shopify_app/pull/950 and are helper methods meant to be called from the main application tests.

lib/generators/shopify_app/shop_model/shop_model_generator.rb
adfaris: These methods are test covered and failed previously but pass currently.

tanema:
Since you renamed the templates, did you tophat the generators?

adfaris:
> Since you renamed the templates, did you tophat the generators?

Could you please elaborate on what you meant by tophat the generators? I can see that the `.tt` files are related to Thor, which rails uses internally for file templates.

tanema:
The `lib/generators/shopify_app/install/templates/shopify_app.rb.tt` is generated into a new project when the command `rails generate shopify_app:install` is run and the file `lib/generators/shopify_app/add_webhook/templates/webhook_job.rb.tt` is used when the command `rails g shopify_app:add_webhook` is called so those generators need to be run on your machine (tophatted) to ensure that they still generate valid files.

adfaris:
> The `lib/generators/shopify_app/install/templates/shopify_app.rb.tt` is generated into a new project when the command `rails generate shopify_app:install` is run and the file `lib/generators/shopify_app/add_webhook/templates/webhook_job.rb.tt` is used when the command `rails g shopify_app:add_webhook` is called so those generators need to be run on your machine (tophatted) to ensure that they still generate valid files.

yes, it has been tophatted and all works as expected. 

Screen Shot 2020-04-20 at 11 46 12 AM


Screen Shot 2020-04-20 at 11 37 42 AM

Screen Shot 2020-04-20 at 11 35 09 AM

PR #970, opened 2020-04-24
Packaging for release 13.1.0

rezaansyed:
This PR is a release for version 13.1.1, which includes the following changes:
* Adds the shop URL as a parameter when redirecting after the callback
* Bump minimum Ruby version to 2.4
* Bug fixes

### DIFF
--- a/lib/shopify_app/version.rb
+++ b/lib/shopify_app/version.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module ShopifyApp
-  VERSION = '13.0.1'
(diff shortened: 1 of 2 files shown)


lib/shopify_app/version.rb
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module ShopifyApp
-  VERSION = '13.0.1'
+  VERSION = '13.1.1'
```
tanema, line 3: 13.1.0

PR #978, opened 2020-04-28
Update shopify_app.rb to match the documentation.

### LINKED ISSUE #971: Default shop session strategy is now ShopifyApp::InMemoryShopSessionStore
As of 13.0.0, the default shop session strategy is now ShopifyApp::InMemoryShopSessionStore rather than "Shop".

https://github.com/Shopify/shopify_app/blob/master/lib/generators/shopify_app/install/templates/shopify_app.rb.tt#L11

This is a somewhat confusing (and undocumented) change. Should the default shop session strategy be changed back to the `Shop` model?

adfaris:
resolves #971 

Update the `shopify_app.rb.tt` from `  config.shop_session_repository = 'ShopifyApp::InMemoryShopSessionStore'
` to `  config.shop_session_repository = 'Shop'
`
also updates the readme to show the renaming of the `shopify_app.rb` to `shopify_app.rb.tt`

cc @tanema

### DIFF
--- a/README.md
+++ b/README.md
@@ -512,7 +512,7 @@ Migrating to 13.0.0
 Version 13.0.0 adds the ability to use both user and shop sessions, concurrently. This however involved a large
 change to how session stores work. Here are the steps to migrate to 13.x
 
-### Changes to `config/initializers/shopify_app.rb.tt`
+### Changes to `config/initializers/shopify_app.rb`
 - *REMOVE* `config.per_user_tokens = [true|false]` this is no longer needed
 - *CHANGE* `config.session_repository = 'Shop'` To  `config.shop_session_repository = 'Shop'`
 - *ADD (optional)*  User Session Storage `config.user_session_repository = 'User'`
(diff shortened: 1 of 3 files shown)


tanema:
The file output from the generators isn't named `shopify_app.rb.tt` is it?

adfaris:
> The file output from the generators isn't named `shopify_app.rb.tt` is it?

The output from the generators is an `.rb' file.

tanema:
Haha it looks like you figured out why I asked that. Just in case, I asked because you changes the README to include the `.tt` extension in the migration guide which was not correct. Thanks for the changes. Just one last question. Did you run the generators and the app runs? Is the Shop model generated as part of the install so the code runs after generation?

adfaris:
> Haha it looks like you figured out why I asked that. Just in case, I asked because you changes the README to include the `.tt` extension in the migration guide which was not correct. Thanks for the changes. Just one last question. Did you run the generators and the app runs? Is the Shop model generated as part of the install so the code runs after generation?

yes, the shop model is created. Please see the screenshots below: 

Screen Shot 2020-04-29 at 4 19 29 PM

Screen Shot 2020-04-29 at 4 02 55 PM
Screen Shot 2020-04-29 at 4 16 44 PM
Screen Shot 2020-04-29 at 4 17 02 PM

PR #982, opened 2020-05-04
Populate current_shopify_domain from the jwt

theundeadmonk:
With the new JWT based auth, we won't always have a session. In such cases, `current_shopify_domain` would return a nil value. 

This PR addresses that by fetching the domain from the JWT when it is not present in the session.

### DIFF
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -164,7 +164,10 @@ module ShopifyApp
     end
 
     def current_shopify_domain
-      shopify_domain = sanitized_shop_name || session[:shopify_domain]
+      shopify_domain = sanitized_shop_name ||
+        jwt_shopify_domain ||
+        session[:shopify_domain]
+
       return shopify_domain if shopify_domain.present?
 
       raise ShopifyDomainNotFound
--- a/test/shopify_app/controller_concerns/login_protection_test.rb
+++ b/test/shopify_app/controller_concerns/login_protection_test.rb
@@ -366,6 +366,20 @@ class LoginProtectionControllerTest < ActionController::TestCase
     end
   end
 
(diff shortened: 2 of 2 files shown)


lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -170,7 +170,9 @@ def fullpage_redirect_to(url)
     end
 
     def current_shopify_domain
-      shopify_domain = sanitized_shop_name || session[:shopify_domain]
+      shopify_domain = sanitized_shop_name ||
+        session[:shopify_domain] ||
+        jwt_shopify_domain
```
rezaansyed, line 175: Minor: can we check for `jwt_shopify_domain` before `session[:shopify_domain]` since we're doing the same approach for when getting the sessions?
```
shopify_domain = sanitized_shop_name ||
        jwt_shopify_domain ||
        session[:shopify_domain]

File: lib/shopify_app/controller_concerns/login_protection.rb [hunk 2]
```diff
@@ -170,7 +170,10 @@ def fullpage_redirect_to(url)
     end
 
     def current_shopify_domain
-      shopify_domain = sanitized_shop_name || session[:shopify_domain]
+      shopify_domain = sanitized_shop_name ||
+        jwt_shopify_domain ||
```
emmack, line 174: Should there be a test for this case?

theundeadmonk: It's a protected method. Do we wanna have test for protected methods @tanema ?

tanema: Ideally I would like something to prove that this works. We have tests for full-page redirect so maybe you can add an extra case for full-page redirect that uses this logic?

PR #986, opened 2020-05-14
Add session token concern

theundeadmonk:
This PR adds a concern that can be used in an unauthenticated controller.
It does the following
 - Adds a before action to check if the passed in shop domain is valid
 - Adds a before action to check if the passed in shop is installed
Both the before actions redirect to login in case of failure

### DIFF
--- a/app/controllers/concerns/shopify_app/require_known_shop.rb
+++ b/app/controllers/concerns/shopify_app/require_known_shop.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module RequireKnownShop
+    extend ActiveSupport::Concern
+
+    included do
+      before_action :check_shop_domain
+      before_action :check_shop_known
+    end
+
+    def current_shopify_domain
+      return if params[:shop].blank?
+      @shopify_domain ||= ShopifyApp::Utils.sanitize_shop_domain(params[:shop])
+    end
+
+    private
+
+    def check_shop_domain
+      redirect_to(ShopifyApp.configuration.login_url) unless current_shopify_domain
+    end
+
+    def check_shop_known
+      @shop = SessionRepository.retrieve_shop_session_by_shopify_domain(current_shopify_domain)
+      redirect_to(shop_login) unless @shop
+    end
+
+    def shop_login
+      url = URI(ShopifyApp.configuration.login_url)
+
+      url.query = URI.encode_www_form(
+        shop: params[:shop],
+        return_to: request.fullpath,
+      )
+
+      url.to_s
+    end
+  end
+end
--- a/test/controllers/concerns/require_known_shop_test.rb
+++ b/test/controllers/concerns/require_known_shop_test.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+class RequireKnownShopTest < ActionController::TestCase
+  class UnauthenticatedTestController < ActionController::Base
+    include ShopifyApp::RequireKnownShop
+
+    def index
+      render html: '<h1>Success</ h1>'
+    end
+  end
+
+  tests UnauthenticatedTestController
+
+  setup do
+    Rails.application.routes.draw do
+      get '/unauthenticated_test', to: 'require_known_shop_test/unauthenticated_test#index'
+    end
+  end
+
+  teardown do
+    Rails.application.reload_routes!
+  end
+
+  test 'redirects to login if no shop param is present' do
+    get :index
+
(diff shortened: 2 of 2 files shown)


app/controllers/concerns/shopify_app/domain_protection.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    private
+
+    def check_shop_domain
+      redirect_to(ShopifyApp.configuration.login_url) unless current_shopify_domain
+    end
+
+    def check_shop_known
+      @shop = Shop.find_by(shopify_domain: @shopify_domain)
```
ragalie, line 24: I think you should be accessing this data via `SessionRepository.retrieve_shop_session_by_domain`. How the fetch logic is implemented is a detail that's hidden behind the interface.

app/controllers/concerns/shopify_app/domain_protection.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module DomainProtection
```
greatestape, line 4: Just want to call attention to the name for other reviewers. We wanted to name this something that indicates its responsibility and reflects that it would be used instead of `LoginProtection`. Does this name do that?

app/controllers/concerns/shopify_app/domain_protection.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    private
+
+    def check_shop_domain
+      redirect_to(ShopifyApp.configuration.login_url) unless current_shopify_domain
+    end
+
+    def check_shop_known
+      @shop = Shop.find_by(shopify_domain: @shopify_domain)
+      redirect_to("#{ShopifyApp.configuration.login_url}?shop=#{@shopify_domain}") unless @shop
```
rezaansyed, line 25: Should we add a `return_to` parameter so that the redirect returns us back to the page it was last at and not always the homepage? Email folks added a PR fix for this yesterday: https://github.com/Shopify/email/pull/4746

theundeadmonk: @rezaansyed I think if the shop doesn't exist we'd want them to go through the login flow again, right?
Email does that because it is trying to masquerade as being part of Shopify and not an app so they redirect to the marketing section.

greatestape: I thought the return_to would be the path the user gets taken to within the app iframe after the login flow finishes.

theundeadmonk: Oh. Okay, I'll make the changes.

app/controllers/concerns/shopify_app/domain_protection.rb
```diff
@@ -0,0 +1,39 @@
...
+      redirect_to(shop_login) unless @shop
+    end
+
+    def shop_login
+      url = URI(ShopifyApp.configuration.login_url)
+
+      url.query = URI.encode_www_form(
+        shop: params[:shop],
+        return_to: request.fullpath,
+        )
```
ragalie, line 34: Weird indentation here

test/controllers/concerns/domain_protection_test.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    shopify_domain = 'shop1.myshopify.com'
+
+    get :index, params: { shop: shopify_domain }
+
+    redirect_url = URI('/login')
+    redirect_url.query = URI.encode_www_form(
+      shop: shopify_domain,
+      return_to: request.fullpath,
+      )
```
ragalie, line 49: Weird indentation

app/controllers/concerns/shopify_app/domain_protection.rb
ragalie: Maybe `RequireKnownShop` or something would be more clear what it does?

app/controllers/concerns/shopify_app/domain_protection.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    private
+
+    def check_shop_domain
+      redirect_to(ShopifyApp.configuration.login_url) unless current_shopify_domain
+    end
+
+    def check_shop_known
+      @shop = SessionRepository.retrieve_shop_session_by_shopify_domain(current_shopify_domain)
```
NabeelAhsen, line 24: This looks good now!

`retrieve_shop_session_by_domain` ~> `retrieve_shop_session_by_shopify_domain`

theundeadmonk:
🎩 on a test app

PR #989, opened 2020-05-15
Use compact and join to join referer.query and sanitized_params

Abdulwahaab710:


clayton-shopify:
It possible to add tests for this functionality?

PR #992, opened 2020-05-21
Added Payload Verification module

### LINKED ISSUE #926: Custom Carrier Verification
When we create an endpoint to receive checkout payloads for inspection, to generate shipping rates, we get the same HMAC security as a webhook. So in my controller I can just include the WebhookVerification module and carry on my merry way. 

It does seem weird though. Maybe there is a better way? I would not want to dupe the module and give it a name for Custom Carrier as that is not DRY. So since this verification happens the same for custom carrier payloads and webhooks, possibly app proxy, is there a plan to perhaps abstract this to just being "verification", and then it is clear what to include when needed?

adfaris:
This PR resolves #926 

Creates a payload_verifcation.rb and moves the `def hmac_valid?(data)` and `def shopify_hmac`. It then includes this module inside `webhook_verication.rb`.  All tests are passing. 

Screen Shot 2020-05-22 at 12 16 46 PM

cc @tanema

### DIFF
--- a/app/controllers/shopify_app/extension_verification_controller.rb
+++ b/app/controllers/shopify_app/extension_verification_controller.rb
@@ -2,19 +2,8 @@
 
 module ShopifyApp
   class ExtensionVerificationController < ActionController::Base
+    include ShopifyApp::PayloadVerification
     protect_from_forgery with: :null_session
     before_action :verify_request
-
-    private
-
-    def verify_request
-      hmac_header = request.headers['HTTP_X_SHOPIFY_HMAC_SHA256']
-      request_body = request.body.read
-      secret = ShopifyApp.configuration.secret
-      digest = OpenSSL::Digest.new('sha256')
-
-      expected_hmac = Base64.strict_encode64(OpenSSL::HMAC.digest(digest, secret, request_body))
-      head(:unauthorized) unless ActiveSupport::SecurityUtils.secure_compare(expected_hmac, hmac_header)
-    end
   end
 end
--- a/app/controllers/shopify_app/extension_verification_controller.rb
+++ b/app/controllers/shopify_app/extension_verification_controller.rb
@@ -5,5 +5,9 @@ module ShopifyApp
     include ShopifyApp::PayloadVerification
     protect_from_forgery with: :null_session
     before_action :verify_request
+
+    def verify_request
+      head(:unauthorized) unless hmac_valid?(request.body.read)
+    end
   end
 end
--- a/app/controllers/shopify_app/extension_verification_controller.rb
+++ b/app/controllers/shopify_app/extension_verification_controller.rb
@@ -6,6 +6,8 @@ module ShopifyApp
     protect_from_forgery with: :null_session
     before_action :verify_request
 
+    private
+
     def verify_request
       head(:unauthorized) unless hmac_valid?(request.body.read)
     end
--- a/lib/shopify_app/controller_concerns/payload_verification.rb
+++ b/lib/shopify_app/controller_concerns/payload_verification.rb
@@ -3,41 +3,21 @@ module ShopifyApp
   module PayloadVerification
     extend ActiveSupport::Concern
 
+    def shopify_hmac
+      request.headers['HTTP_X_SHOPIFY_HMAC_SHA256']
+    end
+
     def hmac_valid?(data)
       secrets = [ShopifyApp.configuration.secret, ShopifyApp.configuration.old_secret].reject(&:blank?)
 
       secrets.any? do |secret|
         digest = OpenSSL::Digest.new('sha256')
-
         ActiveSupport::SecurityUtils.secure_compare(
+          # request.headers['HTTP_X_SHOPIFY_HMAC_SHA256'],
           shopify_hmac,
           Base64.strict_encode64(OpenSSL::HMAC.digest(digest, secret, data))
         )
       end
-    end
-
-    def shopify_hmac
-      request.headers['HTTP_X_SHOPIFY_HMAC_SHA256']
-    end
-
-    def verify_request
-      hmac_header = request.headers['HTTP_X_SHOPIFY_HMAC_SHA256']
-      request_body = request.body.read
-      secret = ShopifyApp.configuration.secret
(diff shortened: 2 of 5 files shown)


tanema:
Hey @adfaris! This is awesome but there are more points in the codebase that this might be reused! There are 2 other places in the shopify_app gem that does the exact same form of verification (on different data) that could all be refactored into your PayloadVerification module

- [ShopifyApp::ExtensionVerificationController](https://github.com/Shopify/shopify_app/blob/master/app/controllers/shopify_app/extension_verification_controller.rb)
- [ShopifyApp::AppProxyVerification](https://github.com/Shopify/shopify_app/blob/master/lib/shopify_app/controller_concerns/app_proxy_verification.rb)

adfaris:
> Hey @adfaris! This is awesome but there are more points in the codebase that this might be reused! There are 2 other places in the shopify_app gem that does the exact same form of verification (on different data) that could all be refactored into your PayloadVerification module
> 
> * [ShopifyApp::ExtensionVerificationController](https://github.com/Shopify/shopify_app/blob/master/app/controllers/shopify_app/extension_verification_controller.rb)
> * [ShopifyApp::AppProxyVerification](https://github.com/Shopify/shopify_app/blob/master/lib/shopify_app/controller_concerns/app_proxy_verification.rb)

Hey @tanema,

I have updated the `payload_verification.rb` and included the module.

lib/shopify_app/controller_concerns/payload_verification.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    def calculated_signature(query_hash_without_signature)
+      sorted_params = query_hash_without_signature.collect { |k, v| "#{k}=#{Array(v).join(',')}" }.sort.join
+
+      OpenSSL::HMAC.hexdigest(
+        OpenSSL::Digest.new('sha256'),
+        ShopifyApp.configuration.secret,
+        sorted_params
+      )
+    end
```
tanema, line 41: There is a lot of repitition on this and I think it could be simpler. These two methods are really just the following.

```suggestion
    def verify_request
      head(:unauthorized) unless hmac_valid?(request.body.read) 
    end

    def calculated_signature(query_hash_without_signature)
      sorted_params = query_hash_without_signature.collect { |k, v| "#{k}=#{Array(v).join(',')}" }.sort.join
      hmac_valid?(sorted_params)
    end
```

The simplified versions of `verify_request` and `calculated_signature` could be put back where they came from because they don't need to be shared around and then other code can simply leverage the `hmac_valid?` method which means we have allowed for secret fallback in these two extra instances.

tanema:
Okay my apologies. It looks like the calculate signature in the AppProxyVerification cannot be re-written like that! 

AppProxyVerification generates it's own hmac so it might be harder to refactor it to re-use the verification code. Sorry about leading you down the wrong path! 🙇

lib/shopify_app/controller_concerns/payload_verification.rb
adfaris: @tanema, 

Thanks for the feedback, I was able to refactor the method `verify_request` and move the logic back in `app/controllers/shopify_app/extension_verification_controller.rb`. But the `calculated_signiture` does not work with the refactor for two reasons. The reasons are: 
1. Hmac in the `calculated_signiture` looks for app-level secrets, whereas the Hmac in `hmac_valid?` looks for secrets at the request level. 
2. Hmac in `calculated_signiture` returns a string, but `hmac_valid?` returns boolean.  Because of these reasons, the refactor didn't work. and I have put the code back to what it was before the refactor. Please let me know how you wanted to proceed.

adfaris:
> Okay my apologies. It looks like the calculate signature in the AppProxyVerification cannot be re-written like that!
> 
> AppProxyVerification generates it's own hmac so it might be harder to refactor it to re-use the verification code. Sorry about leading you down the wrong path! 🙇

I made similar discovery last night. Thanks for letting me know. I have restored that part of the code to what it was before the refactor.

lib/shopify_app/controller_concerns/payload_verification.rb
```diff
@@ -0,0 +1,23 @@
...
+      request.headers['HTTP_X_SHOPIFY_HMAC_SHA256']
+    end
+
+    def hmac_valid?(data)
+      secrets = [ShopifyApp.configuration.secret, ShopifyApp.configuration.old_secret].reject(&:blank?)
+
+      secrets.any? do |secret|
+        digest = OpenSSL::Digest.new('sha256')
+        ActiveSupport::SecurityUtils.secure_compare(
+          # request.headers['HTTP_X_SHOPIFY_HMAC_SHA256'],
```
andyw8, line 16: Was this accidentally left in?

adfaris: @andyw8 @tanema 

Commented code removed

app/controllers/shopify_app/extension_verification_controller.rb
```diff
@@ -2,19 +2,12 @@
 
 module ShopifyApp
   class ExtensionVerificationController < ActionController::Base
+    include ShopifyApp::PayloadVerification
     protect_from_forgery with: :null_session
     before_action :verify_request
 
-    private
```
andyw8, line 8: I think we should keep the `private` declaration in the extracted concern. In a controller, typically only the action methods should be public.

app/controllers/shopify_app/extension_verification_controller.rb
```diff
@@ -2,19 +2,14 @@
 
 module ShopifyApp
   class ExtensionVerificationController < ActionController::Base
+    include ShopifyApp::PayloadVerification
     protect_from_forgery with: :null_session
     before_action :verify_request
 
     private
```
andyw8, line 9: @adfaris Thanks for restoring here - I think we should also add the `private` declaration in `lib/shopify_app/controller_concerns/payload_verification.rb` so that `shopify_hmac` and `hmac_valid?` are private.

adfaris: @andyw8 Thanks for the feedback and those methods are private now.

andyw8:
Thanks @adfaris!

PR #1003, opened 2020-05-29
Update changelog for #926

andyw8:
I missed this in #926.


tanema:
If there is an appropriate place we should probably mention the Payload verification module in the README so that people can find it and use it

andyw8:
Yes, good point. Tracking in #1005.

I also opened #1004 to discuss a way to prevent us forgetting about this kind of thing.

Issue #1009, opened 2020-06-03
invalid_signature | Signature does not match, it may have been tampered with.


duleorlovic:
I created sample app using `rails generate shopify_app`.
You can download this commit https://github.com/trkin/shopify-sample-data-generator/commit/f1907582be984a2c0eb26e575503e2b4658b24f0

When I use brand new account it works fine (it lists products under `/admin/apps/:id`.

I have old account with two disabled organizations. 
When I update keys from that account
```
# https://partners.shopify.com/648793/apps/3898079
export SHOPIFY_API_KEY=a...
export SHOPIFY_API_SECRET_KEY=shpss_e...
export SHOP=dule-test.myshopify.com
export SCOPES=write_products,write_customers,write_draft_orders
```

after I install the app on the store it give me an error

```
OmniAuth::Strategies::OAuth2::CallbackError

invalid_signature | Signature does not match, it may have been tampered with.

Extracted source (around line #25):
    def raise_out!
      raise(env['omniauth.error'] || OmniAuth::Error.new(env['omniauth.error.type']))
    end

Rails.root: /home/orlovic/rails/shopify_sample_data_generator

Application Trace | Framework Trace | Full Trace
omniauth (1.9.1) lib/omniauth/failure_endpoint.rb:25:in `raise_out!'
```

so I assume this error is certainly because of the fact that account has some "disabled" organizations.

tanema:
This would have nothing to do with a inactive organization. This is specifically an error raised from `omniauth-shopify-oauth2` gem when the app gets a callback from shopify with the HMAC and the secure comparison between the payload and the HMAC does not match with your secret. 

This might mean that you have entered the wrong api_secret_key into your app configuration.

duleorlovic:
@tanema thank you for explanation. That was the reason.
But the cause of this confusion is that shopify api cli now uses SHOPIFY_API_SECRET_KEY instead of SHOPIFY_API_SECRET https://github.com/Shopify/shopify-app-cli/issues/387

I think there should be a consistency, so I created pull request https://github.com/Shopify/shopify_app/pull/1010 we can continue discussion there.

PR #1010, opened 2020-06-04
Use new env name SHOPIFY_API_SECRET_KEY #1009

duleorlovic:
Similar to https://github.com/Shopify/shopify-app-cli/issues/387

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ The default and install generators have been updated to source Shopify API key a
 
 ```
 SHOPIFY_API_KEY=your api key
-SHOPIFY_API_SECRET=your api secret
+SHOPIFY_API_SECRET_KEY=your api secret
 ```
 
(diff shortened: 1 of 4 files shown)


resistorsoftware:
This kind of change is so YAK Shaving nightmarish at times. Over the years we have seen so much naming clash. Why not group ALL of these issues into ONE and handle them ALL once and for all!

As an example, the App right now it littered with SHOP_DOMAIN, SHOPIFY_DOMAIN, SHOP_URL, and likely more. 

The debate on this must be fierce :)    Why not SHOPIFY_SECRET_KEY? Oh. Because it is for an API so we need to include API in there. Yaks nod in unison.

tanema:
The current ENV var that is used across all of our libraries is SHOPIFY_API_SECRET. Merging this would break how our CLI scaffolds applications. We will stick with `SHOPIFY_API_SECRET` for now

PR #1013, opened 2020-06-06
Simplify installation command with bundle add

Spone:
We can now use `bundle add` instead of manually `echo`ing the gem name to `Gemfile`

### DIFF
--- a/docs/Quickstart.md
+++ b/docs/Quickstart.md
@@ -40,7 +40,7 @@ $ heroku create name
 4. Add ShopifyApp to Gemfile
 ----------------------------
 
-Run these commands to add the `shopify_app` Gem to your app:
+Run this command to add the `shopify_app` Gem to your app:
 
 ```sh
 $ bundle add shopify_app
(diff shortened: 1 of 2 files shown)


Spone:
I signed the CLA but I'm not able to re-run the check.

andyw8:
Thanks @Spone. Let's also update `docs/Quickstart.md` to keep it consistent?

andyw8:
For reference, the `bundle add` command was added in 1.15.0.pre.1 (2017-04-16)

https://github.com/rubygems/bundler/blob/master/CHANGELOG.md

Spone:
@andyw8 done! :)

docs/Quickstart.md
```diff
@@ -43,8 +43,7 @@ $ heroku create name
 Run these commands to add the `shopify_app` Gem to your app:
```
andyw8, line 43: I just noticed that this should also be changed, i.e. "Run this command...".

Spone: Good catch! Fixed.

tanema:
Thanks!

PR #1014, opened 2020-06-09
Create optional unauthenticated home_controller generator

NabeelAhsen:
## What this PR does

> Relates to:
> https://github.com/Shopify/shopify/issues/250171 
> https://github.com/Shopify/shopify/issues/247743

This PR introduces the option to create an unauthenticated home_controller and an authenticated products_controller by specifying the `--with-session-token` option to home_controller generator. It also provides the capability of using JWT-based auth to all future Shopify apps by default if they choose to.

For context, this is the preferred method of creating an app using Shopify's [next-gen auth flow](https://vault.shopify.io/projects/10253). The home_controller needs to be an unauthenticated controller for an app to go to while waiting for authenticated session tokens.

#### Changes made in this PR

- [x] Create an optional authenticated products_controller
- [x] Create an optional unauthenticated home_controller
- [x] Enable JWT-based auth by default for future apps
- [x] Update README guides relating to this change
- [x] Add default javascript functionality showing how to make calls to protected resources.
- [x] Prefer to find the shop origin using `@shop_origin` before relying on `@current_shopify_session`.

### Current behaviour

1. By running `rails generate shopify_app` or `rails generate shopify_app:home_controller`:
    * The default authenticated home_controller is created along with the default view. 
    * A `root` route is also added to `routes.rb`.

### New behaviour
1. By running `rails generate shopify_app` or `rails generate shopify_app:home_controller`:
    * No new behaviour
2. By running `rails generate shopify_app --with-session-token` or `rails generate shopify_app:home_controller --with-session-token`:
    * An unauthenticated template of home_controller is generated 
    * An authenticated products_controller is generated to serve as an API
    * A new `index.html.erb` view is created that leverages session tokens to make a call
3. The embedded app `embedded_app.html.erb` view prefers to find the shop origin through `@shop_origin`. It defaults to `(@current_shopify_session.domain if @current_shopify_session)`

### :tophat: Testing this behaviour
1. Pull this branch
2. Create a new rails app: `rails new <app-name>`
3. Open the Gemfile of `<app_name>`  and add: 
```rb
gem 'shopify_app', path:'your/path/to/local/shopify_app'
gem 'dotenv-rails'
```
Run a `bundle install`

#### Default behaviour

1. Run `rails generate shopify_app` or `rails generate shopify_app:home_controller` followed by a `rails db:migrate`
      * Ensure that the default authenticated home_controller file is generated
      * Ensure that the default view is created. See gist: https://gist.github.com/NabeelAhsen/a0269613246eb62df5c105c85a8aa4d9
      * Ensure that the routes.rb file is updated accordingly:

```rb
  root :to => 'home#index'
end
```

#### Conditional `--with-session-token` behaviour

> Note this behaviour is a beta feature. Please message me or someone else on the Shopify app authorization team if you want to enrol your app in this beta feature for testing.

1. Run `rails generate shopify_app --with-session-token` or `rails generate shopify_app:home_controller --with-session-token` followed by a `rails db:migrate`
    * Ensure that an unauthenticated home_controller file is generated
    * Ensure that the default view is created. See gist: https://gist.github.com/NabeelAhsen/0a134edf76c85a5a425e47836b2013bf
    * Ensure that the routes.rb file is updated accordingly:

```rb
  root :to => 'home#index'
  get '/products', :to => 'products#index'
end
```
screencast 2020-06-23 15-57-18
_Loading a set of products using javascript_

screencast 2020-06-23 15-58-24
_When your shop doesn't have products_

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Generators
 
 ### Default Generator
 
-The default generator will run the `install`, `shop`, and `home_controller` generators. This is the recommended way to start a new app from scratch:
+The default generator will run the `install`, `shop`, `authenticated_controller`, and `home_controller` generators. This is the recommended way to start a new app from scratch:
 
 ```sh
 $ rails generate shopify_app
@@ -123,6 +123,16 @@ $ rails generate shopify_app:home_controller
 
 This generator creates an example home controller and view which fetches and displays products using the Shopify API.
 
+Options include:
+* __[beta]__ `with-session-token`: This flag generates an unauthenticated home_controller and a protected sample products_controller. It also creates a home view that leverages a session token to fetch products from your products_controller. Use this flag if you plan to build a single-page application or to secure your app using JWT session tokens (e.g. `--with-session-token` or `--with-session-token true`). 
+
+### Products Controller Generator
+
+```sh
+$ rails generate shopify_app:products_controller
+```
+
+This generator creates an example products API controller to fetch products using the Shopify API.
 
 ### App Proxy Controller Generator
 
--- a/lib/generators/shopify_app/home_controller/templates/index.html.erb
+++ b/lib/generators/shopify_app/home_controller/templates/index.html.erb
@@ -1,21 +1,71 @@
-<h2>Products</h2>
+<!DOCTYPE html>
+<html lang="<%= I18n.locale %>">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link
+      rel="stylesheet"
+      href="https://unpkg.com/@shopify/polaris@4.25.0/styles.min.css"
+      />
+<% if @with_session_token %>    <script>
+      document.addEventListener("DOMContentLoaded", async function() {
+        var SessionToken = window["app-bridge"].actions.SessionToken
+        var app = window.app;
 
-<ul>
-  <% @products.each do |product| %>
-    <li><%= link_to product.title, "https://#{@current_shopify_session.domain}/admin/products/#{product.id}", target: "_top" %></li>
-  <% end %>
-</ul>
+        app.dispatch(
+          SessionToken.request(),
+        );
 
-<hr>
+        // Save a session token for future requests
+        window.sessionToken = await new Promise((resolve) => {
+          app.subscribe(SessionToken.ActionType.RESPOND, (data) => {
+            resolve(data.sessionToken || "");
+          });
+        });
 
-<h2>Webhooks</h2>
+        var fetchProducts = function() {
+          var headers = new Headers({ "Authorization": "Bearer " + window.sessionToken });
+          return fetch("/products", { headers })
+            .then(response => response.json())
+            .then(data => {
+              var products = data.products;
 
-<% if @webhooks.present? %>
-  <ul>
-    <% @webhooks.each do |webhook| %>
-      <li><%= webhook.topic %> : <%= webhook.address %></li>
-    <% end %>
-  </ul>
-<% else %>
-  <p>This app has not created any webhooks for this Shop. Add webhooks to your ShopifyApp initializer if you need webhooks</p>
-<% end %>
+              if (products === undefined || products.length == 0) {
+                document.getElementById("products").innerHTML = "<br>No products to display.";
+              } else {
+                var list = "";
+                products.forEach((product) => {
+                  var link = `<a target="_top" href="https://<%%= @shop_origin %>/admin/products/${product.id}">`
+                  list += "<li>" + link + product.title + "</a></li>";
+                });
+                document.getElementById("products").innerHTML = "<ul>" + list + "</ul>";
+              }
+            });
+        }();
+      });
+    </script>
+<% end %>  </head>
+  <body>
+    <h2>Products</h2>
+<% if @with_session_token %>    <div id="products"><br>Loading...</div><% else %>
+    <ul>
+      <%% @products.each do |product| %>
+        <li><%%= link_to product.title, "https://#{@current_shopify_session.domain}/admin/products/#{product.id}", target: "_top" %></li>
+      <%% end %>
+    </ul>
+
+    <hr>
+
+    <h2>Webhooks</h2>
+
+    <%% if @webhooks.present? %>
+      <ul>
+        <%% @webhooks.each do |webhook| %>
+          <li><%%= webhook.topic %> : <%%= webhook.address %></li>
+        <%% end %>
+      </ul>
+    <%% else %>
+      <p>This app has not created any webhooks for this Shop. Add webhooks to your ShopifyApp initializer if you need webhooks</p>
+    <%% end %><% end %>
+  </body>
+</html>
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -115,14 +115,29 @@ The generator adds ShopifyApp and the required initializers to the host Rails ap
...
+```sh
+$ rails generate shopify_app:authenticated_controller
+```
+
+This generator creates an example authenticated controller to protect resources with.
+
+### Products Controller Generator
+
+```sh
+$ rails generate shopify_app:root_controller
```
andyw8, line 137: Should this say `products_controller`?

NabeelAhsen: Yes it should, good catch!

ragalie:
Can this roll out as default, given that the JWT stuff isn't available to everyone yet.

Do we want this to be default yet? `allow_jwt_authentication`

README.md
```diff
@@ -115,14 +115,29 @@ The generator adds ShopifyApp and the required initializers to the host Rails ap
...
-This generator creates an example home controller and view which fetches and displays products using the Shopify API.
+This generator creates an unauthenticated root controller your app can default to.
 
+### Authenticated Controller Generator
+
+```sh
+$ rails generate shopify_app:authenticated_controller
+```
+
+This generator creates an example authenticated controller to protect resources with.
```
greatestape, line 132: Is this controller an example, or is it something they'll likely want to keep as-is?

NabeelAhsen: Hmm good point. I think the word "example" can be omitted here.

greatestape:
I don't think we can ship this new behaviour as the default app generating behaviour until we've rolled out the JWT solution to everyone.

It looks like the generator code has some mechanism for receiving command line args. If we could toggle the generator behaviour base don an arg (like say `--unauthed-root`), that would be ideal. We could ship this earlier and start validating it with users.

I'm not as worried about setting `allow_jwt_authentication` to true by default, since it doesn't change any existing behaviours in the absence of jwts. It just enables support.

lib/generators/shopify_app/root_controller/templates/root_controller.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class RootController < ApplicationController
+  include ShopifyApp::EmbeddedApp
+  include ShopifyApp::RequireKnownShop
+
+  def index
+    @shop_orgin = current_shopify_domain
```
greatestape, line 8: Should we ship an example index.html.erb for this controller to render? It looks like as it stands this code won't work because it'll be expecting one.

NabeelAhsen:
@ragalie I think shipping out `allow_jwt_authentication` as a default won't be an issue, since it just proactively enables support. 

I agree with @greatestape that it'd be safer to be able to toggle/choose the generator behaviour. I can look into this in the meantime. I believe this was originally intended to be part of a larger v14 release.

@theundeadmonk thoughts?

greatestape:
When generating our new unauthed root controller, it'd be neat if we also generated a small sample bit of front-end code that demonstrated how to get the session token and make a back-end request. We could hit the products controller, for example.

If we do this, I suggest we do it with vanilla JS, with our only library dependency being app-bridge.

lib/generators/shopify_app/root_controller/templates/root_controller.rb
NabeelAhsen: I think the next step of this PR could be configuring an entire default view (including a sample SPA and index.html file to render it)

test/generators/home_controller_generator_test.rb
```diff
@@ -17,9 +17,19 @@ class HomeControllerGeneratorTest < Rails::Generators::TestCase
...
+    assert_file "app/controllers/home_controller.rb", /HomeController < AuthenticatedController/
+    assert_file "app/views/home/index.html.erb"
+    assert_no_file "app/controllers/products_controller.rb"
+  end
+
+  test "creates unauthenticated home controller with home index view given --with_session_token option" do
+    run_generator %w(--with-session-token)
+
+    assert_file "app/controllers/home_controller.rb", /HomeController < ApplicationController/
+    assert_file "app/views/home/index.html.erb"
```
mSitkovets, line 32: Don't you also need to assert that the `products_controller` file was created here?

NabeelAhsen: `run_generator` in the context of tests fails to run `generate("anything")` for any other generator called in `ShopifyApp::Generators::HomeControllerGenerator`. 

The reason this isn't an issue is because `generate("shopify_app:products_controller")` is currently tested in `test/generators/products_controller_generator_test.rb`.

NabeelAhsen:
> It doesn't feel good to ship this as a default generator for a couple reasons.
> - It still is pretty new
> - This is changes more than just this gem, it changes all of our documentation and tutorials.
> We need to have partner feedback on if this is something they want more of. We need data on how many SPAs are on our system. And if/when we do add this generator it either needs to be non-default or have a flag to generate the non-SPA version from before. We cannot just replace the app generator.

@tanema Finally got around to bringing this up to speed. 

The latest changes address your points by making this entirely optional. By specifying `--with-session-token` when generating an app or home_controller, developers can choose whether they want to generate a session-token authenticated SPA or not.

I feel this change (i.e making it optional and specifying that it is experimental) will allow us to gather the feedback needed to validate this decision when we open this feature up to the general public.

lib/generators/shopify_app/home_controller/templates/index.html.erb
```diff
@@ -1,21 +1,71 @@
...
-  </ul>
-<% else %>
-  <p>This app has not created any webhooks for this Shop. Add webhooks to your ShopifyApp initializer if you need webhooks</p>
-<% end %>
+              if (products === undefined || products.length == 0) {
+                document.getElementById("products").innerHTML = "<br>No products to display.";
+              } else {
+                var list = "";
+                products.forEach((product) => {
+                  var link = `<a target="_top" href="https://<%%= @shop_origin %>/admin/products/${product.id}">`
```
andyw8, line 38: Why do we use the double-% here?

NabeelAhsen: The `template` method in home_controller_generator.rb essentially does two passes when it copies this file to its final destination: 

- The first time, it replaces all instances of `<%= %>` with their respective control flows. In this example, it conditionally executives the `<% if with_session_token %> <% end %>` flow to determine whether we want a vanilla javascript solution of an erb solution containing products and webhooks. 
- For it's second pass, it replaces all instances of `<%% erb code %>` with `<% erb code %>` to properly template this file when it's generated.

NabeelAhsen: So if `<% if with_session_token %>` passes, we get the following final file: https://gist.github.com/NabeelAhsen/0a134edf76c85a5a425e47836b2013bf

Otherwise, we get the following final index.html.erb file: https://gist.github.com/NabeelAhsen/a0269613246eb62df5c105c85a8aa4d9

andyw8: Got it, thanks for the detailed explanation 👍

PR #1015, opened 2020-06-09
Packaging for release v13.4.0

rezaansyed:
13.4.0
------
* Skip CSRF protection if a valid signed JWT token is present as we trust Shopify to be the only source that can sign it securely.


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+13.4.0
+------
+* Skip CSRF protection if a valid signed JWT token is present as we trust Shopify to be the only source that can sign it securely.
```
andyw8, line 3: Not critical, but I think it's helpful to reference, or link to, the PR.

rezaansyed: Thanks Andy! Just added the reference to the PR

PR #1025, opened 2020-06-29
Add access_token_headers concern

NabeelAhsen:
## What this PR does

This PR introduces a new concern for the Authenticated shopify_app controller. ~~The `access_token_headers.rb` concern sets a custom `X-Shopify-Request-Auth-Code` response header after the controller action.~~ This is part of the Next Gen token-based auth strategy.

This response header is intended to be used as a signal by app clients to fetch an authorization code to exchange with Shopify for an access token. If this header is set to `true`, the app client uses App Bridge to initiate this exchange flow.

~~The standard default behaviour we expect is that every response has the custom `X-Shopify-Request-Auth-Code` response header set to `false`.~~ This concern provides a helper method to set it to `true` when app developers need to refresh their user access tokens (e.g when calls to Shopify API return as 401 Unauthorized).

See the following PRs that reference this behaviour and changes:

- https://github.com/Shopify/app-bridge/pull/1603
- https://github.com/Shopify/next-gen-auth-app-demo/pull/10/files#diff-e54f3c67ff47c764363e39152508a4ddL15-L16
    - An example app that would rely on this concern to communicate the need for a user access token refresh

---

- [ ] Rebase/squash fixup commits

### DIFF
--- a/app/controllers/concerns/shopify_app/access_token_headers.rb
+++ b/app/controllers/concerns/shopify_app/access_token_headers.rb
@@ -4,22 +4,10 @@ module ShopifyApp
   module AccessTokenHeaders
     extend ActiveSupport::Concern
 
-    included do
-      after_action(:set_access_token_headers)
-    end
-
-    ACCESS_TOKEN_REQUIRED_HEADER = 'X-Shopify-Request-Auth-Code'
+    ACCESS_TOKEN_REQUIRED_HEADER = 'X-Shopify-API-Request-Failure-Unauthorized'
 
     def signal_access_token_required
       response.set_header(ACCESS_TOKEN_REQUIRED_HEADER, true)
-    end
-
-    private
-
-    def set_access_token_headers
-      if response.get_header(ACCESS_TOKEN_REQUIRED_HEADER).nil?
-        response.set_header(ACCESS_TOKEN_REQUIRED_HEADER, false)
-      end
     end
   end
 end
--- a/app/controllers/concerns/shopify_app/access_token_headers.rb
+++ b/app/controllers/concerns/shopify_app/access_token_headers.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module AccessTokenHeaders
+    extend ActiveSupport::Concern
+
+    included do
+      after_action(:set_access_token_headers)
+    end
+
+    ACCESS_TOKEN_REQUIRED_HEADER = 'X-Shopify-Request-Auth-Code'
+
+    def signal_access_token_required
+      response.set_header(ACCESS_TOKEN_REQUIRED_HEADER, true)
+    end
+
+    private
+
+    def set_access_token_headers
+      if response.get_header(ACCESS_TOKEN_REQUIRED_HEADER).nil?
+        response.set_header(ACCESS_TOKEN_REQUIRED_HEADER, false)
+      end
+    end
+  end
+end
--- a/test/controllers/concerns/access_token_headers_test.rb
+++ b/test/controllers/concerns/access_token_headers_test.rb
@@ -4,12 +4,8 @@ class AccessTokenHeadersTest < ActionController::TestCase
   class TestController < ActionController::Base
     include ShopifyApp::AccessTokenHeaders
 
-    def without_header
-      response.set_header('Mock-Header', 'mock')
-      render(html: '<h1>Success</h1>')
-    end
-
     def with_header
+      response.set_header('Mock-Header', 'mock')
       signal_access_token_required
       render(html: '<h1>Success</h1>')
     end
@@ -20,7 +16,6 @@ class AccessTokenHeadersTest < ActionController::TestCase
   setup do
     Rails.application.routes.draw do
       get '/with-header', to: 'access_token_headers_test/test#with_header'
-      get '/without-header', to: 'access_token_headers_test/test#without_header'
     end
   end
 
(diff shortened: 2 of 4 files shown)


greatestape:
I'm not sure I follow the design vision of this middleware. It seems like this will ensure every response will have this header (which seems a little strange, why do we want to include it when it's false?), but I can't tell if it makes it any easier for consumers of the library.

Would it be possible to do something with `rescue_from` in the AuthenticatedController class? Assuming everyone's using `graphql-client`, won't all graphql requests that fail because of auth generate the same error? And if we want to let the user configure the type of error that triggers the behaviour, it seems like we could support that via code in the initializer.

test/controllers/concerns/access_token_headers_test.rb
```diff
@@ -0,0 +1,41 @@
...
+    get :without_header
+
+    assert_equal false, response.get_header('X-Shopify-Request-Auth-Code')
+  end
+
+  test 'does not change X-Shopify-Request-Auth-Code header if set' do
+    get :with_header
+
+    assert_equal true, response.get_header('X-Shopify-Request-Auth-Code')
+  end
```
theundeadmonk, line 40: Let's add another test to ensure that this concern doesn't overwrite any other headers that are set.

theundeadmonk:
@greatestape This is only part of the solution that @rezaansyed and @NabeelAhsen are proposing. The second part of this project would involve modifying `shopify_api` in order to capture any unauthorized errors and set the header. I think that we should always set the header, even if we set it to be false, to be consistent in our responses.

greatestape:
Ah got it. Will the solution in shopify_api depend on this new stuff in
shopify_app? Would it be nicer if this logic wasn’t split across two repos
(or three if you count app-bridge)?

test/controllers/concerns/access_token_headers_test.rb
NabeelAhsen: Added a test covering this behaviour.

NabeelAhsen:
@greatestape I think the motivation for this PR is to provide a simple (and timely) default solution that is extendable and non-constraining. In this case, we provide devs the option and tooling to handle any failed responses where they might require a user access token. It's likely the case that devs trying to access ShopifyAPI is the only such case where this failure might occur.

To Adi's point, I feel that a quick way to standardize this would be to keep this header set to false at all times, but change its value when needed.

Making this change on ShopifyAPI (or rescuing from the same error in the Authenticated controller) seems to be the next step -- but I'm not confident yet that these are the only cases a developer will ever need to consider when setting this header.

app/controllers/concerns/shopify_app/access_token_headers.rb
```diff
@@ -0,0 +1,25 @@
...
+    ACCESS_TOKEN_REQUIRED_HEADER = 'X-Shopify-Request-Auth-Code'
+
+    def signal_access_token_required
+      response.set_header(ACCESS_TOKEN_REQUIRED_HEADER, true)
+    end
+
+    private
+
+    def set_access_token_headers
+      unless response.get_header(ACCESS_TOKEN_REQUIRED_HEADER)
```
theundeadmonk, line 20: Small nitpick. Could we replace this with `response.get_header(ACCESS_TOKEN_REQUIRED_HEADER).nil?`? Otherwise LGTM

NabeelAhsen: Updated!

ragalie:
> I think that we should always set the header, even if we set it to be false, to be consistent in our responses.

I disagree. The general practice with headers is to only include them when they're communicating something useful. For example, we wouldn't send a `Cookie` header if there are no cookies, or expect a `Location` header if it's not a 3xx response.

ragalie:
> `X-Shopify-Request-Auth-Code` response header

In general I think it's better for error messages to tell the client what went wrong vs. how to resolve it. That makes it more future proof. So in this case, you might do something like `X-Shopify-API-Request-Failure` or, possibly better, `X-Shopify-API-Request-Failure-Reason`.

Along those lines, why did we decide to go with a header vs. communicating the error in the body, btw? I can imagine some reasons why, but I'm curious what yours are.

ragalie:
> The second part of this project would involve modifying shopify_api in order to capture any unauthorized errors and set the header.

This feels like a `shopify_app` responsibility vs. a `shopify_api` responsibility. `shopify_api` doesn't know anything about the app's request/response cycle; it just knows how to wrap HTTP calls to Shopify's API. Why/how would it be responsible for changing the response from the backend to the frontend?

I agree with Sam on this one; `shopify_api` already propagates an error when the request to the Shopify API fails (or, if it doesn't, we should add logic to raise an error) and then `shopify_app` is responsible for deciding how to handle that error in the request/response cycle.

NabeelAhsen:
> Along those lines, why did we decide to go with a header vs. communicating the error in the body, btw? I can imagine some reasons why, but I'm curious what yours are.

There is a discrepancy between how this error might be communicated between GraphQL and REST clients. A GraphQL client like Apollo might communicate this as a 2xx + data message letting you know your request was Unauthorized. 

A REST client would need to differentiate between:
* Whether the request to its backend was Unauthorized or 
* Whether the request to core from its backend was unauthorized.

I think a custom header distinguishes itself as a better signal to fetch an auth code.

tanema:
I think I ultimately just wonder if this should only be used in rails apps or can be used in any ruby app? This seems to be pretty tied to rails since it relies on ActiveSupport and `response.set_header` I think this belong here but I wonder why this is just a concern without any hook into the controller? Are you expecting developers to call this method themselves? Are we going to add tutorials on how to add this? Will this come in later work?

NabeelAhsen:
@tanema The goal is to provide developers the option to call this method when they need to. This is being captured in docs that we plan on publishing to shopify.dev later this week. 

FYI [here's some documentation surrounding this change](https://docs.google.com/document/d/1eGVUeYa8YZJgDTEjMg5DYGWyqa5RJmE2cNYQeKkYqQQ/edit#heading=h.g4w46rfoxvoa)

@theundeadmonk Setting this header only when needed simplifies the implementation of this concern significantly. Do you think this should still remain a concern?

theundeadmonk:
`The standard default behaviour we expect is that every response has the custom X-Shopify-Request-Auth-Code response header set to false`
I see this line was cancelled. Is there a reason we don't set the header to false on every action by default? I think that's where Tim's concerns lie.

NabeelAhsen:
> > *The standard default behaviour we expect is that every response has the custom X-Shopify-Request-Auth-Code response header set to false*
>
> I see this line was cancelled. Is there a reason we don't set the header to false on every action by default? I think that's > where Tim's concerns lie.

@theundeadmonk I think Mike brought up some good points above, that it's generally better to only include this header when we're communicating something useful. With this approach, we'd set it to true when we need it, like the `Location` header.

With this in mind, I'm wondering whether a concern is the appropriate place for this helper method. `def signal_access_token_required` is only responsible for setting a response header, and does not need before/after hooks.

Do you have any thoughts on this approach?

theundeadmonk:
We can't use hooks for this since the app developer is responsible for setting this header. I'm thinking this is more of a helper method for app devs. Also, if we figure out how to capture shopify_api's unauthorized responses, we can create a wrapper around it that calls this method.

NabeelAhsen:
Closing in favour of the following implementation: https://github.com/Shopify/shopify_app/pull/1032

PR #1031, opened 2020-07-20
Fix ShopifyDomainNotFound when JWT is enabled

DavidRinkoff:
### Problem
When an invalid shop domain is submitted to `SessionsController#create`, it is supposed to display an error like this:
Shopify App — Installation 2020-07-16 14-24-33

When `allow_jwt_authentication` is enabled, an error is raised instead:
Action Controller: Exception caught 2020-07-20 14-34-25

This bug is easily reproducible by setting up shopify_app with `allow_jwt_authentication` set to true, and then navigating to the login page and trying to install with an empty shop domain.

### Solution
This bug was introduced in #958 because when `allow_jwt_authentication` is enabled in the config, we now try to set the return address with the `current_shopify_domain` included as a parameter. `current_shopify_domain` will raise when there is no currently authenticated shop, so an error will be raised if this flow is triggered from the login page.

We should rescue from this error and fall back to the originally returned `base_return_address`.

### Template notes

cc @KevenLYF

### DIFF
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -204,6 +204,8 @@ module ShopifyApp
     def return_address
       return base_return_address unless ShopifyApp.configuration.allow_jwt_authentication
       return_address_with_params(shop: current_shopify_domain)
+    rescue ShopifyDomainNotFound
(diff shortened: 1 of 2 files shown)


greatestape:
Thanks for investigating this and making this PR!

PR #1036, opened 2020-07-30
Update README.md

enotikalt:
Removed superfluous word


andyw8:
@enotikalt the duplicate is correct here, enough it may look a bit awkward.

The first "in" is associated with dive, i.e. "dive in".

The second "in" is associated with the time, i.e. "in less than 5 minutes".

PR #1058, opened 2020-08-27
Create user session if one does not exist but was expected

NabeelAhsen:
> Fixes bug: https://github.com/Shopify/shopify/issues/257357

__This is most likely a breaking change according to how `def activate_shopify_session` behaves.__

### What this PR does

This PR introduces a check to ensure that if a `user_session` is expected rather than a `shop_session`, but no such `user_session` exists, we do one of two things:
    1. Signal to the user that an online access token is needed and return a `401 Unauthorized` __or__
    2. Redirect them to the login URL

---

These are the checks done to determine whether an app is expecting a `user_session` as opposed to a `shop_session`:
- `ShopifyApp.configuration.user_session_repository` is set as part of the app config  and
- `ShopifyApp::SessionRepository.user_storage` is initialized

### Things to watch out for

1. For apps that rely on App Bridge Authentication, a request is only considered an XHR request (`request.xhr?`) if the `HTTP_X_REQUESTED_WITH: XMLHttpRequest` header is set. Currently, none of our App Bridge utility fetch methods ensure we set this header. This is not guaranteed via ApolloClient requests either.
 2. `LoginProtection` is becoming this behemoth of a file that is overdue for a cleanup. It's likely that this is not the final resting place or behaviour of this change.

- [ ] Sync with the following teams regarding the acceptability of this change:
    - [x] Local Deliveries team
    - [ ] Dev Tools Education
    - [x] App Bridge


lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -17,6 +17,11 @@ class ShopifyDomainNotFound < StandardError; end
     ACCESS_TOKEN_REQUIRED_HEADER = 'X-Shopify-API-Request-Failure-Unauthorized'
 
     def activate_shopify_session
+      if user_session_expected? && user_session.blank?
+        signal_access_token_required
+        return redirect_to_login
```
NabeelAhsen, line 22: `redirect_to_login` expects the request to have the header `HTTP_X_REQUESTED_WITH: XMLHttpRequest` to indicate it's an xhr request.

This header is not set by out current utility fetch methods, nor is it set by apollo client when interacting with its graphql client.

To fix this, we can either:
1. Add that header as part of the `authenticatedFetch` definition or
2. Change `if request.xhr?` to something that encompasses a different type of request altogether

test/shopify_app/controller_concerns/login_protection_test.rb
```diff
@@ -240,8 +240,10 @@ class LoginProtectionControllerTest < ActionController::TestCase
     end
   end
 
-  test '#activate_shopify_session with Shopify session, clears top-level auth cookie' do
+  test '#activate_shopify_session with only shop_session, clears top-level auth cookie' do
     with_application_test_routes do
+      ShopifyApp::SessionRepository.user_storage = nil
```
ragalie, line 245: Do these need to get reset after the test completes?

lib/shopify_app/controller_concerns/login_protection.rb
theundeadmonk: I think we have to set the header because otherwise there's no way for the backend to differentiate between AJAX and regular requests.

NabeelAhsen: Should this be the default behaviour in App Bridge's `authenticatedFetch` method?

cc: @vividviolet

theundeadmonk: I'm fine with putting it in authenticatedFetch. We only want authenticatedFetch to be used for xhr requests anyway

test/shopify_app/controller_concerns/login_protection_test.rb
NabeelAhsen: I believe they're reset in the `setup` method of this test file

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -28,6 +33,10 @@ def activate_shopify_session
       end
     end
 
+    def user_session_expected?
```
rezaansyed, line 36: minor: can we make this a private method?

lib/shopify_app/controller_concerns/login_protection.rb
rezaansyed: @NabeelAhsen can you link to the app-bridge branch you have for adding the header for reference?

vividviolet: Makes sense to add it to App Bridge but do we not need a fallback in the case where they're not using any of the App Bridge utils?

NabeelAhsen: Here's the related PR: https://github.com/Shopify/app-bridge/pull/1652

My assumption, at least with the flow described here, is that the fallback case is: if they're using `xhr`/ajax requests, they're still responsible for handling any unauthorized responses that Shopify App returns.

lib/shopify_app/controller_concerns/login_protection.rb
NabeelAhsen: See latest changes!

paulomarg (approved):
LGTM

greatestape (approved):
LGTM!

PR #1060, opened 2020-09-02
Avoid using Script tags

smathieu:
The ITP screens currently uses an embedded script tag to share data between Rails & the browser. This breaks any app that has a content security policy without `unsafe-inline`, which is the recommended settings for all apps.

These changes were tested with an actual Shopify app.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "shopify_app",
-  "version": "14.2.0",
+  "version": "11.4.0",
```
paulomarg, line 3: I think there may have a been an issue with the version number here. In any case, we usually bump the version when we're going through the actual release process, so you shouldn't need to it in this PR.

paulomarg (commented):
It seems like we may need to replace `window` with `document.body.dataset` in `top_level_interaction.js` as well - we're trying to get the redirect URL but it ends up being undefined.

smathieu:
@ragalie Thanks for the review. I don't believe we should keep these variables around for a number of reasons.

1. Using ERB to generate JS is now a bit of an antipattern. There a large number of reasons for this, but in this case:
2. This would raise an error in any app with a content security policy preventing this.
3. Shopify's default security policy for new apps now prevents inline scripts.
4. Our app store policy won't let us publish apps with errors in the JS console. 

Finally, I think these are only used for the ITP screens, which I don't consider an extensible portion of this gem.

package.json
smathieu: Thanks. I'll revert this.

package.json
```diff
@@ -24,5 +24,6 @@
   },
   "scripts": {
     "test": "./node_modules/.bin/karma start --browsers ChromeHeadless --single-run"
-  }
+  },
+  "version": "11.4.0"
```
paulomarg, line 28: Sorry if I'm picking on this, but was this line here before? It seems like it may lead to some confusion when the package lock is updated.

ragalie:
Thanks @smathieu, all makes sense to me!

package.json
smathieu: sorry - I thought I reverted this... This is added by npm, most likely because I'm using an updated version of it. I'll revert it.

paulomarg (approved):

PR #1066, opened 2020-09-04
Update development dependencies (and remove NPM files)

paulomarg:
This PR addresses the `yargs-parser` security warning - that dependency bubbles up to `mocha-debug` using `mocha@*` which was locked to a previous version.

In order to manage the `mocha` version we actually end up using, we started to force it to a specific version.


andyw8:
It looks like `package-lock.json` is updating much more than just mocha (e.g. cliui from 4 to 5).

paulomarg:
> It looks like `package-lock.json` is updating much more than just mocha (e.g. cliui from 4 to 5).

Yes, I guess it all comes from updating `mocha` from 6.1.X to 6.2.3 - that was a dependency update on their side as well.

andyw8:
Ok yeah I see now how big the mocha dependency graph is.

attaboy:
After a little digging, I think it makes sense to add `mocha` as a dependency (since it is required by `karma-mocha`!). I do wonder if we need `mocha-debug` or `karma-mocha-debug`. The latter is marked as deprecated: https://www.npmjs.com/package/karma-mocha-debug

There's also a newer version of `karma-mocha` which might help avoid some of these dependency issues. In fact, it might be worth running through _all_ of the dev dependencies and seeing how trivial it is to update them to the latest as it looks like they're mostly untouched since this testing framework was first adopted, and I bet the underlying tests will still work fine.

attaboy:
Oh and as previously mentioned on Slack, we should remove `package-lock.json` unless anyone has a strong argument to keep it. yarn and npm use different dependency strategies, so these files will never be in sync. We should stick with yarn.

paulomarg:
Considering all dependencies here are for development only, I went ahead and ran `yarn upgrade --latest` to make sure we're up to date. The tests still worked after the upgrade, and I was able to fully remove `mocha-debug` and `karma-mocha-debug` which should be included in the newer versions.

attaboy (approved):
LGTM. It would be good to get someone who has actually worked on these tests to give this a thumbs up too!

paulomarg:
> LGTM. It would be good to get someone who has actually worked on these tests to give this a thumbs up too!

Maybe @tanema did or knows someone who did?

tanema (commented):
I have only made the most minor changes but looks good.

PR #1068, opened 2020-09-10
Allow apps to signal whenever there are insufficient scopes

rezaansyed:


theundeadmonk (approved):

test/shopify_app/controller_concerns/login_protection_test.rb
```diff
@@ -462,6 +468,54 @@ class LoginProtectionControllerTest < ActionController::TestCase
...
+      token: token,
+      api_version: '2020-01',
+      extra: { scopes: ShopifyApp.configuration.scope.split(',') }
+    )
+
+    ShopifyApp::SessionRepository.expects(:retrieve_shop_session_by_shopify_domain)
+      .with(domain).returns(expected_session)
+
+    with_application_test_routes do
+      request.env['jwt.shopify_domain'] = domain
```
paulomarg, line 511: Would it make sense to test that this happens both with and without JWT?

paulomarg (approved):
LGTM!

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -78,6 +78,10 @@ def token
       auth_hash['credentials']['token']
     end
 
+    def shop_scopes
+      auth_hash['extra']['scope'].split(',')
```
andyw8, line 82: Could `auth_hash['extra']['scope']` ever be nil here?

rezaansyed: Only if there are no scopes associated with a shopify_session (shop/user). I guess we can add a check for that as well

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -93,9 +97,13 @@ def set_shopify_session
       session[:shopify_user] = associated_user
       if session[:shopify_user].present?
         session[:shop_id] = nil if shop_session && shop_session.domain != shop_name
-        session[:user_id] = ShopifyApp::SessionRepository.store_user_session(session_store, associated_user)
+
+        ActiveRecord::Base.transaction do
+          session[:user_id] = ShopifyApp::SessionRepository.store_user_session(session_store, associated_user)
+          ShopifyApp::SessionRepository.update_scopes(shop_session, shop_scopes)
+        end
```
NabeelAhsen, line 104: Is there an easy way to test this functionality?

NabeelAhsen (approved):
LGTM

We should reflect the following migration instruction in appropriate documentations:

> Run `rails g shopify_app:shop_model --skip`

stevenmichaelthomas:
Can more context be added into this change to describe why it's needed, and what the impact could be without these changes? Thanks!

PR #1070, opened 2020-09-11
Create JwtAuthenticated concern

NabeelAhsen:

### DIFF
--- a/app/controllers/concerns/shopify_app/jwt_authenticated.rb
+++ b/app/controllers/concerns/shopify_app/jwt_authenticated.rb
@@ -1,0 +1,110 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module JwtAuthenticated
+    extend ActiveSupport::Concern
+
+    included do
+      include ShopifyApp::Localization
+      include ShopifyApp::CsrfProtection
+      include ShopifyApp::EmbeddedApp
+      before_action :login_again_if_different_user_or_shop
+      around_action :activate_shopify_session
+    end
+
+    ACCESS_TOKEN_REQUIRED_HEADER = 'X-Shopify-API-Request-Failure-Unauthorized'
+
+    def activate_shopify_session
+      if user_expected? && user.blank?
+        return signal_access_token_required
+      end
+
+      # TODO: Temporary until we have redirect-less Offline Access Token support
+      return redirect_to_login if current_shopify_session.blank?
+
+      begin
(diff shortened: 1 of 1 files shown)


app/controllers/concerns/shopify_app/jwt_authenticated.rb
```diff
@@ -0,0 +1,110 @@
...
+
+module ShopifyApp
+  module JwtAuthenticated
+    extend ActiveSupport::Concern
+
+    included do
+      include ShopifyApp::Localization
+      include ShopifyApp::CsrfProtection
+      include ShopifyApp::EmbeddedApp
+      before_action :login_again_if_different_user_or_shop
```
ragalie, line 11: I don't think this does anything for JWT.

app/controllers/concerns/shopify_app/jwt_authenticated.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module JwtAuthenticated
+    extend ActiveSupport::Concern
+
+    included do
+      include ShopifyApp::Localization
+      include ShopifyApp::CsrfProtection
```
ragalie, line 9: Do we use CSRF protection with JWT? I didn't think so.

ragalie:
I think you should start by defining the public interface of `JwtAuthenticated` (i.e. what methods does it expose, what behaviours does it automatically inject). Then figure out what you need to implement to make that happen.

That will lead you towards a clean and simple implementation.

NabeelAhsen:
> I think you should start by defining the public interface of `JwtAuthenticated` (i.e. what methods does it expose, what behaviours does it automatically inject). Then figure out what you need to implement to make that happen.
> 
> That will lead you towards a clean and simple implementation.

Makes sense! I think I started in the opposite direction -- By breaking down `login_protection.rb` instead.

app/controllers/concerns/shopify_app/jwt_authenticated.rb
```diff
@@ -0,0 +1,110 @@
...
+
+    def current_shopify_session
+      @current_shopify_session ||=
+        begin
+          shopify_session || user || shop
+        end
+    end
+
+    def shopify_session
+      # TODO: Configure a "true" Shopify session using the `sid` value from JWT
```
mSitkovets, line 41: I'll try doing this in my POC within the old login protection and see if it works

PR #1079, opened 2020-09-22
Ensure post authenticate jobs run after for jwt based callback requests

rezaansyed:
### What does this PR solve?
There is an issue noticed where the OAuth flow for getting an online (user) token fails to redirect back into the callback endpoint. The online token is eventually retrieved on failed requests to the Shopify API and we exchange an auth code for the online token via App Bridge Auth. However, when making the callback request to with session tokens (as JWTs) in the Authorization header, the post authenticate background jobs are not run, which is undesirable.

This PR ensures that the post authenticated background jobs are triggered after the callback request is made with session tokens in the request.

### DIFF
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -6,50 +6,77 @@ module ShopifyApp
     include ShopifyApp::LoginProtection
 
     def callback
-      unless auth_hash
-        return respond_with_error
+      return respond_with_error if invalid_request?
+
+      store_access_token_and_build_session
+
+      if start_user_token_flow?
+        return respond_with_user_token_flow
       end
 
-      if jwt_request? && !valid_jwt_auth?
-        Rails.logger.debug("[ShopifyApp::CallbackController] Invalid JWT auth detected.")
-        return respond_with_error
-      end
+      perform_post_authenticate_jobs
 
-      if jwt_request?
-        Rails.logger.debug("[ShopifyApp::CallbackController] JWT request detected. Setting shopify session...")
-        set_shopify_session
-        head(:ok)
-      else
-        Rails.logger.debug("[ShopifyApp::CallbackController] Not a JWT request. Resetting session options...")
-        reset_session_options
-        set_shopify_session
-
-        if redirect_for_user_token?
-          Rails.logger.debug("[ShopifyApp::CallbackController] Redirecting for user token...")
-          return redirect_to(login_url_with_optional_shop)
-        end
-
-        install_webhooks
-        install_scripttags
-        perform_after_authenticate_job
-
-        redirect_to(return_address)
-      end
+      respond_successfully
     end
 
     private
 
+    def respond_successfully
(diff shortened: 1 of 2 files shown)


ragalie:
No test for the `install_script_tags` action?

ragalie (approved):
LGTM.

This is easier to read than the previous implementation, which is good! Curious why you decided to refactor in this direction where every sub-method contains a conditional on JWT vs. splitting into two callback methods for JWT vs. normal browser? There's very little overlap other than persisting the token and running these callbacks.

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -6,38 +6,57 @@ class CallbackController < ActionController::Base
...
+      perform_post_authenticate_jobs
+
+      respond_successfully
+    end
+
+    private
+
+    def respond_successfully
       if jwt_request?
-        Rails.logger.debug("[ShopifyApp::CallbackController] JWT request detected. Setting shopify session...")
```
lstp, line 19: Are we intentionally not replacing the debug logs?

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -6,38 +6,57 @@ class CallbackController < ActionController::Base
...
-        end
+    def prevent_session_fixation
+      reset_session_options if native_browser_request?
+    end
 
-        install_webhooks
-        install_scripttags
-        perform_after_authenticate_job
+    def store_access_token_and_build_session
+      prevent_session_fixation if native_browser_request?
```
paulomarg, line 41: Do we need to check `native_browser_request?` here, if it's also checked in `prevent_session_fixation`?

paulomarg (approved):
LGTM!

rezaansyed:
> LGTM.
> 
> This is easier to read than the previous implementation, which is good! Curious why you decided to refactor in this direction where every sub-method contains a conditional on JWT vs. splitting into two callback methods for JWT vs. normal browser? There's very little overlap other than persisting the token and running these callbacks.

We had thought of splitting it into two but there seems to be a general pattern of processing the callback where we are doing the following:
```
respond_with_error if invalid_request?
store_access_token
perform_post_authenticate_jobs
respond_successfully
```
The only difference is that the other flow returns early when needing to get an online (user) token. So we had to leave the extra conditional. Also the reason why we have conditional splits between JWT and non-JWT is so that there is symmetry in where they diverge and that it's easier to see the boundaries for further refactoring in the future.

rezaansyed:
> No test for the `install_script_tags` action?

Nice catch! Just added :)

app/controllers/shopify_app/callback_controller.rb
rezaansyed: You're right! Just refactored it.

app/controllers/shopify_app/callback_controller.rb
rezaansyed: Just added the logs back

PR #1081, opened 2020-09-23
Packaging for release 14.4.3

rezaansyed:
14.4.3
------
* Fix to ensure post authenticate jobs are run after callback requests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+14.4.3
+------
+* Fix to ensure post authenticate jobs are run after callback requests [#1079](https://github.com/Shopify/shopify_app/pull/1079)
+
 14.4.2
 ------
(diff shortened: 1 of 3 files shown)


paulomarg (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+14.4.3
+------
+* Fix to ensure post authenticate jobs are run after callback requests
```
andyw8, line 3: Minor, but we can we add a link to the PR? It looks like we used to do this in the changelog (e.g. 13.3.0) but then tailed off.

paulomarg: Oh wow, that's a good catch, will keep it in mind as well when releasing this package :)

rezaansyed: Just added the link to it

PR #1085, opened 2020-09-28
Fix RuboCop failures

andyw8:
This repo isn't locked to a particular RuboCop version, but when running against the latest RuboCop there are some failures.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,6 +10,7 @@ AllCops:
 Style/MethodCallWithArgsParentheses:
   Exclude:
     - '**/Gemfile'
+    - 'test/**/*'
 
 Style/ClassAndModuleChildren:
   Exclude:
--- a/lib/generators/shopify_app/controllers/controllers_generator.rb
+++ b/lib/generators/shopify_app/controllers/controllers_generator.rb
@@ -8,7 +8,7 @@ module ShopifyApp
 
       def create_controllers
         controllers.each do |controller|
-          copy_file controller
+          copy_file(controller)
         end
       end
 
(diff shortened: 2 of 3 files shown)


lib/generators/shopify_app/controllers/controllers_generator.rb
```diff
@@ -8,7 +8,7 @@ class ControllersGenerator < Rails::Generators::Base
 
       def create_controllers
         controllers.each do |controller|
-          copy_file controller
+          copy_file(controller)
```
andyw8, line 11: Was failing the `MethodCallWithArgsParentheses` check.

.rubocop.yml
```diff
@@ -10,6 +10,7 @@ AllCops:
 Style/MethodCallWithArgsParentheses:
   Exclude:
     - '**/Gemfile'
+    - 'test/**/*'
```
andyw8, line 13: There were a number of failures on tests with DSL syntax such as:

```
test/shopify_app/controller_concerns/login_protection_test.rb:472:7: C: Style/MethodCallWithArgsParentheses: Use parentheses for method calls with arguments.
      assert_equal "{\"myshopifyUrl\":\"https://#{shop_domain}\",\"url\":\"#{example_url}\"}", ...
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
test/shopify_app/controller_concerns/login_protection_test.rb:480:9: C: Style/MethodCallWithArgsParentheses: Use parentheses for method calls with arguments.
        get '/' => 'login_protection#index'
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Since these typically don't have parenthesis, I disabled the rule for the whole `test` directory, although there is *some* risk that the `test` directory could contain other code with actual missing parenthesis.

paulomarg (approved):
LGTM. I think disabling the check for the test folder makes sense.

tanema (approved):

PR #1086, opened 2020-09-28
Disable logging args in ShopifyApp jobs

tanema:
Since shop tokens are passed into the parameters of the jobs and there has been a tendency for people to copy and paste these logs while debugging, it seems safest to redact the arguments from the jobs so that this will not happen accidentally in the future.

This will patch ActiveJob::Logging::LogSubscriber for rails < 6.1 and use `log_arguments = false` for rails >= 6.1

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -12,3 +12,8 @@ Authentication Issues
 A great deal of the issues surrounding this repo are around authenticating (installing) the generated app with Shopify.
 
 If you are experiencing issues with your app authenticating/installing the best way to get help fast is to create a repo with the minimal amount of code to demonstrate the issue and a clearly documented set of steps you took to arrive there. This will help us solve your problem quicker since we won't need to spend any time figuring out how to reproduce the bug. Please also include your operating system and browser.
+
+Security
+--------
+
+Please be certain to redact any private information from your logs or code snippets such as Api Keys, Api Secrets, and any authentication tokens such as shop_tokens.
--- a/lib/shopify_app/engine.rb
+++ b/lib/shopify_app/engine.rb
@@ -34,7 +34,7 @@ module ShopifyApp
 
     initializer "shopify_app.redact_job_params" do
       ActiveSupport.on_load(:active_job) do
-        if ActiveJob::Base.respond_to?(:log_arguments?, true)
+        if ActiveJob::Base.respond_to?(:log_arguments?)
           WebhooksManagerJob.log_arguments = false
           ScripttagsManagerJob.log_arguments = false
         elsif ActiveJob::Logging::LogSubscriber.private_method_defined?(:args_info)
--- a/lib/shopify_app/engine.rb
+++ b/lib/shopify_app/engine.rb
@@ -1,5 +1,15 @@
 # frozen_string_literal: true
 module ShopifyApp
+  module RedactJobParams
+    private
+
+    def args_info(job)
+      log_disabled_classes = %w(ShopifyApp::ScripttagsManagerJob ShopifyApp::WebhooksManagerJob)
+      return "" if log_disabled_classes.include?(job.class.name)
+      super
+    end
+  end
+
   class Engine < Rails::Engine
     engine_name 'shopify_app'
     isolate_namespace ShopifyApp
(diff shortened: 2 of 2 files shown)


lib/shopify_app/engine.rb
```diff
@@ -1,5 +1,15 @@
 # frozen_string_literal: true
 module ShopifyApp
+  module RedactJobParams
+    private
+
+    def args_info(job)
+      log_disabled_classes = %w(ShopifyApp::ScripttagsManagerJob ShopifyApp::WebhooksManagerJob)
```
paulomarg, line 7: Is there a way to define this behaviour in the classes themselves? It seems like it would be easy to forget to add new classes to the list.

paulomarg (approved):

lib/shopify_app/engine.rb
tanema: *maybe* we could add an attribute to check on each class but these jobs have been consistently  alone for 5 years now :D

paulomarg: Fair enough 😄 , I just thought that this isn't something we want to forget in the future.

.github/ISSUE_TEMPLATE.md
```diff
@@ -12,3 +12,8 @@ Authentication Issues
 A great deal of the issues surrounding this repo are around authenticating (installing) the generated app with Shopify.
 
 If you are experiencing issues with your app authenticating/installing the best way to get help fast is to create a repo with the minimal amount of code to demonstrate the issue and a clearly documented set of steps you took to arrive there. This will help us solve your problem quicker since we won't need to spend any time figuring out how to reproduce the bug. Please also include your operating system and browser.
+
+Security
+--------
+
+Please be certain to redact any private information from your logs or code snippets such as Api Keys, Api Secrets, and any authentication tokens such as shop_tokens.
```
andyw8, line 19:
```suggestion
Please be certain to redact any private information from your logs or code snippets such as API keys, API secrets, and any authentication tokens such as shop tokens.
```

lib/shopify_app/engine.rb
```diff
@@ -21,5 +31,16 @@ class Engine < Rails::Engine
         app.config.middleware.insert_after(ShopifyApp::SameSiteCookieMiddleware, ShopifyApp::JWTMiddleware)
       end
     end
+
+    initializer "shopify_app.redact_job_params" do
+      ActiveSupport.on_load(:active_job) do
+        if ActiveJob::Base.respond_to?(:log_arguments?, true)
```
andyw8, line 37: I had to look up what the boolean argument meant here. Perhaps we could add an explaining variable, e.g. `include_private_and_protected = true`?

lib/shopify_app/engine.rb
```diff
@@ -21,5 +31,16 @@ class Engine < Rails::Engine
         app.config.middleware.insert_after(ShopifyApp::SameSiteCookieMiddleware, ShopifyApp::JWTMiddleware)
       end
     end
+
+    initializer "shopify_app.redact_job_params" do
+      ActiveSupport.on_load(:active_job) do
```
andyw8, line 36: Do we need to consider older Rails apps that don't use ActiveJob?

tanema: ActiveJob was introduced in Rails 4.2 and this gem required 5.2

.github/ISSUE_TEMPLATE.md
tanema: I specifically used shop_token so that they would see that visually and look in their logs for it visually.

lib/shopify_app/engine.rb
tanema: I think I can just take it out, log_arguments is a public attribute, I forget why I felt the need to check private scope as well.

lstp (approved):

.github/ISSUE_TEMPLATE.md
andyw8: Maybe write it as "`shop_token`s"? (bit clunky, I know)

andyw8 (approved):

netwire88:
Hi @andyw8 , Do you know which version of Rails is required for `log_arguments` to work? When I added `self.log_arguments` to `class OrdersUpdatedJob < ActiveJob::Base`, it didn't work. I believe log_arguments requires rails 6.1, but shopify_app doesn't work on 6.1 yet.

PR #1087, opened 2020-09-29
Distinguishing between real users and Shopify Support through sessions

mSitkovets:
This PR creates a new `Session` table (for apps that use online tokens) using the new `sid` claim returned by the JWT payload so that we can make sure that we are accurately distinguishing between a merchant using an app and a Shopify employee impersonating a user using an app. I also include a method to clean up the `Session` table by deleting all expired sessions when a new `Session` entry is created.


app/controllers/shopify_app/callback_controller.rb
```diff
@@ -118,13 +122,17 @@ def set_shopify_session
...
-        api_version: ShopifyApp.configuration.api_version
+        api_version: ShopifyApp.configuration.api_version,
       )
 
       session[:shopify_user] = associated_user
       if session[:shopify_user].present?
         session[:shop_id] = nil if shop_session && shop_session.domain != shop_name
         session[:user_id] = ShopifyApp::SessionRepository.store_user_session(session_store, associated_user)
+        ShopifyApp::SessionRepository.clean_up_actual_sessions
+        if jwt_shopify_session_id
```
rezaansyed, line 133: What is our rollout strategy for this? As it stands, this would need to be a major version on `shopify_app`. Should this be opt-in for now, then opt-out? cc: @greatestape

lib/shopify_app/session/actual_session_storage.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+module ShopifyApp
+  module ActualSessionStorage
+    extend ActiveSupport::Concern
+    include ::ShopifyApp::SessionStorage
+
+    class_methods do
+      def store(auth_session, session_id, user, expires_at)
```
ragalie, line 8: I wonder if you can simplify the interface here.

`expires_at` is a property of the token, maybe it should live on the `ShopifyAPI::Session` object?

And `session_id` is a property of the user session, and we already have an object that represents that: `user`.

If you can simplify the interface in this way, then the interface will be the same as the interface for the user session storage strategy. Then this can just be another user session storage! That will simplify things a lot, and make migrating easier (since they can just switch to the new session storage when they're ready).

mSitkovets: I understand what you mean about grouping `expires_at` within the `ShopifyAPI::Session` object. 

For the `session_id` to be within `user` do you mean that I should make an object called `user` that wraps the `associated_user` and `jwt_shopify_session_id` within the `CallbackController`? 

And then send this information like `ShopifyApp::SessionRepository.store_actual_session(session_store, user)`

ragalie: I was kind of sloppily making two points at once:
1. We should try to implement all the aspects of the concept in the same object. This makes the system easier to maintain, because when all of the aspects of a concept are represented in a single object, we can often implement logic that only the object needs to understand the details about, and the rest of the system only needs to know the interface that the object provides. "Low coupling, high cohesion" is one way developers talk about this principle.
2. If we don't have to add a new concept, then we shouldn't! The easiest code to maintain is the code we never write. I was suggesting that we can probably make the interface for the existing user session storage the same as the interface for the "actual session" that you're proposing. Then we don't need the new "actual session" interface!

I'm less sure that 2. is true. For example, we usually, but not always, have a session ID in the "normal" OAuth, so there's a way to populate that value in the interface. It's this value: https://github.com/Shopify/shopify_app/blob/master/app/controllers/shopify_app/callback_controller.rb#L133

What do we do when we don't have that value? If our class permits the value not to exist, then we have to handle that case. You might decide that it's not acceptable to be missing the session ID in the "actual session" case, in which case we can't support the same interface. I'd encourage you to think about a fallback when we don't have a session value though (and maybe logging the event), because I think it will simplify the maintainability quite a bit and be easier for partners to understand as they migrate.

Re: how to handle `session_id` and the `user` object, I wouldn't wrap it, I'd define a real class that encapsulates the information about the user session. So maybe something like `ShopifyApp::UserSession`. Then you can construct it using data from these different sources.

tanema:
Why are we creating a new session model? We already have 2 session management concerns that users can use to implement their own models. UserSessionStorage and ShopSessionStorage. This seems to create a hard link between both and somewhat makes it so that having both is the default behaviour. I am not sure if this is the best direction to go in to keep this library flexible.

PR #1089, opened 2020-10-07
Switch boolean HTTP header value to string

robwalker367:
## What
Boolean header value for the custom HTTP header `X-Shopify-API-Request-Failure-Unauthorized` was giving the following error in `cardreader-giveaway` app.
Screen Shot 2020-10-07 at 10 08 51 AM
This was because the webrick handler in the rack gem iterates over the key/values of the HTTP headers in the response and calls `.split(...)` on the values. The value of the `X-Shopify-API-Request-Failure-Unauthorized` is a `TrueClass` instance, thus the above error.

This PR switches the boolean header value to a string. 

### DIFF
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -100,7 +100,7 @@ module ShopifyApp
     end
 
     def signal_access_token_required
-      response.set_header(ACCESS_TOKEN_REQUIRED_HEADER, true)
+      response.set_header(ACCESS_TOKEN_REQUIRED_HEADER, "true")
     end
 
     protected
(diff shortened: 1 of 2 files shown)


slucaskim (approved):

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -100,7 +100,7 @@ def login_again_if_different_user_or_shop
     end
 
     def signal_access_token_required
-      response.set_header(ACCESS_TOKEN_REQUIRED_HEADER, true)
```
andyw8, line 103: I'd never heard of this convention for booleans before, perhaps we could use a constant to help explain it, e.g. `STRUCTURED_FIELD_TRUE = "?1"` ?

robwalker367: Good catch. I looked into this in more detail and found that in some Shopify repos (e.g., `Shopify/app-bridge` in `user-authorized-fetch.ts`) we rely on the `X-Shopify-API-Request-Failure-Unauthorized` header having a string value of `"true"`, not the convention from the HTTP working group spec of `"?1"` and `"?0"`. So I think it's best to simply use a string value of "true" here.

rezaansyed (approved):

andyw8 (approved):

andyw8 (approved):

PR #1097, opened 2020-10-19
Enforce presence of application key and secret during initialization

tjoyal:
Thoughts on making it explicit when the key is not set when it is expected for it to be present so it fail fast?

### DIFF
--- a/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
+++ b/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
@@ -1,7 +1,7 @@
 ShopifyApp.configure do |config|
   config.application_name = "<%= @application_name %>"
-  config.api_key = ENV.fetch('SHOPIFY_API_KEY')
-  config.secret = ENV.fetch('SHOPIFY_API_SECRET')
+  config.api_key = ENV.fetch('SHOPIFY_API_KEY', '').presence || raise('Missing SHOPIFY_API_KEY')
+  config.secret = ENV.fetch('SHOPIFY_API_SECRET', '').presence || raise('Missing SHOPIFY_API_SECRET')
   config.old_secret = "<%= @old_secret %>"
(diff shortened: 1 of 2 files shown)


andyw8:
I support this.

tjoyal:
I ended up with this in my app, not sure if we prefer that:
```rb
  config.api_key = ENV.fetch('SHOPIFY_API_KEY', '').presence || raise("Missing SHOPIFY_API_KEY")
  config.secret = ENV.fetch('SHOPIFY_API_SECRET', '').presence || raise("Missing SHOPIFY_API_SECRET")
```

I use `presence` as I use dotenv and have this so the keys are git versioned but not the secrets:
```
# .env
SHOPIFY_API_KEY=
SHOPIFY_API_SECRET=

# .env.local
SHOPIFY_API_KEY=the-actual-key
SHOPIFY_API_SECRET=the-actual-key

# .gitignore
.env.local
```

paulomarg:
I actually like throwing a more specific error if one of the keys isn't present, it can be a bit of a pain point when apps start up without the necessary envs and fail later on for seemingly random reasons.

tanema (approved):

andyw8:
@tjoyal I think this may be causing some problems: https://github.com/Shopify/shopify-app-cli/issues/1006

tjoyal:
@andyw8 To make sure we are on the same page, I do not think we are challenging here the fact that is raises but rather the low quality error message? Seems like a fair ask.

andyw8:
@tjoyal The error message is a secondary concern. But I think enforcing that the env vars are always present might have impacted how the generator runs. We're still investigating, no need to take action yet.

andyw8:
https://github.com/Shopify/shopify-app-cli/pull/1026

andyw8:
https://github.com/Shopify/shopify_app/pull/1144

PR #1098, opened 2020-10-19
Update ngrok template

tjoyal:
Originally added here: https://github.com/Shopify/shopify_app/pull/802

It took me a long while to understand that \h in ruby is not the same as some other languages.....

> /\h/ - A hexdigit character ([0-9a-fA-F])
https://ruby-doc.org/core-2.7.1/Regexp.html

That being said, paid ngrok account and use more characters in their premium domains.

Also adding some `\.` to make sure we treat them as dots and not wildcards.

### DIFF
--- a/test/generators/install_generator_test.rb
+++ b/test/generators/install_generator_test.rb
@@ -90,7 +90,7 @@ class InstallGeneratorTest < Rails::Generators::TestCase
   test "adds host config to development.rb" do
     run_generator
     assert_file "config/environments/development.rb" do |config|
-      assert_match "config.hosts = (config.hosts rescue []) << /\\h+.ngrok.io/", config
+      assert_match "config.hosts = (config.hosts rescue []) << /\\w+\\.ngrok\\.io/", config
     end
   end
 end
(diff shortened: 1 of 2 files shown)


test/generators/install_generator_test.rb
```diff
@@ -90,7 +90,7 @@ class InstallGeneratorTest < Rails::Generators::TestCase
   test "adds host config to development.rb" do
     run_generator
     assert_file "config/environments/development.rb" do |config|
-      assert_match "config.hosts = (config.hosts rescue []) << /\\h+.ngrok.io/", config
+      assert_match "config.hosts = (config.hosts rescue []) << /\\w+\\.ngrok\\.io/", config
```
tjoyal, line 93: Would it be easier to read if the test was like that?
```rb
assert_match 'config.hosts = (config.hosts rescue []) << /\w+\.ngrok\.io/', config
```

paulomarg: Looks the same to me, I think it's fine as is.

paulomarg (approved):

tanema (approved):

PR #1102, opened 2020-11-05
Update HTML files to use new design language styles

### LINKED ISSUE #1099: Update relevant HTML + CSS to new admin design language
The new design language has now rolled out across the Shopify admin. We need to update the relevant HTML files to reflect the new styling.

eg.
Screen Shot 2020-10-23 at 12 43 08 PM

jgray7019:
Closes #1099 

This PR updates all styles and HTML used in the `html.erb` files to adhere to the new design language that has shipped across the store admin.
[Design Language Figma](https://www.figma.com/file/4dAAt5iFPSaxUKiYVKrkYj/Admin-Design-Language-(Web)%3A-Components?node-id=2474%3A62) for reference.

## Tophatting
1. You will need to set up a [test app using the `shopify_app` gem](https://github.com/Shopify/shopify_app#becoming-a-shopify-app-developer).
2. Checkout this branch of the gem. Re-start your server in the repo for your test app.
3. Run the local version of your app in any browser (`bundle exec rails server`) to view the install app page (`new.html.erb`). This page was not totally following Polaris styles before so I have updated the HTML and styles to better follow Polaris guidelines.
Screen Shot 2020-11-04 at 6 03 03 PM

4. Using Safari, install your app on a development store to be redirected to the `top_level_interaction.html.erb` page.
Screen Shot 2020-11-04 at 6 29 08 PM

5. Click `Continue` and you will see the `request_storage_access.html.erb` page.
Screen Shot 2020-11-04 at 6 29 52 PM

6. Change you Safari version to `12.0.0` to see the `enable_cookies.html.erb` page. You can do this by clicking on `Develop > User Agent > Other`
Screen Shot 2020-11-05 at 11 18 59 AM
And then changing the version to `12.0.0`
Screen Shot 2020-11-05 at 11 19 17 AM
This should now show you the correct page:
Screen Shot 2020-11-04 at 6 30 28 PM


paulomarg (approved):
🎩 ed, and from the code / app perspective it seems good to me!

In any case, I'm re-adding the team review if someone wants to take a closer look at the style itself, but as far as I can tell the app itself is working just fine.

app/views/shopify_app/partials/_button_styles.html.erb
```diff
@@ -15,12 +14,14 @@
     min-width:3.6rem;
     margin:0;
     padding:0.7rem 1.6rem;
-    background:linear-gradient(to bottom, white, #f9fafb);
-    border:1px solid #c4cdd5;
-    box-shadow:0 1px 0 0 rgba(22, 29, 37, 0.05);
-    border-radius:3px;
+    background-color: #ffffff;
```
tiffanytse, line 17: Are there variables we can use for colors? Or does this all need to be hardcoded?

app/views/shopify_app/partials/_button_styles.html.erb
```diff
@@ -29,30 +30,44 @@
     user-select:none;
     text-decoration:none;
     transition-property:background, border, box-shadow;
-    transition-duration:200ms;
+    transition-duration:.1s;
```
tiffanytse, line 33: might be easier to read as `100ms` instead; since someone could parse this as 1 second

app/views/shopify_app/sessions/new.html.erb
```diff
@@ -5,119 +5,48 @@
...
-      border-color: #3f4eae;
-    }
-
-    button.marketing-button:focus {
-      box-shadow: 0 0 0.1875em 0.1875em rgba(94,110,191,0.5);
-      background-color: #223274;
-      color: #fff;
-    }
-  </style>
+  <%= render 'shopify_app/partials/layout_styles' %>
```
tiffanytse, line 8: 👏

app/views/shopify_app/sessions/new.html.erb
```diff
@@ -5,119 +5,48 @@
...
+                    <input id="shop" name="shop" type="text" autofocus="autofocus" placeholder="example.myshopify.com" class="Polaris-TextField__Input">
+                    <div class="Polaris-TextField__Backdrop"></div>
+                  </div>
+                  <% if flash[:error] %>
+                    <div class="Polaris-InlineError"><%= flash[:error] %></div>
+                  <% end %>
+                  <button type="submit" class="Polaris-Button Polaris-Button--primary Polaris-Button--sizeLarge">
+                    <span class="Polaris-Button__Content"><span>Install app</span>
+                  </button>
+                  </div>
```
tiffanytse, line 42: What is this closing? Line 27?

tiffanytse (commented):
Im not sure how much of this is just auto generated and pulled in; or what you've written yourself. Just left a few comments. Looks good otherwise!

app/views/shopify_app/partials/_button_styles.html.erb
jgray7019: To answer the global comment above, everything has been written myself but copied exactly as is from the Polaris computed styles.
It all needs to be hardcoded as CSS `var()` is [not compatible](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) at all with IE

app/views/shopify_app/sessions/new.html.erb
jgray7019: Good catch! It was not needed. Removed!

tiffanytse (approved):

PR #1103, opened 2020-11-06
Update jwt#shopify_user_id to return an integer

rezaansyed:
It seems that we've introduced a slight bug for developers who are developing with the shopify_app gem.

Now that core returns the sub field of a jwt as a string, the following validation fails for any app that is developing with online access tokens using shopify_app: https://github.com/Shopify/shopify_app/blob/2401342e466c35f1617ebc541a0093a3ade351aa/app/controllers/shopify_app/callback_controller.rb#L87

In this scenario, the associated_user_id parsed by omniauth remains an Integer, but the jwt_shopify_user_id parsed by our middleware is now a String. There is a type mismatch. The easiest solution for this is to cast the sub field as an integer in our middleware to ensure that everything still works with omniauth as intended (I think this is fine because it's the developer's prerogative to decide what to do with the sub field).

This issue was flagged by Sneha here: https://shopify.slack.com/archives/CG71G24BZ/p1604674034310200

### DIFF
--- a/lib/shopify_app/session/jwt.rb
+++ b/lib/shopify_app/session/jwt.rb
@@ -25,7 +25,7 @@ module ShopifyApp
     end
 
     def shopify_user_id
-      @payload && @payload['sub']
+      @payload['sub'].to_i if @payload && @payload['sub']
     end
 
     private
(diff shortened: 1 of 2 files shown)


theundeadmonk:
I'm wondering if it would be better to convert it to a string in the place where we perform the comparison check rather than to an integer.

theundeadmonk (approved):

paulomarg (approved):

NabeelAhsen (approved):

PR #1111, opened 2020-11-17
Use Session Token auth strategy by default

rezaansyed:
This PR updates the authentication strategy of `shopify_app` to use Session Tokens by default. For the home_controller generator, the default behaviour will be to create an unauthenticated home_controller and an authenticated products_controller. 

This also supports existing apps that use cookie based authentication without any breaking changes. However, if any new apps want to use the cookie based auth strategy, they will need to pass in the `with-cookie-authentication` in the home_controller generator (not recommended). 

### Tophat instructions
1. Pull this branch
2. Create a new rails app: `rails new <app-name>`
3. Open the Gemfile of `<app_name>`  and add: 
```rb
gem 'shopify_app', path:'your/path/to/local/shopify_app'
gem 'dotenv-rails'
```
4. Run a `bundle install`

**Default behaviour**
Run `rails generate shopify_app` or `rails generate shopify_app:home_controller` followed by a `rails db:migrate`
    * Ensure that an unauthenticated home_controller file is generated
    * Ensure that the default view is created. See gist: https://gist.github.com/rezaansyed/3df08891d08ae3607e3245327f9078e6
    * Ensure that the routes.rb file is updated accordingly:

```rb
  root :to => 'home#index'
  get '/products', :to => 'products#index'
end
```

**Passing in `with-cookie-authentication`**
Run `rails generate shopify_app --with-cookie-authentication` or `rails generate shopify_app:home_controller --with-cookie-authentication` followed by a `rails db:migrate`
      * Ensure that the default authenticated home_controller file is generated
      * Ensure that the default view is created. See gist: https://gist.github.com/NabeelAhsen/a0269613246eb62df5c105c85a8aa4d9
      * Ensure that the routes.rb file is updated accordingly:

```rb
  root :to => 'home#index'
end
```


NabeelAhsen (approved):
Sweet, let's create some 🎩 instructions on this generator and specify which files we expect to generate as a result!

README.md
```diff
@@ -107,6 +107,7 @@ Options include:
 (e.g. `--scope read_products, write_orders, write_products` or `--scope "read_products, write_orders, write_products"`)
 For more information, refer to the [docs](http://docs.shopify.com/api/tutorials/oauth).
 * `embedded` - the default is to generate an [embedded app](http://docs.shopify.com/embedded-app-sdk), if you want a legacy non-embedded app then set this to false, `--embedded false`
+* __[Not recommended]__ `with-cookie-authentication` - sets up the authentication strategy of the app to use cookies. By default, it uses JWT based session tokens.
```
paulomarg, line 110: Even though we don't want cookies for embedded apps, we still need them for non-embedded ones right? Is `not recommended` the right wording here?

paulomarg (approved):

README.md
rezaansyed: We would still need them for non-embedded apps. This is a good point. By default, non-embedded apps would be setup to use JWTs, which will not work for them. Would removing `not recommended` be sufficient in supporting them?

paulomarg: Is there a way to prompt the user for their option when running the generator? That would enable us to set things up to match what they're going for.

rezaansyed: @paulomarg we could do that. Although it might make things a bit more confusing now that the generator branches off for embedded and non-embedded apps. Do you suggest we add an option specifying what app is being generated?

snehashah-shopify: Looks like there's already [an `embedded` option](https://github.com/Shopify/shopify_app/pull/1111/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R109). Maybe that can be leveraged to ensure the right path is taken based on the app choice. _(Although the documentation link points to EASDK and I'm assuming that's an outdated link)._

And the doc wording here can accordingly be updated to `[Not recommended for embedded apps]`

paulomarg: I think having the user tell us explicitly whether they're creating an embedded or non-embedded app may be helpful. I agree with Sneha that the `embedded` flag seems like a good fit for this (depending on what else it might do).

rezaansyed: Since `embedded` is the default configuration, I can make sure that the `non-embedded` option uses cookies.

lib/generators/shopify_app/home_controller/templates/index.html.erb
```diff
@@ -47,25 +47,25 @@
...
-      </ul>
+    <ul>
+      <%% @webhooks.each do |webhook| %>
+      <li><%%= webhook.topic %> : <%%= webhook.address %></li>
+      <%% end %>
+    </ul>
     <%% else %>
-      <p>This app has not created any webhooks for this Shop. Add webhooks to your ShopifyApp initializer if you need webhooks</p>
-    <%% end %><% end %>
+    <p>This app has not created any webhooks for this Shop. Add webhooks to your ShopifyApp initializer if you need webhooks</p>
```
greatestape, line 68: This reminds me. Should we start adding the uninstall webhook by default now, since it's effectively required for proper re-install behaviour when you use session tokens?

test/generators/home_controller_generator_test.rb
```diff
@@ -17,18 +17,17 @@ class HomeControllerGeneratorTest < Rails::Generators::TestCase
...
   end
 
-  test "creates the default home controller with home index view" do
+  test "creates the default unauthenticated home controller with home index view" do
     run_generator
 
-    assert_file "app/controllers/home_controller.rb", /HomeController < AuthenticatedController/
+    assert_file "app/controllers/home_controller.rb", /HomeController < ApplicationController/
     assert_file "app/views/home/index.html.erb"
-    assert_no_file "app/controllers/products_controller.rb"
```
greatestape, line 25: Should this assertion move to the authenticated controller test below?

greatestape (approved):
LGTM. I think we just need to decide on the README stuff.

README.md
greatestape: So I take it the new strategy is to drop the "with-cookie-authentication" flag and leverage the "embedded" flag? That makes sense to me. 👍

lib/generators/shopify_app/home_controller/home_controller_generator.rb
```diff
@@ -29,16 +27,26 @@ def add_home_index_route
 
       private
 
+      def embedded?
+        options['embedded'] == 'true'
+      end
+
       def embedded_app?
-        ShopifyApp.configuration.embedded_app? || options['embedded'] == 'true'
```
paulomarg: Thinking about this from the CLI's perspective, we may want to still consider the option here so we can control it from the command line (even though I don't think the CLI currently supports it).

README.md
```diff
@@ -121,10 +122,10 @@ After running the `install` generator, you can start your app with `bundle exec
 $ rails generate shopify_app:home_controller
 ```
 
-This generator creates an example home controller and view which fetches and displays products using the Shopify API.
+This generator creates an example home controller and view which fetches and displays products using the Shopify API. By default, this generator creates an unauthenticated home_controller and a sample protected products_controller.
 
 Options include:
-* __[beta]__ `with-session-token`: This flag generates an unauthenticated home_controller and a protected sample products_controller. It also creates a home view that leverages a session token to fetch products from your products_controller. Use this flag if you plan to build a single-page application or to secure your app using JWT session tokens (e.g. `--with-session-token` or `--with-session-token true`). 
+* __[Not recommended]__ `with-cookie-authentication` - This flag generates an authenticated home_controller, where the authentication strategy relies cookies. By default, this generator creates an unauthenticated home_controller and protected sample products_controller.
```
mSitkovets, line 128: missing word: relies `on` cookies

lib/generators/shopify_app/home_controller/home_controller_generator.rb
```diff
@@ -28,16 +27,26 @@ def add_home_index_route
...
 
-      def with_session_token?
-        @with_session_token
+      def with_cookie_authentication?
+        options['with_cookie_authentication']
       end
 
       def home_controller_template
-        with_session_token? ? 'unauthenticated_home_controller.rb' : 'home_controller.rb'
+        return 'unauthenticated_home_controller.rb' unless authenticated_home_controller_required?
```
mSitkovets, line 43: what's the difference between the `authenticated_home_controller_required?` flag and the `with_cookie_authentication?`. In what case would you have `with_cookie_authentication?` be false and also `authenticated_home_controller_required?` be false?

test/generators/install_generator_test.rb
```diff
@@ -73,6 +73,14 @@ class InstallGeneratorTest < Rails::Generators::TestCase
...
 
+  test "uses cookie based auth strategy for non embedded app" do
+    run_generator %w(--embedded false)
+    assert_file "config/initializers/shopify_app.rb" do |shopify_app|
+      assert_match 'config.allow_jwt_authentication = false', shopify_app
+      assert_match 'config.allow_cookie_authentication = true', shopify_app
+    end
+  end
+
   test "creats and injects into omniauth initializer" do
```
mSitkovets, line 84: spelling of `creates`

mSitkovets (approved):

lib/generators/shopify_app/home_controller/home_controller_generator.rb
rezaansyed: @paulomarg so this is how you would pass in the options now:
**default embedded app**: `rails generate shopify_app`
**default non-embedded app (uses cookies)**: `rails generate shopify_app --embedded false`
**embedded app with cookies**: `rails generate shopify_app --with-cookie-authentication`

Is this what you meant?

lib/generators/shopify_app/home_controller/templates/index.html.erb
rezaansyed: I agree although I think that should be part of a separate PR.

lib/generators/shopify_app/home_controller/home_controller_generator.rb
paulomarg: Yes, as long as we can control this from the command line I'm happy! I thought we were dropping the `embedded` option altogether, and once we've ironed out all of these choices we should change the CLI to ask whether the app is embedded or not.

rezaansyed: Not sure about removing the `embedded` option as I'm not sure how else we would specify that an app is non-embedded? If it is to be removed, I think that should be part of a separate PR.

paulomarg: Sorry, I wasn't super clear there. I'd like for the flag to stay, I thought before that you were actually removing it, so my mistake!

I think the options you suggested above sound perfect - and non-embedded + cookie is definitely not recommended 👍

PR #1113, opened 2020-11-19
Move CI pipeline from Travis to Github Actions

mllemango:
Using Github Actions for CI now. 

Using `macos` instead of `ubuntu` solves the issue of intermittent failures for yarn test. 


.github/workflows/build.yml
```diff
@@ -0,0 +1,43 @@
+name: CI
+
+on: 
+  push:
+
+jobs:
+  build:
+    runs-on: macos-latest
```
andyw8, line 8: Let's add a comment here to explain why we picked `macos`?

.github/workflows/build.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache node modules
+        uses: actions/cache@v2
+        with:
+          # npm cache files are stored in `~/.npm` on Linux/macOS
+          path: ~/.npm
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+      - name: Cache gems
```
andyw8, line 22: Could we use the `bundler-cache` option to simplify this a little?

https://github.com/ruby/setup-ruby#caching-bundle-install-automatically

mllemango:
> > Using macos instead of ubuntu solves the issue of intermittent failures for yarn test.
> 
> Oh, interesting. Was that just from trial-and-error or was there a recommendation somewhere?

Complete trial and error 😄

andyw8 (approved):

PR #1115, opened 2020-11-30
Test to see if assets are in the configuration

### LINKED ISSUE #1105: Rails 6 and Trying to Not Use Sprockets
I use Webpack for assets. However, when sending the production code to Heroku it complains:

    You set your `config.assets.compile = true` in production.

Which, I did not do, as that is not even a config method at this time. I think this is due to Sprockets. But if I remove Sprockets by commenting it out of the application.rb, the current install of Shopify fails as it believes it needs assets from Sprockets? 

What is the easy way to eliminate sprockets, just use Webpack, and have my cake and eat it too?

jeffse:
For installations using webpacker, it's possible to avoid loading sprockets altogether by forgoing the `require 'rails/all'` command and including only the Rails frameworks one needs, e.g.,
```
require 'rails'
# Pick the frameworks you want:
require 'active_model/railtie'
require 'active_job/railtie'
require 'active_record/railtie'
# require 'active_storage/engine'
require 'action_controller/railtie'
# require 'action_mailer/railtie'
# require 'action_mailbox/engine'
# require 'action_text/engine'
require 'action_view/railtie'
# require 'action_cable/engine'
# require 'sprockets/railtie'
# require 'rails/test_unit/railtie'
```
Currently, shopify_app fails on initialize if sprockets is not included.
```
home/jeff/.rvm/gems/ruby-2.6.5/gems/railties-6.0.3.4/lib/rails/railtie/configuration.rb:96:in `method_missing': undefined method `assets' for #<Rails::Application::Configuration:0x000055ababee1fe0> (NoMethodError)     
```

This fix essentially tests for the presence of sprockets before proceeding to add the assets to the precompile list.
Fixes #1105 .

### DIFF
--- a/lib/shopify_app/engine.rb
+++ b/lib/shopify_app/engine.rb
@@ -15,13 +15,15 @@ module ShopifyApp
     isolate_namespace ShopifyApp
 
     initializer "shopify_app.assets.precompile" do |app|
-      app.config.assets.precompile += %w[
-        shopify_app/redirect.js
-        shopify_app/top_level.js
-        shopify_app/enable_cookies.js
-        shopify_app/request_storage_access.js
-        storage_access.svg
-      ]
+      if app.config.respond_to?(:assets)
+        app.config.assets.precompile += %w[
+          shopify_app/redirect.js
+          shopify_app/top_level.js
+          shopify_app/enable_cookies.js
+          shopify_app/request_storage_access.js
+          storage_access.svg
+        ]
+      end
     end
 
     initializer "shopify_app.middleware" do |app|


tanema (approved):
Thanks!

jeffse:
Closing this PR for now... I discovered that it causes failure since the javascript assets for the initial install for the app still rely on sprockets. I'll see if this can't be rectified somehow, but as such the work here is incomplete.

itsniper:
@jeffse after thinking about #1105 over the weekend I was going to try this approach. But you beat me to it, thanks 👍! It would be nice to solve this so we can fully remove Sprockets 🤔.

PR #1119, opened 2020-12-07
Fix rubocop error

mllemango:
New Github Actions for the CI introduces an updated version of Rubocop. Originally the Rubocop check used (I think) 0.93.1, but now we're using 1.5.2.


NabeelAhsen (approved):
:shipit:

thecodepixi (approved):

andyw8:
@mllemango the RuboCop version uses will be that from latest rubocop-shopify (since we don't commit a `Gemfile.lock`, it will fetch the latest).

PR #1121, opened 2020-12-10
Packaging for release 16.0.0

jgray7019:
### **16.0.0**

Update all `html.erb` and `css` files to correspond with updated store admin design language [#1102](https://github.com/Shopify/shopify_app/pull/1102)

### DIFF
--- a/lib/shopify_app/version.rb
+++ b/lib/shopify_app/version.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module ShopifyApp
-  VERSION = '15.1.0'
+  VERSION = '16.0.0'
 end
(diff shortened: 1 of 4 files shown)


lib/shopify_app/version.rb
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module ShopifyApp
-  VERSION = '15.0.1'
+  VERSION = '15.1.0'
```
paulomarg, line 3: Should we consider making this a major bump? Even though it doesn't break anything it will affect the styling of existing apps, which may become inconsitent.

jgray7019: Yeah I can definitely update it, I was on the fence between the two.
I referred to previous styling updates (eg. bump to version 8.2.0) where it was just a minor bump but this is a bit of a larger change so major makes sense to me.

mllemango (approved):

PR #1131, opened 2021-01-07
Update environment variable guidance in README

andyw8:
I've moved the API key notes to before the instruction to run the generator. Otherwise, the user will run into an error due to the missing env vars.

I've also updated the guidance to suggest dotenv-rails rather than dotenv, since it makes configuration simpler.


tanema (approved):
This is great Andy, thanks

paulomarg (approved):

PR #1134, opened 2021-01-13
Rails 6.1 not yet supported

andyw8:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+Unreleased
+----------
+* Rails 6.1 is not yet supported [#1134](https://github.com/Shopify/shopify_app/pull/1134)
+
 16.1.0
 ------
 * Use Session Token auth strategy by default for new embedded apps [#1111](https://github.com/Shopify/shopify_app/pull/1111)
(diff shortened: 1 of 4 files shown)


mllemango (approved):

alexaitken (approved):

axlekb:
Just to clarify, this would only be for an embedded app, right?

andyw8:
@axlekb Correct. Using the [shopify_api](https://github.com/Shopify/shopify_api) gem directly is fine.

axlekb:
Hi @andyw8, yes, I am aware that API would be unaffected as it doesn't require any cookies, however I was more wondering about cookies for oauth portion of this gem.  We have a non-embedded app, and we are running 6.1...  We haven't noticed any problems, but just want to make sure that I'm not missing anything.

kirillplatonov:
Hey @andyw8 👋

Does it really mandatory to restrict the Rails version to be < 6.1?
I'm using shopify_app 16.1 with an embedded app with no issues. The only change that is needed is enabling old SameSite cookies protection in Rails config like this:
```ruby
# config/application.rb

config.load_defaults 6.1
config.action_dispatch.cookies_same_site_protection = nil
```

PR #1144, opened 2021-01-17
Don't run app config for generator

andyw8:
In https://github.com/Shopify/shopify_app/pull/1097 a change was required that the Shopify environment variables were present. As this code is in an initializer, it also runs when the generators run, but at that point in time the environment variables may not yet be available (e.g. because dotenv-rails hasn't been added yet).

Based on a suggestion in https://stackoverflow.com/a/32789186, we can check if the initializer is running from a generator or not.

### DIFF
--- a/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
+++ b/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
@@ -1,16 +1,18 @@
-ShopifyApp.configure do |config|
-  config.application_name = "<%= @application_name %>"
-  config.api_key = ENV.fetch('SHOPIFY_API_KEY', '').presence || raise('Missing SHOPIFY_API_KEY')
-  config.secret = ENV.fetch('SHOPIFY_API_SECRET', '').presence || raise('Missing SHOPIFY_API_SECRET')
-  config.old_secret = "<%= @old_secret %>"
-  config.scope = "<%= @scope %>" # Consult this page for more scope options:
-                                 # https://help.shopify.com/en/api/getting-started/authentication/oauth/scopes
-  config.embedded_app = <%= embedded_app? %>
-  config.after_authenticate_job = false
-  config.api_version = "<%= @api_version %>"
-  config.shop_session_repository = 'Shop'
-  config.allow_jwt_authentication = <%= !with_cookie_authentication? %>
-  config.allow_cookie_authentication = <%= with_cookie_authentication? %>
+unless defined? Rails::Generators
+  ShopifyApp.configure do |config|
+    config.application_name = "<%= @application_name %>"
+    config.api_key = ENV.fetch('SHOPIFY_API_KEY', '').presence || raise('Missing SHOPIFY_API_KEY')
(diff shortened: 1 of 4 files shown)


test/generators/install_generator_test.rb
```diff
@@ -23,10 +23,8 @@ class InstallGeneratorTest < Rails::Generators::TestCase
     run_generator
     assert_file "config/initializers/shopify_app.rb" do |shopify_app|
       assert_match 'config.application_name = "My Shopify App"', shopify_app
-      assert_match "config.api_key = ENV.fetch('SHOPIFY_API_KEY', '').presence " \
-        "|| raise('Missing SHOPIFY_API_KEY')", shopify_app
-      assert_match "config.secret = ENV.fetch('SHOPIFY_API_SECRET', '').presence " \
-        "|| raise('Missing SHOPIFY_API_SECRET')", shopify_app
+      assert_match "config.api_key = ENV.fetch('SHOPIFY_API_KEY', '')", shopify_app
```
andyw8, line 26: Rather than copy the exact updated template code into the test, I've simplified this to check for just the 'essential' part. Verifying the template code doesn't prove that the generator works – ideally we should generate an app as part of the test suite.

paulomarg (approved):

lib/generators/shopify_app/install/templates/shopify_app.rb.tt
```diff
@@ -1,16 +1,18 @@
...
-  config.old_secret = "<%= @old_secret %>"
-  config.scope = "<%= @scope %>" # Consult this page for more scope options:
-                                 # https://help.shopify.com/en/api/getting-started/authentication/oauth/scopes
-  config.embedded_app = <%= embedded_app? %>
-  config.after_authenticate_job = false
-  config.api_version = "<%= @api_version %>"
-  config.shop_session_repository = 'Shop'
-  config.allow_jwt_authentication = <%= !with_cookie_authentication? %>
-  config.allow_cookie_authentication = <%= with_cookie_authentication? %>
+unless defined? Rails::Generators
```
xiaoboa, line 1: @andyw8  with this change, the reference of `ShopifyApp.configuration` in generator code(not template code) should also be removed, there is one at [home_controller_generator.rb](https://github.com/Shopify/shopify_app/blob/master/lib/generators/shopify_app/home_controller/home_controller_generator.rb#L35)
```ruby
      def embedded_app?
        ShopifyApp.configuration.embedded_app?
      end
```
the method `embedded_app?` is no longer needed any more. 
Otherwise the generated code of `HomeController` will be an authenticated home_controller instead of the unauthenticated home_controller, which is conflict with the [README.md](https://github.com/Shopify/shopify_app/blob/master/README.md#home-controller-generator)

>  By default, this generator creates an unauthenticated home_controller and a sample protected products_controller.

paulomarg: That's a good point @xiaoboa, I'm actually investigating that myself. We still need need the method if we're creating non-embedded apps but I'll make sure that home_controller is not authenticated.

Thanks!

PR #1149, opened 2021-01-19
No Ruby 3 yet

andyw8:


shopify_app.gemspec
```diff
@@ -9,7 +9,7 @@ Gem::Specification.new do |s|
   s.author      = "Shopify"
   s.summary     = 'This gem is used to get quickly started with the Shopify API'
 
-  s.required_ruby_version = ">= 2.5"
+  s.required_ruby_version = ">= 2.5", "< 3"
```
andyw8, line 12: This may not be right, I need to verify.

https://guides.rubygems.org/specification-reference/#required_ruby_version

Issue #1128, opened 2020-12-21
Configuring the script tags is causing errors


yehiasalam:
Hi,

I ran my custom app using the shopify-cli and shopify_app, and everything was working fine. I'm trying right now to add an external js file to the frontstore, using the config.scripttags as instructed in the docs:

```
ShopifyApp.configure do |config|
  config.application_name = "My Shopify App"
  config.api_key = ENV.fetch('SHOPIFY_API_KEY', '').presence || raise('Missing SHOPIFY_API_KEY')
  config.secret = ENV.fetch('SHOPIFY_API_SECRET', '').presence || raise('Missing SHOPIFY_API_SECRET')
  config.old_secret = ""
  config.scope = "write_script_tags" # Consult this page for more scope options:
                                 # https://help.shopify.com/en/api/getting-started/authentication/oauth/scopes
  config.embedded_app = true
  config.after_authenticate_job = false
  config.api_version = "2020-10"
  config.shop_session_repository = 'Shop'
  config.allow_jwt_authentication = true
  config.allow_cookie_authentication = false

  config.scripttags = [
    {event:'onload', src: 'https://x.x.com/widget/widget.js'}
  ]

end
```

However, whenever I install the app i keep getting the following error, and this happens after adding:

```
  config.scripttags = [
    {event:'onload', src: 'https://x.x.com/widget/widget.js'}
  ]
```

![Screenshot 2020-12-21 235958](https://user-images.githubusercontent.com/201734/102826352-26a94300-43e9-11eb-86c6-1bf0fc6e9dad.png)

Not sure what might the problem might be. It shouldn't be related to Auth i guess since it was working fine before adding the scripttags config.

paulomarg:
Hey @yehiasalam, thank you for reporting this. We'll look into it!

tanema:
Does this happen in all browsers? Was authentication seemingly successful before this error?

m0n9oose:
Just got pretty much the same exception:

<img width="827" alt="Screenshot 2021-01-12 at 13 22 30" src="https://user-images.githubusercontent.com/1478556/104314107-3ba15f80-54d9-11eb-977e-38d5aa7a0064.png">

I'm using JWT authentication (in process of migration from OAuth), Rails 6.1, shopify_app 16.1.0.

We use `webhooks` setting, but not `scripttags`. Problem happened when I uninstalled my dev app and tried to install it again.

Here's my initialiser:

```ruby
ShopifyApp.configure do |config|
[...]

PR #1151, opened 2021-01-19
Ensure offline and online tokens are fetched via available auth hash

rezaansyed:
Fixes https://github.com/Shopify/shopify_app/issues/1128

We have noticed that the callback controller was fetching a token based on the session object for post perform jobs. This is not necessary as the offline and/or online tokens can be retrieved via the auth hash.

### DIFF
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -89,6 +89,14 @@ module ShopifyApp
       auth_hash.uid
     end
 
+    def offline_access_token
+      ShopifyApp::SessionRepository.retrieve_shop_session_by_shopify_domain(shop_name)&.token
+    end
+
+    def online_access_token
+      ShopifyApp::SessionRepository.retrieve_user_session_by_shopify_user_id(associated_user_id)&.token
+    end
+
     def associated_user
       return unless auth_hash.dig('extra', 'associated_user').present?
 
@@ -132,7 +140,7 @@ module ShopifyApp
 
       WebhooksManager.queue(
         shop_name,
-        shop_session&.token || user_session.token,
+        offline_access_token || online_access_token,
         ShopifyApp.configuration.webhooks
       )
     end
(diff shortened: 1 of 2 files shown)


resistorsoftware:
Thanks for this. Took care of the problem nicely.

tanema (approved):
looks great!

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -89,6 +89,14 @@ def shop_name
       auth_hash.uid
     end
 
+    def offline_access_token
+      ShopifyApp::SessionRepository.retrieve_shop_session_by_shopify_domain(shop_name)&.token
+    end
```
greatestape, line 94: Maybe we should stick with "shop" and "user" rather than "offline" and "online" since that's what we're using elsewhere in the code.

rezaansyed: That's a good point. I thought this approach would help us slowly move towards adopting the offline and online terminology.

resistorsoftware: If I can interject here. You guys hit the nail on the head. To me, some outside plebe with no idea what is going on, this whole concept of Shop versus User has never really been clearly explained anywhere, by anyone. The explanations as they are, remain weak and abstract. How about some case studies. You never explain the difference between a User and a User from the sense of Shopify Plus either. So that is confusing. And now to go, online, offline in lingo, say what? 

We in the community suffer through so many bike shed paintings already. Avoiding those at all costs is doing a huge favor! It used to me myshopify_domain, then it was shopify_domain; it is SHOPIFY_API_SECRET but someone in JS world wants it to be SHOPIFY_API_SECRET_KEY, and I am like OH FOR THE LOVE OF PAINTING YAKS AND SHAVING THEIR SHEDS.

Just my two cents. Have at 'er.

greatestape: Now that we've chatted about it, I get why it'd be nice to call these "offline_access_token" and "online_access_token". Most of our use of "shop" and "user" is separate from this.

Having said that, we do have some follow-up work, for example, with the use of "shop_token" in the webhooks manager code.

greatestape (approved):
LGTM

PR #1152, opened 2021-01-20
Packaging for release 17.0.2

rezaansyed:
17.0.2 (January 20, 2021)
------
* Fix failing script tags and webhooks installs after performing OAuth [#1151](https://github.com/Shopify/shopify_app/pull/1151)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Unreleased
 ----------
 
+17.0.2 (January 20, 2021)
+------
+* Fix failing script tags and webhooks installs after performing OAuth [#1151](https://github.com/Shopify/shopify_app/pull/1151)
+
 17.0.1 (January 18, 2021)
 ------
 * Don't attempt to read Shopify environment variables when the generators are running, since they may not be present yet [#1144](https://github.com/Shopify/shopify_app/pull/1144)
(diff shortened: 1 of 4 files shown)


paulomarg (approved):
LGTM, just checking: are the other gems in gemspec being upgraded supposed to be here?

rezaansyed:
> LGTM, just checking: are the other gems in gemspec being upgraded supposed to be here?

@paulomarg this was from running bundle as specified in the [Releasing doc](https://github.com/Shopify/shopify_app/blob/master/docs/Releasing.md). I did not anticipate any changes other than the gem's version.

andyw8:
Since these are just patch level updates I think it's fine to include. (If they were major/minor then I'd maybe mention them in the changelog).

paulomarg:
Makes sense to me, we just need to make sure to 🎩  changes before releasing if we're updating the deps as part of the release process.

rezaansyed:
I 🎩'd to make sure that the install webhook and scripttags are working (new changes introduced). Is that sufficient or do I need to 🎩 existing behaviour?

paulomarg:
Yeah, like Andy said these are minor bumps so it should be fine - I was considering updating the releasing instructions to reflect this.

andyw8:
I just realised the whole debate above was kinda pointless – the `Gemfile.lock` is ignored when the gem by a project.

PR #1155, opened 2021-01-21
Only skip environment variables from initializer for generators

paulomarg:
Once we started skipping the initializer for generators (#1144), we ended up in a situation where the default config generated by the template and the default values in `ShopifyApp` configuration had opposing values, therefore causing new apps to generate an authenticated `home_controller.rb` file rather than its unauthenticated counterpart because cookie auth is still the default.

This PR addresses that by only skipping the values that depend on environment variables when running generators so that the actual config of the app still applies on generator runs and creates consistent apps.

- [X] For security fixes, the [Disclosure Policy](https://github.com/Shopify/shopify_app/blob/master/SECURITY.md#disclosure-policy) must be followed.

### DIFF
--- a/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
+++ b/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
@@ -1,17 +1,20 @@
-unless defined? Rails::Generators
-  ShopifyApp.configure do |config|
-    config.application_name = "<%= @application_name %>"
-    config.api_key = ENV.fetch('SHOPIFY_API_KEY', '').presence || raise('Missing SHOPIFY_API_KEY. See https://github.com/Shopify/shopify_app#api-keys')
-    config.secret = ENV.fetch('SHOPIFY_API_SECRET', '').presence || raise('Missing SHOPIFY_API_SECRET. See https://github.com/Shopify/shopify_app#api-keys')
-    config.old_secret = "<%= @old_secret %>"
-    config.scope = "<%= @scope %>" # Consult this page for more scope options:
-                                   # https://help.shopify.com/en/api/getting-started/authentication/oauth/scopes
-    config.embedded_app = <%= embedded_app? %>
-    config.after_authenticate_job = false
-    config.api_version = "<%= @api_version %>"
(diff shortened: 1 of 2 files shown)


lib/generators/shopify_app/install/templates/shopify_app.rb.tt
```diff
@@ -1,17 +1,18 @@
...
+  config.scope = "<%= @scope %>" # Consult this page for more scope options:
+                                  # https://help.shopify.com/en/api/getting-started/authentication/oauth/scopes
+  config.embedded_app = <%= embedded_app? %>
+  config.after_authenticate_job = false
+  config.api_version = "<%= @api_version %>"
+  config.shop_session_repository = 'Shop'
+  config.allow_jwt_authentication = <%= !with_cookie_authentication? %>
+  config.allow_cookie_authentication = <%= with_cookie_authentication? %>
+
+  if defined?(Rails::Server) || defined?(Rails::Console) || (File.basename($PROGRAM_NAME) === 'rake')
```
andyw8, line 13: Since Rails 5, some `rake` commands are also callable with `rails`, e.g. `rails routes` instead of `rake routes` – can we check this is handled correctly?

andyw8: Also, the comparison can be `==`, the `===` is very rarely needed in Ruby.

paulomarg: I'm starting to wonder if the `raise` call is worth going through all of this trouble. We could simply let the keys be empty if the value isn't set rather than checking for all possible scenarios.

What if we just log something if `Server` is defined and let things slide with an empty key for other cases?

andyw8: Starting the server with missing env vars would be the most common mistake. so how about we raise on that, but ignore the others?

paulomarg: Excellent point, I agree. Making it happen 🙂

andyw8 (approved):

rezaansyed (approved):

PR #1156, opened 2021-01-21
Create a turbolinks asset file to generate during install

NabeelAhsen:
### Context

> Builds on https://github.com/Shopify/turbolinks-jwt-sample-app/pull/21

This PR is an attempt to consolidate and version our suggested Turbolinks solution with session tokens. Currently, we don't have a library for this type of work, and it has made iterating our suggestions difficult.

> Example: We recently changed our Turbolinks refresh interval suggestion from 50 seconds to 2 seconds. This was difficult to communicate with partners because we don't have a versioned solution to this outside of our public documents.

### What this change is about

* This PR introduces a new javascript asset file that gets generated on gem install
* Using this file is entirely optional
  * Partners who need Turbolinks + session token capability with their SSR apps can require it in their `index.js` files
  * This file will not automatically be bootstrapped by Webpack, but it will be available to partners
* This move puts us in a position to better communicate our Turbolinks solution with partners in a timely manner
* Now that [Turbolinks is no longer being maintained](https://github.com/turbolinks/turbolinks#turbolinks-is-no-longer-under-active-development), this change also helps us come up with a plan to sunset support for this feature in the near future

### What this change is not about

* This is not a complete solution. Coupling our Turbolinks solution with the Shopify App gem alone is not sufficient for all partners who need this capability and don't use this gem

### Next steps

* It might take a while for us to figure out a "good enough" Turbolinks solution. With this change, we hope to able to iterate and version our suggestions in one place now
* Identify a better vehicle for the behaviour this file provides apps
  * Could this be part of an NPM package? Is there a Shopify library we can introduce this behaviour in?

---

### DIFF
--- a/lib/generators/shopify_app/install/templates/shopify_app_turbolinks.js
+++ b/lib/generators/shopify_app/install/templates/shopify_app_turbolinks.js
@@ -1,0 +1,1 @@
+// TODO: Determine what the final shape of this file is
--- a/lib/generators/shopify_app/install/templates/shopify_app_turbolinks.js
+++ b/lib/generators/shopify_app/install/templates/shopify_app_turbolinks.js
@@ -1,1 +1,65 @@
-// TODO: Determine what the final shape of this file is
+/*
+ * If your app is a server-side rendered app that uses Turbolinks with session tokens,
+ * require this file at the end of `app/javascript/shopify_app/index.js`:
+ *
+ * >> require('./shopify_app_turbolinks');
+ */
+
+import { getSessionToken } from "@shopify/app-bridge-utils";
+
+const SESSION_TOKEN_REFRESH_INTERVAL = 2000; // Request a new token every 2s
+
+document.addEventListener("turbolinks:request-start", function (event) {
+  var xhr = event.data.xhr;
+  xhr.setRequestHeader("Authorization", "Bearer " + window.sessionToken);
+});
+
+document.addEventListener("turbolinks:render", function () {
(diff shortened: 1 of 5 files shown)


test/generators/install_generator_test.rb
```diff
@@ -119,4 +119,23 @@ class InstallGeneratorTest < Rails::Generators::TestCase
...
     end
   end
+
+  test "generates javascript assets required when embedded" do
+    ShopifyApp.stubs(:use_webpacker?).returns(false)
+
+    run_generator %w(--embedded)
+    assert_file "app/assets/javascripts/shopify_app.js"
+    assert_file "app/assets/javascripts/flash_messages.js"
+  end
```
NabeelAhsen, line 129: I am currently missing key understanding into how apps behave when they use Turbolinks but don't rely on Webpack. How do we let apps opt-in to the `turbolink.js` asset if they don't use Webpack?

lib/generators/shopify_app/install/templates/shopify_app_turbolinks.js
```diff
@@ -0,0 +1,65 @@
+/*
+ * If your app is a server-side rendered app that uses Turbolinks with session tokens,
+ * require this file at the end of `app/javascript/shopify_app/index.js`:
+ *
+ * >> require('./shopify_app_turbolinnks');
```
andyw8, line 5:
```suggestion
 * >> require('./shopify_app_turbolinks');
```

NabeelAhsen:
Closing this PR for now. We most likely need a new library for javascript functionality in the near future.

PR #1118, opened 2020-12-07
Create concern to authenticate deep links using Turbolinks

NabeelAhsen:
> Related https://github.com/Shopify/shopify_app/issues/1092

This is related to supporting session tokens using Turbolinks. This PR introduces an optional controller concern that helps authenticate multi-page apps when accessing a link directly. 

Currently, with third party cookies disabled, a user is unable to access protected endpoints like `/products` or `/widgets`. When using Turbolinks, the recommended way of issuing a valid session token via App Bridge is by creating a splash page as the `root` resource of the app.

More info: https://shopify.dev/tutorials/authenticate-server-side-rendered-apps-with-session-tokens-app-bridge-turbolinks

This concern redirects an app to the `root` controller along with a `return_to` and `shop` parameter if it detects that an action is unauthenticated (i.e it fails to detect a valid session token session). Once a session token has been issued by App Bridge, Turbolinks can use this to redirect the app back to the `return_to` parameter in `shopify_app.js`:

```js
  // Wait for a session token before trying to load an authenticated page
  await retrieveToken(app);

  // Redirect to the requested page
  Turbolinks.visit(data.loadPath);
```

To use this concern, add it to the sample `AuthenticatedController` that gets generated by the Shopify App gem:

```ruby
# frozen_string_literal: true

class AuthenticatedController < ApplicationController
  include ShopifyApp::TurbolinksAuthenticated # <= Add this here
  include ShopifyApp::Authenticated
end
```

- [x] Show this works using app extensions and direct app-links
- [ ] Update `CHANGELOG.md` if the changes would impact users
- [x] Update `README.md`, if appropriate.
- [x] Update any relevant pages in `docs/`, if necessary
- [x] For security fixes, the [Disclosure Policy](https://github.com/Shopify/shopify_app/blob/master/SECURITY.md#disclosure-policy) must be followed.

PR #1158, opened 2021-01-22
[Bug fix] Redirect EnsureAuthenticatedLinks to login page if shop domain not found

NabeelAhsen:
> Fixes the issue brought up by a partner https://github.com/Shopify/shopify_app/pull/1118#issuecomment-765350708

In a world without cookies, an embedded app needs to handle attempted access to protected resources both in an embedded and non-embedded state. This PR fixes the scenario where `ShopifyApp::LoginProtection::ShopifyDomainNotFound` is raised when:
* The app is in a non-embedded state
* No shop parameter is passed in
* The app is unauthenticated
* The browser does not support 3p cookies (and subsequently does not have a 1p cookie saved on its domain with a sample `shop` domain)

In this scenario, rather than raise an exception, we should redirect to the app's `login_url`.

### DIFF
--- a/app/controllers/concerns/shopify_app/ensure_authenticated_links.rb
+++ b/app/controllers/concerns/shopify_app/ensure_authenticated_links.rb
@@ -13,7 +13,9 @@ module ShopifyApp
     def redirect_to_splash_page
       splash_page_path = root_path(return_to: request.fullpath, shop: current_shopify_domain)
       redirect_to(splash_page_path)
-    rescue ShopifyApp::LoginProtection::ShopifyDomainNotFound
+    rescue ShopifyApp::LoginProtection::ShopifyDomainNotFound => error
+      Rails.logger.warn("[ShopifyApp::EnsureAuthenticatedLinks] Redirecting to login: [#{error.class}] "\
+                         "Could not determine current shop domain")
       redirect_to(ShopifyApp.configuration.login_url)
     end
 
--- a/app/controllers/concerns/shopify_app/ensure_authenticated_links.rb
+++ b/app/controllers/concerns/shopify_app/ensure_authenticated_links.rb
@@ -13,6 +13,8 @@ module ShopifyApp
     def redirect_to_splash_page
       splash_page_path = root_path(return_to: request.fullpath, shop: current_shopify_domain)
       redirect_to(splash_page_path)
+    rescue ShopifyApp::LoginProtection::ShopifyDomainNotFound
+      redirect_to(ShopifyApp.configuration.login_url)
     end
 
     def missing_expected_jwt?
--- a/test/controllers/concerns/ensure_authenticated_links_test.rb
+++ b/test/controllers/concerns/ensure_authenticated_links_test.rb
@@ -55,4 +55,12 @@ class EnsureAuthenticatedLinksTest < ActionController::TestCase
 
     assert_response :ok
   end
+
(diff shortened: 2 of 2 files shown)


rezaansyed (approved):
LGTM!

paulomarg (approved):

thecodepixi (commented):
This makes sense to me, but I'm curious about why we don't want to raise anything here? Would we want to communicate to the app user that some information was missing or that something went wrong? Or is the scenario simply a "you need to actually log in to do that" in which case landing on the login page is enough?

NabeelAhsen:
> This makes sense to me, but I'm curious about why we don't want to raise anything here? Would we want to communicate to the app user that some information was missing or that something went wrong? Or is the scenario simply a "you need to actually log in to do that" in which case landing on the login page is enough?

I think the message that is being conveyed here is:

>If the app can't determine anything about the current shop in context, then the app is most likely not logged in

With the current implementation, app developers see a nasty rails error thrown. The likely move in this scenario would be to (almost always?) go login and get the app into an embedded state.

The `LoginProtection` concern previously took care of this at the cost of depending on cookies. This is behaviour that we'd ideally like to isolate and silo in the future as we move away from cookie dependency.

paulomarg:
> _Edit:_ I'm still on the fence whether hiding this error message is the right call or not, but rescuing it here is more consistent with the experience users had in the world of cookies.

Those are all good points. If we're not sure if this is the right behaviour, would it make sense to log this, so we have some means of keeping track of it?

NabeelAhsen:
> Those are all good points. If we're not sure if this is the right behaviour, would it make sense to log this, so we have some means of keeping track of it?

The behaviour I'm proposing here is actually mirrored here under a slightly different context: https://github.com/Shopify/shopify_app/blob/5af43efad6718ff24ca97afc4ee7e5029a923167/app/controllers/concerns/shopify_app/require_known_shop.rb#L19-L21

There is work underway that aims to add deliberate observability (and logging) to this gem. I'm tempted to defer the decision to log anything here until we can inventory the circumstances for redirects, including stale sessions, cookie dependencies, and expired tokens. What are your thoughts on this?

paulomarg:
That sounds sensible to me. I would advocate for adding the logging ahead of the bigger refactoring **only if** this is a risky scenario where we could break apps. Since this is a very specific case and every condition points to there not being a session it is probably safe to defer it.

That being said, I would definitely flag this as a case to consider. IMHO no error should go away silently, but this has a user-facing result (of redirecting to login) and thus is detectable (and reportable) by end users if they see it happening.

NabeelAhsen:
> That sounds sensible to me. I would advocate for adding the logging ahead of the bigger refactoring **only if** this is a risky scenario where we could break apps. Since this is a very specific case and every condition points to there not being a session it is probably safe to defer it.
> 
> That being said, I would definitely flag this as a case to consider. IMHO no error should go away silently, but this has a user-facing result (of redirecting to login) and thus is detectable (and reportable) by end users if they see it happening.

On second thought, you've convinced me that it's probably worth adding logging now rather than deferring it. It would help partners navigate this behaviour if it's something they did not expect by switching to session tokens.

paulomarg (approved):
LGTM! Strangely I convinced you to do this by trying to make the point not to 😄 

In any case, I think this kind of logging is always valuable when things go wrong, so I'm all for it.

PR #1164, opened 2021-01-26
Fix omniauth strategy not being set correctly for apps using session tokens

rezaansyed:
## What is the issue?
Currently, new apps using `shopify_app` have session tokens enabled and cookies based auth disabled by default. The problem is that even at the top level actions where cookies are accessible and appropriate to use, `LoginProtection` gates it's usage in `user_session_by_cookie` and `shop_session_by_cookie` using the `allow_cookie_authentication` setting. This has led to calls for the `user_session` and `shop_session` resulting in `nil`, leading the app to falsely believe that it doesn't have an offline and/or online token.

### OAuth redirect loops for offline tokens
1. **[/login]** Initiate OAuth to fetch the offline/shop access token
1. Goes through OAuth process
1. **[/callback]** Store the offline access token
1. **[/callback]** Respond with [redirect to login to fetch online token](https://github.com/Shopify/shopify_app/blob/6c5a46ec49044d10940446ae94aac021fd248bea/app/controllers/shopify_app/callback_controller.rb#L28)
1. 🔥 **[/login]** Fails to set `session[:user_tokens]` to true as [`shop_session` check](https://github.com/Shopify/shopify_app/blob/master/app/controllers/shopify_app/sessions_controller.rb#L97) returns [`nil`](https://github.com/Shopify/shopify_app/blob/b478a4204f6dfce1cea549c222575d0ef4514f12/lib/shopify_app/controller_concerns/login_protection.rb#L70)
1. OmniAuth [falsely assumes app requires an offline token](https://github.com/Shopify/shopify_app/blob/master/lib/generators/shopify_app/install/templates/shopify_provider.rb#L19) and starts process to fetch one
1. Repeat steps 1-6 until Shopify stops looping process

### OAuth redirect loops for online tokens
1. **[/login]** Initiate OAuth to fetch the online/user access token
1. Goes through OAuth process
1. **[/callback]** Store the online access token
1. **[/callback]** Checks to see if it needs to [start online token flow](https://github.com/Shopify/shopify_app/blob/6c5a46ec49044d10940446ae94aac021fd248bea/app/controllers/shopify_app/callback_controller.rb#L72)
1. 🔥 **[/callback]** `user_session` results in [`nil`](https://github.com/Shopify/shopify_app/blob/6c5a46ec49044d10940446ae94aac021fd248bea/lib/shopify_app/controller_concerns/login_protection.rb#L54)
1. **[/callback]** Respond with redirect to login to fetch online token
1.  Repeat steps 1-6 until Shopify stops looping process

## What is the fix?

`SessionsController` and `CallbackController` can both operate from the top-level. Use of session cookies are justified in these scenarios.

### OAuth redirect loops for offline tokens
Override the `shop_session_by_cookie` method in `SessionsController` to ensure we are still using session cookies as before on the top level.

### OAuth redirect loops for online tokens
Override the `user_session_by_cookie` method in `CallbackController` to ensure we are still using session cookies as before on the top level.

### Before

https://user-images.githubusercontent.com/60748788/105908241-f9594180-5ff3-11eb-8b8c-c2fb8fa35481.mov

### After

https://user-images.githubusercontent.com/60748788/105908009-b13a1f00-5ff3-11eb-9c82-7b691ba48653.mov

### DIFF
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -65,6 +65,13 @@ module ShopifyApp
       end
     end
 
+    # Override user_session_by_cookie from LoginProtection to bypass allow_cookie_authentication
+    # setting check because session cookies are justified at top level
+    def user_session_by_cookie
+      return unless session[:user_id].present?
+      ShopifyApp::SessionRepository.retrieve_user_session(session[:user_id])
+    end
+
     def start_user_token_flow?
       if jwt_request?
         false
--- a/app/controllers/shopify_app/sessions_controller.rb
+++ b/app/controllers/shopify_app/sessions_controller.rb
@@ -92,9 +92,18 @@ module ShopifyApp
       end
     end
 
+    # Override shop_session_by_cookie from LoginProtection to bypass allow_cookie_authentication
+    # setting check because session cookies are justified at top level
+    def shop_session_by_cookie
+      return unless session[:shop_id].present?
+      ShopifyApp::SessionRepository.retrieve_shop_session(session[:shop_id])
+    end
+
     # rubocop:disable Lint/SuppressedException
     def set_user_tokens_option
-      if shop_session.blank?
+      current_shop_session = shop_session
+
+      if current_shop_session.blank?
         session[:user_tokens] = false
         return
       end
@@ -102,9 +111,9 @@ module ShopifyApp
       session[:user_tokens] = ShopifyApp::SessionRepository.user_storage.present?
 
       ShopifyAPI::Session.temp(
-        domain: shop_session.domain,
-        token: shop_session.token,
-        api_version: shop_session.api_version
+        domain: current_shop_session.domain,
+        token: current_shop_session.token,
+        api_version: current_shop_session.api_version
       ) do
         ShopifyAPI::Metafield.find(:token_validity_bogus_check)
       end
--- a/test/controllers/callback_controller_test.rb
+++ b/test/controllers/callback_controller_test.rb
@@ -113,7 +113,9 @@ module ShopifyApp
       assert_equal 'valid-shop-id', session[:shop_id]
     end
 
-    test '#callback clears a stale shop_id if user session is for a different shop' do
+    test '#callback clears a stale shop_id if user session is for a different shop when using cookie auth' do
+      ShopifyApp.configuration.allow_jwt_authentication = false
+      ShopifyApp.configuration.allow_cookie_authentication = true
       mock_shopify_user_omniauth
       session[:shop_id] = 'valid-shop-id'
       other_shop_session = ShopifyAPI::Session.new(
@@ -311,6 +313,9 @@ module ShopifyApp
     end
 
     test "#callback redirects to the root_url when not using JWT authentication" do
+      ShopifyApp.configuration.allow_jwt_authentication = false
+      ShopifyApp.configuration.allow_cookie_authentication = true
+
       mock_shopify_omniauth
 
       get :callback, params: { shop: 'shop' }
--- a/test/controllers/sessions_controller_test.rb
+++ b/test/controllers/sessions_controller_test.rb
@@ -12,8 +12,6 @@ module ShopifyApp
     setup do
       @routes = ShopifyApp::Engine.routes
       ShopifyApp::SessionRepository.shop_storage = ShopifyApp::InMemoryShopSessionStore
-      ShopifyApp.configuration = nil
-      ShopifyApp.configuration.embedded_app = true
 
       I18n.locale = :en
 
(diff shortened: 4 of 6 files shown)


app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -94,17 +94,20 @@ def authenticate_with_partitioning
 
     # rubocop:disable Lint/SuppressedException
     def set_user_tokens_option
-      if shop_session.blank?
+      store_session = shop_session ||
```
ragalie, line 97: Is this safe? A nefarious merchant could put someone else's shop name in the query params and then we would make some API call against that other shop?

rezaansyed: You're right. Since this check would happen on the top level, we should allow the use of the cookie in this case.

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -69,7 +69,7 @@ def start_user_token_flow?
       if jwt_request?
         false
       else
-        ShopifyApp::SessionRepository.user_storage.present? && user_session.blank?
+        ShopifyApp::SessionRepository.user_storage.present? && associated_user.blank?
```
greatestape, line 72: This changes the meaning of this code, right? Before it would decide to start the user flow if the app had user storage configured and a user session didn't exist. Now it's going to start a user token flow when user storage is present and it hasn't just completed a user token flow, because that's the only time when associated_user will be populated.

Does that matter?

NabeelAhsen, line 72: Are there tests associated with this change that also need to change?

NabeelAhsen: I'm curious about whether this can ever be the case:

* User storage is configured
* A `user_session` is not blank
* `associated_user` is blank

NabeelAhsen (approved):
I'm not a fan of the overrides here, but let's mark follow-up work to see if we can refactor the `*_session_by_cookie` methods to behave differently under different contexts

rezaansyed:
> I'm not a fan of the overrides here, but let's mark follow-up work to see if we can refactor the `*_session_by_cookie` methods to behave differently under different contexts

This is to limit the blast radius of our changes. `SessionController` and `CallbackController` are both controllers that operate on the top level and are justified in using session cookies. We need to look further into splitting up components whose lifecycles are in an embedded context from those with lifecycles at the top level.

app/controllers/shopify_app/callback_controller.rb
rezaansyed: That is the case when you're arriving at the callback controller after fetching an offline/shop token. But currently that case isn't likely to exist.

tanema (approved):

greatestape (approved):
LGTM

test/controllers/sessions_controller_test.rb
```diff
@@ -13,6 +13,8 @@ class SessionsControllerTest < ActionController::TestCase
       @routes = ShopifyApp::Engine.routes
       ShopifyApp::SessionRepository.shop_storage = ShopifyApp::InMemoryShopSessionStore
       ShopifyApp.configuration = nil
+      ShopifyApp.configuration.allow_jwt_authentication = true
+      ShopifyApp.configuration.allow_cookie_authentication = false
       ShopifyApp.configuration.embedded_app = true
```
greatestape, line 18: Do we still need this?

test/shopify_app/controller_concerns/login_protection_test.rb
```diff
@@ -172,6 +175,7 @@ class LoginProtectionControllerTest < ActionController::TestCase
   end
 
   test "#current_shopify_session is memoized and does not retreive session twice" do
+    ShopifyApp.configuration.allow_cookie_authentication = true
     with_application_test_routes do
       session[:shop_id] = "foobar"
```
greatestape, line 180: It doesn't seem like this needs to be cookie-based. It just needs to work one way or the other.

How about this?

```suggestion
    with_application_test_routes do
      request.env['jwt.shopify_domain'] = 'foobar'
```

You'd also need to update the `expects` to:
```ruby
ShopifyApp::SessionRepository.expects(:retrieve_shop_session_by_shopify_domain).returns(session).once
```

, but it lets us avoid changing config values.

test/controllers/callback_controller_test.rb
```diff
@@ -20,6 +20,8 @@ class CallbackControllerTest < ActionController::TestCase
       @routes = ShopifyApp::Engine.routes
       ShopifyApp.configuration = nil
       ShopifyApp.configuration.embedded_app = true
+      ShopifyApp.configuration.allow_jwt_authentication = true
+      ShopifyApp.configuration.allow_cookie_authentication = false
```
greatestape, line 24: Do we need this?

greatestape (approved):
LGTM

PR #1165, opened 2021-01-27
Tidy up tests with useless assertions

greatestape:
We went from `:shopify` -> `:shop_id` in version 13, but we had tests that still passed despite using the old name. This tidies up those tests. It also removes one test that always passed because the exceptional behaviour it was testing was actually the default behaviour that always happened.

### DIFF
--- a/test/controllers/sessions_controller_test.rb
+++ b/test/controllers/sessions_controller_test.rb
@@ -254,17 +254,19 @@ module ShopifyApp
       assert_redirected_to "#{ShopifyApp.configuration.root_url}?shop=shop.myshopify.com"
     end
 
-    test "#destroy should clear shopify from session and redirect to login with notice" do
+    test "#destroy should reset rails session and redirect to login with notice" do
       shop_id = 1
       session[:shopify] = shop_id
       session[:shopify_domain] = 'shop1.myshopify.com'
       session[:shopify_user] = { 'id' => 1, 'email' => 'foo@example.com' }
+      session[:foo] = 'bar'
 
       get :destroy
 
       assert_nil session[:shopify]
       assert_nil session[:shopify_domain]
       assert_nil session[:shopify_user]
+      assert_nil session[:foo]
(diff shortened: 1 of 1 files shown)


andyw8:
Hmm, I'm surprised we didn't have failing tests, are we missing some coverage?

greatestape:
It was only the tests that were out of date, and I think those tests were just testing that that key was missing after the operation under test.

So yeah, we had a coverage gap because we were testing for the removal of a key we no longer used, but the tests are a little vulnerable to getting out of date because they can test the wrong thing and still pass.

paulomarg:
I agree, this change looks right to me, but I feel like it may be worth adding a test that actually fails if `shop_id` isn't there.

greatestape:
@paulomarg I think we do have coverage of that. I think that value is set onto the rails session in the callback controller, and we've got tests there to cover it: https://github.com/Shopify/shopify_app/blob/master/test/controllers/callback_controller_test.rb#L42

Looking closer at the tests I updated, I think there are two deeper issues:

- "#new should trust the shop param over the current session"
  - There's no test of the inverse of this: "#new should fall back to the shop value in the current session if shop param missing"
  - That behaviour (using the value in the rails session during #new) doesn't seem to exist in the code, so this test is wrong and useless
- the #destroy tests
  - These are testing that a specific value gets cleared, which does happen, but the actual behaviour of the code is that the entire rails session is cleared. This is a broader behaviour and arguably our test should express that that's the behaviour it expects

If other folks can sanity check my findings and agree this makes sense, I can see about removing that first test and at least updating the language in the #destory tests.

paulomarg:
Both points make sense to me @greatestape, though I haven't checked the `using the value in the rails session during #new` part myself.

rezaansyed (approved):
LGTM

paulomarg (approved):

paulomarg (approved):
LGTM!

PR #1169, opened 2021-01-28
WebhookJobKlass.perform_later should add ** for Ruby 3.0 compatibility

andrewcsmith:
Ruby 3.0 removes the implicit conversion of Hash to kwargs, so we need to add ** to make this explicit.

Previously, all webhook hits would break since it would attempt to pass a positional Hash to a Job that only accepts keywords.

### DIFF
--- a/app/controllers/shopify_app/webhooks_controller.rb
+++ b/app/controllers/shopify_app/webhooks_controller.rb
@@ -8,7 +8,7 @@ module ShopifyApp
     def receive
       params.permit!
       job_args = { shop_domain: shop_domain, webhook: webhook_params.to_h }
-      webhook_job_klass.perform_later(job_args)
+      webhook_job_klass.perform_later(**job_args)
       head(:ok)
     end
 


app/controllers/shopify_app/webhooks_controller.rb
```diff
@@ -8,7 +8,7 @@ class WebhooksController < ActionController::Base
     def receive
       params.permit!
       job_args = { shop_domain: shop_domain, webhook: webhook_params.to_h }
-      webhook_job_klass.perform_later(job_args)
```
andyw8, line 11: Perhaps it would be clearer to inline the arguments i.e.:

`webhook_job_klass.perform_later(shop_domain: shop_domain, webhook: webhook_params.to_h)`

andrewcsmith: Yep, even better. Added to this PR.

andyw8 (approved):

oliwoodsuk:
> Hey guys. Is there any way this can be added to the next release? It looks like it may have got stuck.

Just bumped into this issue. Simple fix, but would be good to get merged.
I always find it a bit tricky to test webhooks coming from Shopify in Development, so didn't spot this until it was in Production.

PR #1172, opened 2021-01-30
Remove forward & add localtunnel to the tunneling services

3lackRos3:
ForwardHQ has shutdown, so I've removed it and added localtunnel npm package as I've using it personally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,7 +459,7 @@ App Tunneling
 
 Your local app needs to be accessible from the public Internet in order to install it on a Shopify store, to use the [App Proxy Controller](#app-proxy-controller-generator) or receive Webhooks.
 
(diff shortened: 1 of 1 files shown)


3lackRos3:
Have signed the CLA. Not sure how to run the CLA check again.

thecodepixi:
Hi 👋 Can you double check the CLA form? I re-ran the check but it still seems to be failing for you.

3lackRos3:
Hi @thecodepixi. 

I tried to submit the CLA again but I get this error message "GitHub username has already signed a CLA".

thecodepixi:
It looks like the actual commit is being associated with another account (199529 under "participants"). Is that also your account? That might be why the CLA is showing as incomplete.

3lackRos3:
It should be mine. I think it must have happened because I have the wrong email in "git config". Do you want me to close this and open another PR?

thecodepixi:
That would be great! I think we'd rather remove the broken link rather than replace it with something new right now. If it's easier for you, feel free to close this and open another to remove the link 👍

PR #1175, opened 2021-02-02
Allow use of omniauth v2

mkevinosullivan:


lib/generators/shopify_app/install/templates/omniauth.rb
```diff
@@ -1,3 +1,6 @@
 # frozen_string_literal: true
+OmniAuth.config.allowed_request_methods = [:post, :get]
+OmniAuth.config.silence_get_warning = true
```
andyw8, line 3: I wonder if we should have a comment in the generated code to explain why we silence the warning here...

mkevinosullivan: 👍🏻 see comment lines added

tanema: Also should we have a migration guide somewhere in the readme for this?

alexaitken: Is it possible to set these values for only `OmniAuth::Strategies::Shopify` instead of all strategies?

An app that uses multiple OmniAuth strategies may want to restrict those to only `POST` with CSRF protection.

mkevinosullivan: > Is it possible to set these values for only `OmniAuth::Strategies::Shopify` instead of all strategies?
> 
> An app that uses multiple OmniAuth strategies may want to restrict some to only `POST` with CSRF protection.

From what I can discern from https://github.com/omniauth/omniauth/blob/master/lib/omniauth/strategy.rb, I think this configuration is global only, and not on a per-provider basis.

PR #1176, opened 2021-02-03
Fixed bug in generated products controller.

yourivdlans:
After generating a new app with Rails 6.0.3.4 and shopify_app 17.0.5 (shopify_api 9.3.0) the example would not work as javascript would throw the following error: `products.forEach is not a function`.

Looking at the response for "/products" it shows a Ruby object instead of JSON.

Screen Shot 2021-02-03 at 14 47 46
(Test data)

Making sure the products returned are JSON using `as_json` fixes the issue.

### DIFF
--- a/lib/generators/shopify_app/products_controller/templates/products_controller.rb
+++ b/lib/generators/shopify_app/products_controller/templates/products_controller.rb
@@ -3,6 +3,6 @@
 class ProductsController < AuthenticatedController
   def index
     @products = ShopifyAPI::Product.find(:all, params: { limit: 10 })
-    render(json: { products: @products })
+    render(json: { products: @products.map(&:as_json) })
   end
 end


andyw8:
Although this may fix your specific problem, it shouldn't really be necessary – the serialization to JSON should be automatic. It may that the default behaviour or Rails or a gem has changed recently. We'll need to investigate.

paulomarg:
I've started an app from scratch with the same versions and did not get this issue. Here are my steps:

- `rails new`
- Add `shopify_app` to Gemfile
- `rails generate shopify_app`
- `rails db:migrate`

Have you done anything differently?

yourivdlans:
@pauldowman Thats what I did as well. Could it be that you're using an older version of the `shopify_api` gem?
I had `9.3.0` installed.

andyw8:
@yourivdlans I think you meant to reply to @paulomarg

andyw8:
If you call `/products.json` rather than `/products` does it make any difference?

yourivdlans:
@andyw8 I did, I've updated my comment.

There is no difference in using `/products` or `/products.json`.

I've also tried downgrading the `shopify_api` gem from 9.3.0 to 9.2.0 but still the same result.

andyw8:
@yourivdlans is it possible to share your `Gemfile.lock`? That would allow us to check for any possible gem version differences in the dependencies.

yourivdlans:
@andyw8 Of course, here it is: https://gist.github.com/yourivdlans/464cc9161395f83afcf6f5d6e8c15c1b

paulomarg:
> @paulomarg Thats what I did as well. Could it be that you're using an older version of the `shopify_api` gem?
> I had `9.3.0` installed.

I made sure to use the same versions you pointed out, so this is quite curious. Thanks for the lock file, I will use that instead of mine to see if I get the issue as well.

paulomarg:
It seems like this may be related to Ruby 3 somehow - I used your Gemfile with v2 and it worked, but it failed once I moved over to v3. I didn't dig into the reasons for it, but a possible workaround would be to use a previous version of Ruby.

yourivdlans:
Oh interesting, reverting to Ruby 2.x fixes the issue! Thanks for taking the time to help me out.

andyw8:
I've created https://github.com/Shopify/shopify_app/issues/1178 to track this.

PR #1180, opened 2021-02-05
Introduce ScopesVerification for handle changes in scopes requested by app

rezaansyed:
## What is the problem?
shopify_app currently does not have any mechanism for handling changes in scopes requested by an app. Changes in scopes requested were previously resolved once cookies were expired, forcing merchants to go through OAuth and grant the net new scopes.

With session tokens, we need to keep track of the current scopes for the merchant and trigger OAuth when there is a difference between the requested scopes and the current scopes.


app/controllers/concerns/shopify_app/scopes_verification.rb
```diff
@@ -0,0 +1,96 @@
...
+      start = Time.now
+      access_scopes = yield
+      finish = Time.now
+
+      f = File.new(filename, 'a')
+      f << "#{finish - start}\n"
+      f.close
+
+      access_scopes
+    end
```
NabeelAhsen, line 44: The measurement code used to profile the different methods

README.md
```diff
@@ -132,6 +132,22 @@ This generator creates an example home controller and view which fetches and dis
 Options include:
 * __[Not recommended for embedded apps]__ `with-cookie-authentication` - This flag generates an authenticated home_controller, where the authentication strategy relies on cookies. By default, this generator creates an unauthenticated home_controller and protected sample products_controller.
 
+### Scopes Storage Generator
+```sh
+$ rails generate shopify_app:scopes_storage
+```
+
+This generator does the following:
```
paulomarg, line 140: If I have the right end of the stick, this isn't needed for non-embedded apps, since cookies will expire eventually and trigger a re-auth, right? If that's the case, should we mention that at all here?

README.md
```diff
@@ -132,6 +132,22 @@ This generator creates an example home controller and view which fetches and dis
 Options include:
 * __[Not recommended for embedded apps]__ `with-cookie-authentication` - This flag generates an authenticated home_controller, where the authentication strategy relies on cookies. By default, this generator creates an unauthenticated home_controller and protected sample products_controller.
 
+### Scopes Storage Generator
```
paulomarg, line 135: Not at all related to this PR, but this README is starting to feel too long to me - do we want to look into breaking it up into smaller pages?

README.md
```diff
@@ -300,6 +316,44 @@ class AuthenticatedController < ApplicationController
...
 
+### ScopesVerification
+The `ShopifyApp::ScopesVerification` concern helps merchant grant new access scopes requested by the app. The concern compares the current access scopes granted by the merchant and compares it with the scopes requested by the app. If there is a mismatch in configuration, the merchant is redirected to login via OAuth and grant the net new scopes.
+
+This requires the app to maintain a record of the current access scopes per merchant. This can be done by defining the template class methods in the model that includes `ShopifyApp::ShopSessionStorage`:
+```ruby
+class Shop < ActiveRecord::Base
+  include ShopifyApp::ShopSessionStorage
+
+  def self.update_merchant_scopes(shop, scopes)
```
paulomarg, line 327: While I like this pattern and how it allows us to add this behaviour only if the migration is set up, I wonder if we couldn't use `scopes=` and `scopes` as our method names, which feels like it might be easier to understand. Would that work at all?

app/controllers/concerns/shopify_app/scopes_verification.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module ScopesVerification
+    extend ActiveSupport::Concern
+    include ScopeUtilities
```
paulomarg, line 6: I wonder if it's worth isolating this code in its own class, do we expect to use it in other places?

app/controllers/concerns/shopify_app/scopes_verification.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    included do
+      before_action do
+        login_on_scope_changes(current_merchant_scopes, configuration_scopes)
+      end
+    end
+
+    protected
+
+    def login_on_scope_changes(current_merchant_scopes, configuration_scopes)
```
paulomarg, line 16: Would it make sense not to pass in the scope sets as params here?

lib/generators/shopify_app/scopes_storage/scopes_storage_generator.rb
```diff
@@ -0,0 +1,53 @@
...
+  module Generators
+    class ScopesStorageGenerator < Rails::Generators::Base
+      include Rails::Generators::Migration
+      source_root File.expand_path('../templates', __FILE__)
+
+      class_option :with_cookie_authentication, type: :boolean, default: false
+      class_option :embedded, type: :string, default: 'true'
+
+      def create_scopes_storage_in_shop_model
+        if yes?("Do you want to add scopes column to the Shop model? [y/n]")
```
paulomarg, line 15: I wonder if this prompt may seem unclear to a developer without context, but I can't think of a suggestion other than mentioning it's recommended for embedded apps.

lib/generators/shopify_app/scopes_storage/scopes_storage_generator.rb
```diff
@@ -0,0 +1,53 @@
...
+          migration_template('db/migrate/add_scopes_column.erb', 'db/migrate/add_scopes_column.rb')
+          copy_file('shop_with_scopes.rb', 'app/models/shop.rb')
+        end
+      end
+
+      def include_scopes_verification_in_home_controller
+        template('home_controller.rb', 'app/controllers/home_controller.rb') unless with_cookie_authentication?
```
paulomarg, line 22: Would it be possible to introduce the new concern via a regexp rather than replacing the entire file? It seems to me like this leads us to maintain multiple copies of the home controller, which might be error prone.

paulomarg (commented):
Logic-wise everything makes sense to me! Most of my comments are just questions that popped into my head, not necessarily changes we want to make!

lib/generators/shopify_app/scopes_storage/scopes_storage_generator.rb
NabeelAhsen: Hmm, maybe this string could provide a little more context as to why this might be needed for some developers.

```suggestion
        scopes_column_prompt = <<~PROMPT
          It is highly recommended that embedded apps store the access scopes granted by merchants during app installation. The following migration will add a `scopes` column to the Shop model.
          
          Do you want to add this migration? [y/n]
        PROMPT
        
        if yes?(scopes_column_prompt)
```

app/controllers/concerns/shopify_app/scopes_verification.rb
NabeelAhsen: I think the intent is to allow partners the freedom to specify their own sources for these strings. I don't know of a better way to offer this flexibility, or whether it's the right move.

README.md
NabeelAhsen: We have an issue that will look to break this README up very soon. Always looking for suggestions on how to best split this doc up!

app/controllers/concerns/shopify_app/scopes_verification.rb
paulomarg: Makes a lot of sense to me! IMHO if they add value we should keep them.

lib/generators/shopify_app/scopes_storage/scopes_storage_generator.rb
NabeelAhsen: I think that part of this PR should help outline a clear path forward that makes this the default output of our generators in future versions.

lib/generators/shopify_app/scopes_storage/templates/home_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class HomeController < ApplicationController
+  include ShopifyApp::EmbeddedApp
+  include ShopifyApp::RequireKnownShop
+  include ShopifyApp::ScopesVerification
```
NabeelAhsen, line 6: I believe templates work with erb templating code as well. Something like 
```suggestion
  <%= 'include ShopifyApp::ScopesVerification' if user_answered_yes_to_prompt? %>
```

NabeelAhsen: This variable should be configurable in the install generator class or the method itself that calls `template(src, dest)`

NabeelAhsen, line 6: Following the rule of threes: We should really look into aggregating this into a single concern, kind of like `ShopifyApp::Authenticated` in a future release.

This aggregate concern would be basically `ShopifyApp::Unauthenticated` or something.

README.md
rezaansyed: We could include it for non-embedded apps as well because technically this is an issue that plagues all apps. The current way for any app to request new scopes is to wait for cookie expiry or rotate the app secret. I'm just not sure if the HomeController is the right place to do this check?

README.md
rezaansyed: Yes! We could probably break down the sections into different files?

app/controllers/concerns/shopify_app/scopes_verification.rb
rezaansyed: Thinking about this, I think it doesn't give them freedom with the current approach. Because the concern already has the `before_action` which calls the `current_merchant_scopes` and `configuration_scopes ` methods. They would have to override those methods in order to customize.

README.md
paulomarg: That's true. I imagine that for non-embedded apps this might fit better into `AuthenticatedController`?

lib/generators/shopify_app/scopes_storage/scopes_storage_generator.rb
rezaansyed: I just pushed up a change introducing the `scopes` and `scopes=` instance methods for the Shop model.

lib/generators/shopify_app/scopes_storage/templates/home_controller.rb
rezaansyed: Yeah that makes sense to me!

README.md
rezaansyed: I just pushed up a change introducing the `scopes` and `scopes=` instance methods for the Shop model.

README.md
```diff
@@ -300,6 +316,44 @@ class AuthenticatedController < ApplicationController
 end
 ```
 
+### ScopesVerification
+The `ShopifyApp::ScopesVerification` concern helps merchant grant new access scopes requested by the app. The concern compares the current access scopes granted by the merchant and compares it with the scopes requested by the app. If there is a mismatch in configuration, the merchant is redirected to login via OAuth and grant the net new scopes.
```
greatestape, line 320: how about "helps the app request new scopes from the merchant"

README.md
```diff
@@ -300,6 +316,44 @@ class AuthenticatedController < ApplicationController
...
+
+This requires the app to maintain a record of the current access scopes per merchant. This can be done by defining the template class methods in the model that includes `ShopifyApp::ShopSessionStorage`:
```
greatestape, line 322: "per shop"

paulomarg (approved):
LGTM!

lib/generators/shopify_app/scopes_storage/scopes_storage_generator.rb
```diff
@@ -0,0 +1,70 @@
...
+      include Rails::Generators::Migration
+      source_root File.expand_path('../templates', __FILE__)
+
+      class_option :with_cookie_authentication, type: :boolean, default: false
+      class_option :embedded, type: :string, default: 'true'
+
+      def create_scopes_storage_in_shop_model
+        scopes_column_prompt = <<~PROMPT
+          It is highly recommended that apps record the access scopes granted by\
+          merchants during app installation.
```
NabeelAhsen, line 17: Is the `\` intentional?

PR #1188, opened 2021-02-17
Add warning to Omniauth generator if shopify_provider exists

NabeelAhsen:
This is a PR that will help unblock us on the upcoming [scope changes feature](https://github.com/Shopify/shopify_app/tree/scope-strategies). This PR adds a warning to the `InstallGenerator` when injecting a shopify_provider to the `config/initializers/omniauth.rb` file.

The reason this is needed is so that we can warn developers who have an existing app and are running the `InstallGenerator`. If a shopify provider already exists in their `omniauth.rb`, we do not want to over-write it; instead, we should output a warning, and let developers know that an injection was attempted:

```
[WARNING] The Shopify App generator attempted to add the following Shopify Omniauth provider 'config/initializers/omniauth.rb':

# frozen_string_literal: true

provider :shopify,
  scope: ShopifyApp.configuration.scope,
  setup: lambda { |env|
     # New changes made to this provider
  }


Consider updating 'config/initializers/omniauth.rb' to match the configuration above.
```


rezaansyed (approved):
LGTM 👍🏼

lib/generators/shopify_app/install/install_generator.rb
```diff
@@ -70,8 +72,37 @@ def insert_hosts_into_development_config
         )
       end
 
+      protected
```
andyw8, line 75: Why `protected` rather than `private`?

lib/generators/shopify_app/install/install_generator.rb
```diff
@@ -70,8 +72,37 @@ def insert_hosts_into_development_config
...
+
+      def shopify_provider_exists
```
andyw8, line 77:
```suggestion
      def shopify_provider_exists?
```

lib/generators/shopify_app/install/install_generator.rb
NabeelAhsen: Oh good catch, this was leftover from testing. It should be private!

rezaansyed (approved):
LGTM

andyw8 (approved):

PR #1190, opened 2021-02-19
Create OmniAuthConfiguration object to build future OmniAuth strategies

NabeelAhsen:
This PR introduces the following:
1.  A new `OmniAuthConfiguration` object
2. New configurations for `ShopifyApp.configuration`
  - `shop_access_scopes`: These default to the old `ShopifyApp.configuration.scope` config
  - `user_access_scopes`: These default to the old `ShopifyApp.configuration.scope` config

The intent is to use this object in the future to help us define and build omniauth strategies. This will be especially important in upcoming changes related to scope changes.

Here's how partners can use this in their `omniauth.rb` files:

```ruby
Rails.application.config.middleware.use(OmniAuth::Builder) do
  provider :foo
      api_key,
      secret,
      scope: 'read_products',
      setup: lambda { |env|
        configuration = ShopifyApp::OmniauthConfiguration.new(env['omniauth.strategy'], Rack::Request.new(env))
        configuration.build_options
      }
end
```

For now, this class is not being used anywhere else in the library. A future PR will introduce default strategies for partners who use offline access tokens saved in their Shop records and online access tokens saved in their User records.

Based on the parent branch: https://github.com/Shopify/shopify_app/compare/scope-strategies


lib/shopify_app/omniauth/omniauth_configuration.rb
```diff
@@ -0,0 +1,66 @@
...
+      return '' unless shop_domain.present?
+      "https://#{shopify_auth_params[:shop]}"
+    end
+
+    def default_request_online_tokens?
+      strategy.session[:user_tokens] && !update_shop_scopes?
+    end
+
+    def update_shop_scopes?
+      ShopifyApp.configuration.shop_access_scopes_strategy.scopes_mismatch?(shop_domain)
```
rezaansyed, line 53: We can leave out the use of strategies in this PR. Let's just have the current default behaviour working.

rezaansyed (approved):
🍞

lib/shopify_app.rb
```diff
@@ -58,4 +58,7 @@ def self.use_webpacker?
   require 'shopify_app/session/session_storage'
   require 'shopify_app/session/shop_session_storage'
   require 'shopify_app/session/user_session_storage'
+
+  # omniauth_configuration
```
andyw8, line 62: Unnecessary comment?

lib/shopify_app/configuration.rb
```diff
@@ -81,6 +83,14 @@ def has_scripttags?
     def enable_same_site_none
       !Rails.env.test? && (@enable_same_site_none.nil? ? embedded_app? : @enable_same_site_none)
     end
+
+    def shop_access_scopes
+      @shop_access_scopes || scope
```
andyw8, line 88: It seems a little odd to return a singular item `scope` when the method name implies a collection.

rezaansyed: Agreed! However, the `scope` attribute has been used to define a collection of scopes stored in a string for a while now. Changing this will be a breaking change for shopify_app developers.

lib/shopify_app.rb
rezaansyed: We wanted to be categorize the `OmniAuthConfiguration` class similar to other modules and classes above.

NabeelAhsen: We followed the convention of the file, which seems to have similar comments breaking requirements apart. Do you think this belongs with an existing collection within the same file?

andyw8: Ah, I was only seeing the excerpt. Let's stick with that convention then.

andyw8 (approved):

lib/shopify_app/configuration.rb
NabeelAhsen: Good point, we can mark this as future work to update `scope` to `scopes`.

jerrymendoza:
is this implementation still valid? I can't find any reference to OmniAuthConfiguration, I would like to modify ShopifyAPI::Context in runtime

PR #1191, opened 2021-02-22
Add access scopes to Shop and User model generators

rezaansyed:
### Problem
Embedded apps using session tokens do not automatically handle changes in access scopes for offline and online tokens. In order to handle changes to access scopes requested, we need to store the access scopes of access tokens.

### What does this PR do?
This PR creates optional database migrations in the generators to allow apps to store access scopes in the Shop and User models.  

#### Run `rails g shopify_app:shop_model` 
  - You should be prompted to add `AddShopAccessScopesColumn` migration
  - Prompt yes to generate migration
  - Run `rails db:migrate` to successfully add new column to Shop model
#### Run `rails g shopify_app:user_model` 
  - You should be prompted to add `AddUserAccessScopesColumn` migration
  - Prompt yes to generate migration
  - Run `rails db:migrate` to successfully add new column to User model

### DIFF
--- a/lib/generators/shopify_app/shop_model/shop_model_generator.rb
+++ b/lib/generators/shopify_app/shop_model/shop_model_generator.rb
@@ -8,12 +8,35 @@ module ShopifyApp
       include Rails::Generators::Migration
       source_root File.expand_path('../templates', __FILE__)
 
+      class_option :force_access_scopes_migration, type: :boolean, default: false
+
       def create_shop_model
         copy_file('shop.rb', 'app/models/shop.rb')
       end
 
       def create_shop_migration
         migration_template('db/migrate/create_shops.erb', 'db/migrate/create_shops.rb')
+      end
+
+      def create_shop_with_access_scopes_migration
+        scopes_column_prompt = <<~PROMPT
+          It is highly recommended that apps record the access scopes granted by \
+          merchants during app installation. See app/models/shop.rb to modify how \
+          access scopes are stored and retrieved.
+
+          [WARNING] You will need to update the access_scopes accessors in the Shop model \
+          to allow shopify_app to store and retrieve scopes when going through OAuth.
+
+          The following migration will add an `access_scopes` column to the Shop model. \
+          Do you want to include this migration? [y/n]
+        PROMPT
+
+        if force_access_scopes_migration? || Rails.env.test? || yes?(scopes_column_prompt)
+          migration_template(
+            'db/migrate/add_shop_access_scopes_column.erb',
+            'db/migrate/add_shop_access_scopes_column.rb'
+          )
+        end
       end
 
       def update_shopify_app_initializer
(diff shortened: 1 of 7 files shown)


NabeelAhsen (approved):

lib/generators/shopify_app/shop_model/shop_model_generator.rb
```diff
@@ -16,6 +16,26 @@ def create_shop_migration
         migration_template('db/migrate/create_shops.erb', 'db/migrate/create_shops.rb')
       end
 
+      def create_shop_with_access_scopes_migration
+        scopes_column_prompt = <<~PROMPT
+          It is highly recommended that apps record the access scopes granted by \
+          merchants during app installation. See app/models/shop.rb to modify how \
+          access scopes are stored and retrieved.
+
+          [WARNING] Your app will fail the ScopesVerification concern if access scopes are not stored.
```
paulomarg, line 25: Does anything use this by default? If so, we may want to mention that in the message as well, so we don't 'scare' people into accepting it?

lib/generators/shopify_app/user_model/user_model_generator.rb
```diff
@@ -16,6 +16,27 @@ def create_user_migration
...
       end
 
+      def create_scopes_storage_in_user_model
+        scopes_column_prompt = <<~PROMPT
+          It is highly recommended that apps record the access scopes granted by \
+          merchants during app installation. See app/models/user.rb to modify how \
+          access scopes are stored and retrieved.
+
+          [WARNING] There may be unintended behaviour during the OAuth flow \
+          if access_scopes are not stored.
```
paulomarg, line 26: This feels a little vague without the context of what exactly might fail, can we be a bit more specific without making the message confusing?

paulomarg (changes requested):
🎩 ed  and it runs nicely on a pre-existing app.

However, when creating an app with the CLI, it intercepts the output of `bundle install`, so the prompt never appears to the user. I did verify that typing `y` and entering will still work as expected, so we just need a way to skip this prompt in the CLI.

If we can add a flag to force the column to be added when running `bundle install`, we can update the CLI to pass in that flag so the process can stay the same.

lib/generators/shopify_app/shop_model/shop_model_generator.rb
rezaansyed: I can provide a better reason for it.

lib/generators/shopify_app/user_model/user_model_generator.rb
rezaansyed: Updated the messaging around this.

paulomarg (approved):
LGTM, we just need to make sure this isn't released until the CLI is updated to pass in the `force_access_scopes_migration` flag.

rezaansyed:
> LGTM, we just need to make sure this isn't released until the CLI is updated to pass in the `force_access_scopes_migration` flag.

Yep will make sure to sync on that prior to releasing. Thanks Paulo!

NabeelAhsen:
@rezaansyed  @paulomarg  I can help look into the CLI flag issue early this week.

PR #1192, opened 2021-02-22
Store access scopes after successful OAuth

rezaansyed:
### Problem
Embedded apps using session tokens do not automatically handle changes in access scopes for offline and online tokens. In order to handle changes to access scopes requested, we need to store the access scopes after successfully completing OAuth flows.

### What does this PR do?
On successful completion of OAuth, apps will now store the access scopes of an access token in `ShopifyApp::ShopSessionStorageWithScopes` and `ShopifyApp::UserSessionStorageWithScopes`.

#### Go through OAuth for offline/shop tokens
- Appropriate Shop record should have `access_scopes` with the expected values

#### Go through OAuth for online/user tokens
- Appropriate User record should have `access_scopes` with the expected values

### DIFF
--- a/lib/generators/shopify_app/shop_model/templates/shop.rb
+++ b/lib/generators/shopify_app/shop_model/templates/shop.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 class Shop < ActiveRecord::Base
-  include ShopifyApp::ShopSessionStorage
+  include ShopifyApp::ShopSessionStorageWithScopes
 
   def api_version
     ShopifyApp.configuration.api_version
--- a/lib/shopify_app/session/in_memory_user_session_store.rb
+++ b/lib/shopify_app/session/in_memory_user_session_store.rb
@@ -1,14 +1,16 @@
 # frozen_string_literal: true
 module ShopifyApp
   class InMemoryUserSessionStore < InMemorySessionStore
-    def self.store(session, user)
-      id = super
-      repo[user.shopify_user_id] = session
-      id
-    end
+    class << self
+      def store(session, user)
+        id = super
+        repo[user.shopify_user_id] = session
+        id
+      end
 
-    def self.retrieve_by_shopify_user_id(user_id)
-      repo[user_id]
+      def retrieve_by_shopify_user_id(user_id)
+        repo[user_id]
+      end
     end
   end
 end
--- a/lib/shopify_app/session/shop_session_storage_with_scopes.rb
+++ b/lib/shopify_app/session/shop_session_storage_with_scopes.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module ShopSessionStorageWithScopes
+    extend ActiveSupport::Concern
+    include ::ShopifyApp::SessionStorage
+
+    included do
+      validates :shopify_domain, presence: true, uniqueness: { case_sensitive: false }
+    end
+
+    class_methods do
+      def store(auth_session, *_args)
+        shop = find_or_initialize_by(shopify_domain: auth_session.domain)
+        shop.shopify_token = auth_session.token
+        shop.access_scopes = auth_session.access_scopes
+
+        shop.save!
+        shop.id
+      end
+
+      def retrieve(id)
+        shop = find_by(id: id)
+        construct_session(shop)
+      end
+
+      def retrieve_by_shopify_domain(domain)
+        shop = find_by(shopify_domain: domain)
+        construct_session(shop)
+      end
+
+      private
+
+      def construct_session(shop)
+        return unless shop
+
+        ShopifyAPI::Session.new(
+          domain: shop.shopify_domain,
+          token: shop.shopify_token,
+          api_version: shop.api_version,
+          access_scopes: shop.access_scopes
+        )
+      end
+    end
+
+    def access_scopes=(scopes)
+      super(scopes)
+    rescue NotImplementedError, NoMethodError
+      raise NotImplementedError, "#access_scopes= must be defined to handle storing access scopes: #{scopes}"
+    end
+
+    def access_scopes
+      super
+    rescue NotImplementedError, NoMethodError
+      raise NotImplementedError, "#access_scopes= must be defined to hook into stored access scopes"
+    end
+  end
+end
--- a/test/shopify_app/session/session_repository_test.rb
+++ b/test/shopify_app/session/session_repository_test.rb
@@ -45,47 +45,40 @@ module ShopifyApp
     test '.retrieve_user_session_by_shopify_user_id retrieves a user session by JWT' do
       SessionRepository.user_storage = InMemoryUserSessionStore
 
-      user_id = 'abra'
-      InMemoryUserSessionStore.expects(:retrieve_by_shopify_user_id).with(user_id)
+      InMemoryUserSessionStore.expects(:retrieve_by_shopify_user_id).with(mock_shopify_user_id)
 
-      SessionRepository.retrieve_user_session_by_shopify_user_id(user_id)
+      SessionRepository.retrieve_user_session_by_shopify_user_id(mock_shopify_user_id)
     end
 
     test '.retrieve_shop_session_by_shopify_domain retrieves a shop session by JWT' do
       SessionRepository.shop_storage = InMemoryShopSessionStore
 
-      shopify_domain = 'abra-shop'
-      InMemoryShopSessionStore.expects(:retrieve_by_shopify_domain).with(shopify_domain)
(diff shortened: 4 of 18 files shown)


rezaansyed:
Tophat of app with offline and online tokens:

https://user-images.githubusercontent.com/60748788/108783604-c4310800-753b-11eb-8ad5-b9bed1f28432.mov

rezaansyed:
Tophatting for backwards compatibility where existing apps don't have access tokens defined on the Shop or User model. There should not be any access scopes stored.

https://user-images.githubusercontent.com/60748788/108785764-12480a80-7540-11eb-8d94-b43e544a7b0a.mov

lib/shopify_app/session/shop_session_storage.rb
```diff
@@ -26,15 +28,26 @@ def retrieve_by_shopify_domain(domain)
...
+
       private
 
       def construct_session(shop)
         return unless shop
+        begin
+          scopes = shop.access_scopes
+        rescue NotImplementedError, NoMethodError
+          scopes = nil
+        end
```
NabeelAhsen, line 44: I forget what our original contention was around doing this. Is it still feasible to move this to its own method?

```suggestion
  def scopes(shop)
    begin
      shop.access_scopes
    rescue NotImplementedError, NoMethodError
      nil
    end
  end
```

lib/shopify_app/session/user_session_storage.rb
```diff
@@ -27,14 +28,31 @@ def retrieve_by_shopify_user_id(user_id)
...
       private
 
       def construct_session(user)
         return unless user
+        begin
+          scopes = user.access_scopes
+        rescue NotImplementedError, NoMethodError
+          scopes = nil
+        end
+
```
NabeelAhsen, line 50: Similarly here as well?

```ruby
def scopes(user)
  begin
    user.access_scopes
  rescue
    nil
  end
end

def construct_session(user)
  return unless user
  
  Shopify::Session.new(
    domain: user.shopify_domain,
    token: user.shopify_token,
    api_version: user.api_version,
    extra: { scopes: scopes(user) }
  )
end
```

NabeelAhsen (approved):
Non-blocking comments, looks good!

lib/shopify_app/session/shop_session_storage.rb
rezaansyed: Sure! I see no issues with this!

lib/shopify_app/session/in_memory_user_session_store.rb
```diff
@@ -1,14 +1,24 @@
...
+        id = super
+        repo[user.shopify_user_id] = session
+        id
+      end
+
+      def retrieve_access_scopes(id)
+        repo[id].extra[:scopes]
+      end
+
+      def retrieve_access_scopes_by_shopify_user_id(shopify_user_id)
```
ragalie, line 15: This interface seems weird to me. Why do we need to expand the interface like this? Don't we already return the scopes as part of the session?

lib/generators/shopify_app/shop_model/templates/shop.rb
```diff
@@ -2,6 +2,16 @@
 class Shop < ActiveRecord::Base
   include ShopifyApp::ShopSessionStorage
 
+  # Override access_scopes= to specify how access scopes are saved
+  def access_scopes=(scopes)
```
ragalie, line 6: The other default implementations of session storage methods are on `ShopSessionStorage`. Why put these here instead of there?

rezaansyed: This is to account for existing apps that already have access scopes stored and don't want to use the new migration we provide. For example, if they have an `AccessScopes` table, they would override these methods as follows

```ruby
class Shop < ActiveRecord::Base
  include ShopifyApp::ShopStorage

  def access_scopes=(scopes)
    access_scopes = AccessTokens.find_or_initialize_by(self.shopify_domain)
    AccessTokens.all_scopes = scopes
    access_scopes.save!
  end

  def access_scopes
   record = AccessScopes.find_by(shopify_domain: self.shopify_domain)
   record.all_scopes
  end
end
```

lib/shopify_app/session/in_memory_user_session_store.rb
rezaansyed: It forces the caller to know the internals of the `ShopifyAPI::Session` object. We don't have an `access_scopes` field on there so we have to pass it into the `extra` hash. So accessing it would be something like this:
```ruby
session.extra[:scopes]
```

ragalie: Then I think we should expand the `Session` interface to add an `access_scopes` attribute since it's a logical part of that object, imo.

lib/generators/shopify_app/shop_model/templates/shop.rb
ragalie: Oh no, I love that we've added this to the interface.

My question is purely about code organization; why is this code here instead of in `ShopSessionStorage` like the implementations of all the other methods (e.g. `store`, `retrieve`)

lib/shopify_app/session/session_storage.rb
```diff
@@ -16,5 +16,17 @@ def with_shopify_session(&block)
         &block
       )
     end
+
+    def access_scopes=(scopes)
+      super(scopes)
+    rescue NotImplementedError, NoMethodError
+      Rails.logger.warn("#access_scopes= must be defined to handle storing access scopes: #{scopes}")
```
ragalie, line 23: I don't think this is a great pattern. Now a) the rest of the system needs to support both the "can store access scopes" and "can not store access scopes" states indefinitely and b) the developer might think they properly implemented access scope support and the only sign that they messed it up is `warn` error messages.

It would be better IMO to create a `ShopSessionStorageWithScopes` interface that adds the `access_scopes` support (and errors if it doesn't work) and then to change behaviour elsewhere based on which interface is active. Then we make a breaking change in a future major version where we say we're only supporting the new interface.

rezaansyed: This makes sense. I'll refactor this piece. Thanks again Mike!

lib/shopify_app/session/in_memory_user_session_store.rb
rezaansyed: Covering that in this [PR](https://github.com/Shopify/shopify_api/pull/843)

lib/shopify_app/session/shop_session_storage_with_scopes.rb
```diff
@@ -0,0 +1,58 @@
...
+          api_version: shop.api_version,
+          access_scopes: shop.access_scopes
+        )
+      end
+    end
+
+    def access_scopes=(scopes)
+      super(scopes)
+    rescue NotImplementedError, NoMethodError => exception
+      raise exception.class, "#access_scopes= must be defined to handle storing access scopes: #{scopes}"
```
ragalie, line 49: We don't do this with the other assumed methods (like `find_by`), but I think this is a nice and convenient for them. I would standardize on a particular error class though, either `NotImplementedError` or a custom error.

ragalie (approved):
LGTM, comment is not a blocker

paulomarg (approved):
LGTM

test/shopify_app/session/session_repository_test.rb
```diff
@@ -94,15 +87,32 @@ class SessionRepositoryTest < ActiveSupport::TestCase
...
+    end
+
+    def mock_shopify_session
+      ShopifyAPI::Session.new(
+        domain: mock_shopify_domain,
+        token: 'abracadabra',
+        api_version: :unstable,
+        extra: {
+          scopes: 'read_products',
+        }
```
NabeelAhsen, line 114: Is this still used anywhere in this file?

lib/shopify_app/session/shop_session_storage_with_scopes.rb
NabeelAhsen: I like the idea of raising `NotImplementedError` here

NabeelAhsen (approved):
LGTM!

We should roadmap future work to only support the new SessionStorage interface before a major release.

lib/shopify_app/session/shop_session_storage_with_scopes.rb
rezaansyed: Yeah `NotImplementedError` makes sense to me

test/shopify_app/session/session_repository_test.rb
rezaansyed: Yeah I just forgot to update this. Good catch!

PR #1193, opened 2021-02-23
Add new_shopify_cli_app flag for bypassing prompts when using shopify…

rezaansyed:
Related to this: https://github.com/Shopify/shopify-app-cli/pull/1086#discussion_r581119425

### DIFF
--- a/lib/generators/shopify_app/shop_model/shop_model_generator.rb
+++ b/lib/generators/shopify_app/shop_model/shop_model_generator.rb
@@ -8,7 +8,7 @@ module ShopifyApp
       include Rails::Generators::Migration
       source_root File.expand_path('../templates', __FILE__)
 
-      class_option :force_access_scopes_migration, type: :boolean, default: false
+      class_option :new_shopify_cli_app, type: :boolean, default: false
 
       def create_shop_model
         copy_file('shop.rb', 'app/models/shop.rb')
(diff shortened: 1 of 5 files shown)


NabeelAhsen:
This flag should likely be available to the install_generator as well, because it's responsible for calling `shop_model` generator.

paulomarg (approved):
LGTM, just checking: is the standard here to use snake_case for the parameter name? The Polaris guidelines point towards preferring kebab-case for command line arguments.

NabeelAhsen:
> LGTM, just checking: is the standard here to use snake_case for the parameter name? The Polaris guidelines point towards preferring kebab-case for command line arguments.

Hey Paulo, I'll be taking over this PR. Thank you for point that out! I'll make this kebab-case + there are a few more changes needed + tests written before this should ship.

NabeelAhsen:
> LGTM, just checking: is the standard here to use snake_case for the parameter name? The Polaris guidelines point towards preferring kebab-case for command line arguments.

So it turns out that the options can be passed in on the command line using kebab-case! Rubocop complains if any variable isn't snake-case, so there appear to be some magic that rewrites these options into underscored symbols.

So something like this works for me locally:

```terminal
$ rails generate shopify_app --new-shopify-cli-app
```

lib/generators/shopify_app/shopify_app_generator.rb
```diff
@@ -9,7 +9,7 @@ def initialize(args, *options)
 
       def run_all_generators
         generate("shopify_app:install #{@opts.join(' ')}")
-        generate("shopify_app:shop_model")
+        generate("shopify_app:shop_model #{@pts.join(' ')}")
```
rezaansyed, line 12: should be `@opts`

lib/generators/shopify_app/shop_model/shop_model_generator.rb
```diff
@@ -49,8 +49,8 @@ def create_shop_fixtures
 
       private
 
-      def force_access_scopes_migration?
-        options['force_access_scopes_migration']
+      def new_shopify_cli_app?
+        options['new_shopify_cli_app']
```
paulomarg, line 53: Interesting, I didn't know Rails handled converting the param format.

paulomarg (approved):

PR #1197, opened 2021-02-25
Use access scopes strategies to handle changes to scopes requested by app

rezaansyed:
## Problem
Embedded apps using session tokens do not automatically handle changes in access scopes for offline and online tokens. In order to handle changes to access scopes requested, we need to determine when to trigger OAuth to grant new scopes.

## What does this PR do?
This PR introduces access scopes strategies, which will be used in the gem to determine whether new scopes need to be granted for a shop and/or user. The following strategies are set by default as part of our `ShopifyApp::Configuration`:

```ruby
ShopifyApp.configuration.shop_access_scopes_strategy = ShopifyApp::AccessScopes::ShopStrategy
ShopifyApp.configuration.user_access_scopes_strategy = ShopifyApp::AccessScopes::UserStrategy
```

It should be noted that `shop_access_scopes_strategy` and `user_access_scopes_strategy` are read-only.

These strategies are currently used to determine whether scopes are mismatched in three scenarios.

### When loading the app
We use the `ShopifyApp::ShopAccessScopesVerification` concern at the `HomeController` to check whether a shop has all the expected access scopes as in the configuration. It will redirect to OAuth if scopes are mismatched and request the updated list of scopes.

handle-scopes-on-app-load

### In the OmniAuth request middleware
When building the OmniAuth strategy, we need to determine whether the shop/offline token needs to be updated based on changes in scopes.

omniauth-middleware-scope-handling

### In the callback controller if a user/online token needs to be updated
After completing the OAuth flow to get a shop/offline token, an app may have a valid user/online token that doesn't have the appropriate scopes. Based on the `ShopifyApp::AccessScopes::UserStrategy`, we trigger OAuth to update the user's access scopes on mismatches.

handle-user-access-scopes-in-callback-controller

#### For app using offline tokens only
1. Run `rails server` and load app for the first time with the first set of access scopes.
2. Shut down `rails server`
3. In `shopify_app.rb` on the app, change `config.scope` or `config.shop_access_scopes` to new set of scopes
4. Run `rails server` and load app
5. You will sent through to the OAuth flow before coming back to the app with updated scopes

#### For app using offline and online tokens
1. Run `rails server` and load app for the first time with the first set of access scopes.
2. Shut down `rails server`
3. In `shopify_app.rb` on the app, change `config.scope` or `config.user_access_scopes` to new set of scopes
4. Run `rails server` and load app
5. You will sent through to the OAuth flow before coming back to the app with updated scopes
6. If you change both `config.shop_access_scopes` and `config.user_access_scopes` it will go through OAuth twice for offline and online tokens

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ You can find documentation on gem usage, concepts, mixins, installation, and mor
   * [Generators](/docs/shopify_app/generators.md)
   * [ScriptTags](/docs/shopify_app/script-tags.md)
   * [Session repository](/docs/shopify_app/session-repository.md)
+  * [Handling changes in access scopes](/docs/shopify_app/handling-access-scopes-changes.md)
   * [Testing](/docs/shopify_app/testing.md)
   * [Webhooks](/docs/shopify_app/webhooks.md)
 
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -76,8 +76,18 @@ module ShopifyApp
       if jwt_request?
         false
       else
-        ShopifyApp::SessionRepository.user_storage.present? && user_session.blank?
+        return false unless ShopifyApp::SessionRepository.user_storage.present?
+        update_user_access_scopes?
       end
+    end
+
+    def update_user_access_scopes?
+      return true if user_session.blank?
+      user_access_scopes_strategy.update_access_scopes?(user_id: session[:user_id])
+    end
+
+    def user_access_scopes_strategy
+      ShopifyApp.configuration.user_access_scopes_strategy
     end
 
     def jwt_request?
--- a/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
+++ b/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
@@ -7,6 +7,9 @@ ShopifyApp.configure do |config|
   config.after_authenticate_job = false
   config.api_version = "<%= @api_version %>"
   config.shop_session_repository = 'Shop'
+
+  config.reauth_on_access_scope_changes = true
+
   config.allow_jwt_authentication = <%= !with_cookie_authentication? %>
   config.allow_cookie_authentication = <%= with_cookie_authentication? %>
 
--- a/lib/shopify_app/access_scopes/shop_strategy.rb
+++ b/lib/shopify_app/access_scopes/shop_strategy.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module AccessScopes
+    class ShopStrategy
+      class << self
+        def update_access_scopes?(shop_domain)
+          shop_access_scopes = shop_access_scopes(shop_domain)
+          configuration_access_scopes != shop_access_scopes
+        end
+
+        private
+
+        def shop_access_scopes(shop_domain)
+          ShopifyApp::SessionRepository.retrieve_shop_session_by_shopify_domain(shop_domain)&.access_scopes
+        end
+
+        def configuration_access_scopes
+          ShopifyAPI::ApiAccess.new(ShopifyApp.configuration.shop_access_scopes)
+        end
+      end
+    end
+  end
+end
--- a/lib/shopify_app/access_scopes/user_strategy.rb
+++ b/lib/shopify_app/access_scopes/user_strategy.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module AccessScopes
+    class UserStrategy
+      class InvalidInput < StandardError; end
+
+      class << self
+        def update_access_scopes?(user_id: nil, shopify_user_id: nil)
+          return update_access_scopes_for_user_id?(user_id) if user_id
+          return update_access_scopes_for_shopify_user_id?(shopify_user_id) if shopify_user_id
+          raise(InvalidInput, '#update_access_scopes? requires user_id or shopify_user_id parameter inputs')
+        end
+
+        private
+
+        def update_access_scopes_for_user_id?(user_id)
+          user_access_scopes = user_access_scopes_by_user_id(user_id)
+          configuration_access_scopes != user_access_scopes
+        end
+
+        def update_access_scopes_for_shopify_user_id?(shopify_user_id)
+          user_access_scopes = user_access_scopes_by_shopify_user_id(shopify_user_id)
+          configuration_access_scopes != user_access_scopes
+        end
+
+        def user_access_scopes_by_user_id(user_id)
+          ShopifyApp::SessionRepository.retrieve_user_session(user_id)&.access_scopes
+        end
+
+        def user_access_scopes_by_shopify_user_id(shopify_user_id)
+          ShopifyApp::SessionRepository.retrieve_user_session_by_shopify_user_id(shopify_user_id)&.access_scopes
+        end
+
+        def configuration_access_scopes
+          ShopifyAPI::ApiAccess.new(ShopifyApp.configuration.user_access_scopes)
+        end
+      end
+    end
+  end
+end
--- a/lib/shopify_app/configuration.rb
+++ b/lib/shopify_app/configuration.rb
@@ -18,6 +18,8 @@ module ShopifyApp
     attr_accessor :after_authenticate_job
     attr_accessor :api_version
 
+    attr_accessor :reauth_on_access_scope_changes
+
     # customise urls
     attr_accessor :root_url
     attr_writer :login_url
@@ -70,6 +72,16 @@ module ShopifyApp
 
     def shop_session_repository
       ShopifyApp::SessionRepository.shop_storage
+    end
+
+    def shop_access_scopes_strategy
+      return ShopifyApp::AccessScopes::NoopStrategy unless reauth_on_access_scope_changes
+      ShopifyApp::AccessScopes::ShopStrategy
+    end
+
+    def user_access_scopes_strategy
+      return ShopifyApp::AccessScopes::NoopStrategy unless reauth_on_access_scope_changes
+      ShopifyApp::AccessScopes::UserStrategy
     end
 
     def has_webhooks?
--- a/lib/shopify_app/omniauth/omniauth_configuration.rb
+++ b/lib/shopify_app/omniauth/omniauth_configuration.rb
@@ -50,7 +50,7 @@ module ShopifyApp
     end
 
     def update_shop_scopes?
-      false
+      ShopifyApp.configuration.shop_access_scopes_strategy.update_access_scopes?(shop_domain)
     end
 
     def shop_domain
--- a/lib/shopify_app/utils.rb
+++ b/lib/shopify_app/utils.rb
@@ -20,5 +20,17 @@ module ShopifyApp
     rescue ActiveResource::ConnectionError
       logger.error("[ShopifyAPI::ApiVersion] Unable to fetch api_versions from Shopify")
     end
+
+    def self.shop_login_url(shop:, return_to:)
+      return ShopifyApp.configuration.login_url unless shop
+      url = URI(ShopifyApp.configuration.login_url)
+
+      url.query = URI.encode_www_form(
+        shop: shop,
+        return_to: return_to,
+      )
+
+      url.to_s
+    end
   end
 end
--- a/test/controllers/callback_controller_test.rb
+++ b/test/controllers/callback_controller_test.rb
@@ -20,6 +20,8 @@ module ShopifyApp
       @routes = ShopifyApp::Engine.routes
       ShopifyApp.configuration = nil
       ShopifyApp.configuration.embedded_app = true
+      ShopifyApp.configuration.reauth_on_access_scope_changes = true
+      mock_user_scopes_match_strategy
 
       I18n.locale = :en
 
@@ -372,6 +374,16 @@ module ShopifyApp
 
       get :callback
       assert_response :ok
+    end
+
+    test "#callback redirects to login for user token flow if user session access scopes mismatch by user_id" do
+      mock_user_scopes_mismatch_strategy
+      mock_shopify_user_omniauth
+      _session = mock_user_session
+
+      get :callback
+
+      assert_redirected_to ShopifyApp.configuration.login_url
     end
 
     private
--- a/app/controllers/concerns/shopify_app/shop_access_scopes_verification.rb
+++ b/app/controllers/concerns/shopify_app/shop_access_scopes_verification.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module ShopAccessScopesVerification
+    extend ActiveSupport::Concern
+
+    included do
+      before_action :login_on_scope_changes
+    end
+
+    protected
+
+    def login_on_scope_changes
+      redirect_to(shop_login) if scopes_mismatch?
+    end
+
+    private
+
+    def scopes_mismatch?
+      ShopifyApp.configuration.shop_access_scopes_strategy.update_access_scopes?(current_shopify_domain)
+    end
+
+    def current_shopify_domain
+      return if params[:shop].blank?
+      ShopifyApp::Utils.sanitize_shop_domain(params[:shop])
+    end
+
+    def shop_login
+      ShopifyApp::Utils.shop_login_url(shop: params[:shop], return_to: request.fullpath)
+    end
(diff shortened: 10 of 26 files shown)


lib/shopify_app/omniauth/omniauth_configuration.rb
```diff
@@ -50,6 +50,8 @@ def default_request_online_tokens?
     end
 
     def update_shop_scopes?
+      ShopifyApp.configuration.shop_access_scopes_strategy.scopes_mismatch?(shop_domain)
+    rescue NotImplementedError
```
ragalie, line 54: Ideally we don't use errors for control flow. If we're not using a session store that supports access tokens can we just return `false` in `scopes_mismatch?`?

ragalie:
My high level feedback is that I don't think we need `scopes_mismatch?` to be open for extension. The extensibility doesn't seem worth the extra layer of indirection to me.

What's the case where we wouldn't want to redirect to get all the scopes if they're mismatched?

rezaansyed:
> My high level feedback is that I don't think we need `scopes_mismatch?` to be open for extension. The extensibility doesn't seem worth the extra layer of indirection to me.
> 
> What's the case where we wouldn't want to redirect to get all the scopes if they're mismatched?

@ragalie I think the better way of thinking about this is that the strategy should tell the caller whether scopes need to be updated. Currently, we're using it to determine whether a redirect happens (`ShopAccessScopesVerification` and `CallbackController`) but also to determine how the OmniAuth strategy gets built. So maybe `scope_mismatch?` isn't the right word, but rather, it should be `requires_scope_update?`?

Also, I think having the indirection of strategies is valuable for custom scenarios. For example, if developers only want to require scope updates when the merchant's scopes are not a subset of the configured scopes. Having it be a strategy allows to scopes to be handled at the appropriate points via a single interface.

test/controllers/callback_controller_test.rb
```diff
@@ -374,6 +375,18 @@ class CallbackControllerTest < ActionController::TestCase
       assert_response :ok
     end
 
+    test "#callback redirects to login for user token flow if user session access scopes mismatch by user_id" do
+      ShopifyApp.configuration.user_access_scopes_strategy = AccessScopesStrategyHelpers::MockUserScopesMismatchStrategy
+      mock_shopify_user_omniauth
+      # rubocop:disable Lint/UselessAssignment
+      session = mock_user_session
```
andyw8, line 382: If you write `_session` you won't need to disable this lint.

README.md
```diff
@@ -281,6 +281,40 @@ class Shop < ActiveRecord::Base
 end
 ```
 
+### Access Scopes Strategies
+The Shopify App gem now provides handling changes to scopes for both shop/offline and user/online tokens. By default, the `ShopifyApp::AccessScopes::ShopStrategy` and `ShopifyApp::AccessScopes::UserStrategy` strategies are used on app loads and during OAuth.
```
andyw8, line 285: Write 'now' here implies a recent change – which is true today, but will become less true over time, so I think should be dropped.

README.md
```diff
@@ -335,6 +367,8 @@ class AuthenticatedController < ApplicationController
   include ShopifyApp::Authenticated
 end
 ```
+### ShopAccessScopesVerification
+The `ShopifyApp::ShopAccessScopesVerification` concern helps merchant grant new access scopes requested by the app. The concern compares the current access scopes granted by the shop and compares it with the scopes requested by the app. If there is a mismatch in configuration, the merchant is redirected to login via OAuth and grant the net new scopes.
```
andyw8, line 371:
```suggestion
The `ShopifyApp::ShopAccessScopesVerification` concern helps merchants grant new access scopes requested by the app. The concern compares the current access scopes granted by the shop and compares it with the scopes requested by the app. If there is a mismatch in configuration, the merchant is redirected to login via OAuth and grant the net new scopes.
```

README.md
```diff
@@ -281,6 +281,40 @@ class Shop < ActiveRecord::Base
...
+    end
+  end
+end
+```
+
+#### Custom user access scopes strategy
+
+```ruby
+class CustomUserStrategy
+  class << self
```
NabeelAhsen, line 305: Could we move these guides out now to the appropriate docs under `/docs`?

lib/generators/shopify_app/install/templates/shopify_app.rb.tt
```diff
@@ -7,6 +7,9 @@ ShopifyApp.configure do |config|
   config.after_authenticate_job = false
   config.api_version = "<%= @api_version %>"
   config.shop_session_repository = 'Shop'
+
+  config.shop_access_scopes_strategy = ShopifyApp::AccessScopes::ShopStrategy
```
NabeelAhsen, line 11: Should we introduce the `config.shop_access_scopes` configuration in this PR?

I don't think we've documented the advanced layer anywhere yet, but this PR seems like a good place to introduce these configurations.

lib/shopify_app/access_scopes/user_strategy.rb
```diff
@@ -0,0 +1,35 @@
...
+          configuration_access_scopes != user_access_scopes
+        end
+
+        def update_access_scopes_for_shopify_user_id?(shopify_user_id)
+          user_access_scopes = user_access_tokens_by_shopify_user_id(shopify_user_id)
+          return true unless user_access_scopes
+          configuration_access_scopes != user_access_scopes
+        end
+
+        private
```
NabeelAhsen, line 19: I don't know how much this might complicate things, but it would be sweet to unify the `shop_strategy` and `user_strategy` under a single API: 

```ruby
module ShopifyApp
  module AccessScopes
    class Strategy
      def update_access_scopes?(*args)
      end
    end
  end
end 
```

This would make it easy for developers to template their own strategies. So for example, the `UserStrategy` could look something like this:

```suggestion
module ShopifyApp
  module AccessScopes
    class UserStrategy
      class << self
       def update_access_scopes?(user_id:, shopify_user_id:)
          return update_access_scopes_for_user_id(user_id) if user_id
          return update_access_scopes_for_shopify_user_id(shopify_user_id) if shopify_user_id
          raise CustomError
        end

        private
        
        def update_access_scopes_for_user_id?(user_id)
          user_access_scopes = user_access_tokens_by_user_id(user_id)
          return true unless user_access_scopes
          configuration_access_scopes != user_access_scopes
        end

        def update_access_scopes_for_shopify_user_id?(shopify_user_id)
          user_access_scopes = user_access_tokens_by_shopify_user_id(shopify_user_id)
          return true unless user_access_scopes
          configuration_access_scopes != user_access_scopes
        end
```

NabeelAhsen (commented):
Just leaving comments. I've yet to test these changes locally, but the new structure looks good!

lib/generators/shopify_app/install/templates/shopify_app.rb.tt
rezaansyed: We need to determine whether we want to make `shop_access_scopes_strategy` configurable right now. @tanema any thoughts on whether the strategies around access scopes should be configurable by partners? We can always make it non-configurable to begin with and then make the move to allow partners to extend it if it's desired.

tanema (approved):
Okay this looks great and solves a really big sticking point that people have had in the past. I think this PR will close a few issues.

NabeelAhsen (approved):
Tested locally, everything _seems_ to work for both offline and online access tokens.

_Re: configurability:_ The steps I took, besides adding `ShopifyApp::ShopAccessScopesVerification` to my home_controller, made it no way obvious that the strategies are configurable (this might be a good thing for now). 

I think that these strategies are fine the way the way they are, because they can't be extended without a doc that outlines a specific use-case.

lib/shopify_app/omniauth/omniauth_configuration.rb
rezaansyed: We could do that. Only reason we're not is because if the developers are trying to use the strategies to check scope matches (i.e. in ScopesVerification), they may assume that they have the appropriate scopes when access_scopes aren't actually stored. This basically silently fails for them and can possibly hide the underlying issue for them later on. In cases like OmniAuth and Callback controller, it's okay to fail silently because we want default behaviour.

rezaansyed: This won't be a problem when we later move onto a major version where the gem expects to have access_scopes by default.

lib/shopify_app/access_scopes/shop_strategy.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module AccessScopes
+    class ShopStrategy
+      class << self
+        def update_access_scopes?(shop_domain)
+          shop_access_scopes = shop_access_tokens(shop_domain)
+          return true unless shop_access_scopes
```
ragalie, line 9: Does this result in infinite loops if the app developer doesn't store access scopes? Is this backwards compatible?

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -76,10 +76,22 @@ def start_user_token_flow?
...
-        ShopifyApp::SessionRepository.user_storage.present? && user_session.blank?
+        return false unless ShopifyApp::SessionRepository.user_storage.present?
+        update_user_access_scopes?
       end
     end
 
+    def update_user_access_scopes?
+      return true if user_session.blank?
+      user_access_scopes_strategy.update_access_scopes?(user_id: session[:user_id])
+    rescue NotImplementedError
```
ragalie, line 87: What's the scenario when a `NotImplementedError` would be raised?

lib/shopify_app/configuration.rb
```diff
@@ -50,6 +53,9 @@ def initialize
       @webhooks_manager_queue_name = Rails.application.config.active_job.queue_name
       @disable_webpacker = ENV['SHOPIFY_APP_DISABLE_WEBPACKER'].present?
       @allow_cookie_authentication = true
+
+      @shop_access_scopes_strategy = ShopifyApp::AccessScopes::ShopStrategy
```
ragalie, line 57: I think you probably do need two strategies. You need a noop strategy if the developer isn't storing access scopes (so would always return `false` for `update_access_scopes?`) and the current strategy for developers who are storing them.

So you will need a configuration setting for `redirect_if_scopes_dont_match` or something (naming is hard).

lib/shopify_app/access_scopes/shop_strategy.rb
rezaansyed: No because if developer does not store access scopes, we would get a `NotImplemented` error. If this happens, the `ScopesVerification` fails (this is fine because it's opt-in). `CallbackController` and `OmniAuthConfiguration` will mark scope updates as not required and won't perform additional OAuth flows.

lib/shopify_app/omniauth/omniauth_configuration.rb
ragalie: See my comment here about a potentially more elegant way to handle this: https://github.com/Shopify/shopify_app/pull/1197/files#r586703216

lib/shopify_app/access_scopes/shop_strategy.rb
```diff
@@ -0,0 +1,25 @@
...
+          configuration_access_scopes != shop_access_tokens(shop_domain)
+        end
+
+        private
+
+        def shop_access_tokens(shop_domain)
```
greatestape, line 15: Why is this called `shop_access_tokens`?

lib/shopify_app/configuration.rb
rezaansyed: I like this idea and as discussed, it'll make sure that the strategies aren't partially turned on.

lib/shopify_app/access_scopes/shop_strategy.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module AccessScopes
+    class ShopStrategy
+      class << self
+        def update_access_scopes?(shop_domain)
+          shop_access_scopes = shop_access_tokens(shop_domain)
+          return true unless shop_access_scopes
+          configuration_access_scopes != shop_access_tokens(shop_domain)
```
greatestape, line 10: Why are you calling `shop_access_tokens` again?

lib/shopify_app/access_scopes/user_strategy.rb
```diff
@@ -0,0 +1,43 @@
...
+
+module ShopifyApp
+  module AccessScopes
+    class UserStrategy
+      class InvalidInput < StandardError; end
+
+      class << self
+        def update_access_scopes?(user_id: nil, shopify_user_id: nil)
+          return update_access_scopes_for_user_id?(user_id) if user_id
+          return update_access_scopes_for_shopify_user_id?(shopify_user_id) if shopify_user_id
```
greatestape, line 11: What if the user passes both, and the user doesn't need to update but the shop does?

lib/shopify_app/access_scopes/shop_strategy.rb
rezaansyed: Good catch! Fixed to `offline_access_tokens`

lib/shopify_app/access_scopes/shop_strategy.rb
rezaansyed: Fixed to `offline_access_tokens`

lib/shopify_app/access_scopes/user_strategy.rb
rezaansyed: If both are passed in, it would query it based on the `user_id`.

rezaansyed: This strategy does not concern itself with shop scope updates. That's `ShopStrategy` and if only a shop needs to update, then [OmniAuth](https://github.com/Shopify/shopify_app/pull/1197/files#diff-f173c9e33b743fa6bec8731bfcfe755501e57aea2a9e98a8e14a3ad9444f2004R53) will be able to determine that.

`ShopAccessScopesVerification` concern also does this check for shop scopes on app load.

lib/shopify_app/utils.rb
```diff
@@ -20,5 +20,17 @@ def self.fetch_known_api_versions
     rescue ActiveResource::ConnectionError
       logger.error("[ShopifyAPI::ApiVersion] Unable to fetch api_versions from Shopify")
     end
+
+    def self.shop_login_url(shop:, return_to:)
```
ragalie, line 24: Do we really not already have a helper for this? Seems unlikely to me that we need to write a new helper for this.

ragalie (approved):
LGTM!

lib/shopify_app/utils.rb
rezaansyed: Thought this was odd too. But since we're not relying on `LoginProtection` inside of the `HomeController`, we can't leverage those helper functions for constructing the login url.

lib/generators/shopify_app/install/templates/shopify_app.rb.tt
```diff
@@ -7,6 +7,9 @@ ShopifyApp.configure do |config|
   config.after_authenticate_job = false
   config.api_version = "<%= @api_version %>"
   config.shop_session_repository = 'Shop'
+
+  config.reauth_on_access_scope_changes = true
```
andyw8, line 11: 👋 @rezaansyed do you recall why we made this configurable, i.e. what's the use for _not_ enabling it? We're troubleshooting an issue with `https://github.com/Shopify/shopify-graphiql-app` which seems related to this being missing from the config.

(cc @teddyhwang @nelsonwittwer)

teddyhwang: Wouldn't the app break if the developers didn't automatically reauth? How did this work before this change, developers would have to do an OAuth in their app code themselves?

rezaansyed: Yup, that's correct

greatestape: Maybe what you want is to remove this from the template and default the value to on, but still support it as a config value so it can be explicitly disabled if an for some reason doesn't want this behaviour?

djkgamc: One bespoke case I noticed is that a logged in user may not have the appropriate permission to authorize the scope. 

Not sure if this is the best strategy or if it's already handled in the current implementation, but see here:
https://shashankkumar.medium.com/managing-access-scopes-for-shopify-apps-c93b3cf3541b

PR #1201, opened 2021-03-04
Enable CI build GitHub Action to run on pull requests on a fork

andyw8:
This wasn't running for external contributors, e.g. https://github.com/Shopify/shopify_app/pull/1187

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,10 @@
 name: CI
 
-on: 
+on:
   push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
 
 jobs:
   build:


mllemango (approved):

.github/workflows/build.yml
```diff
@@ -1,7 +1,10 @@
 name: CI
 
-on: 
+on:
   push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
Paulinakhew, line 7:
```suggestion
  [push, pull_request]:
    branches: 
      - master
```

Syntax taken from [github docs](https://docs.github.com/en/actions/reference/events-that-trigger-workflows). We can put the duplicates in a list and put `master` by itself

Paulinakhew (changes requested):

.github/workflows/build.yml
Paulinakhew: if we don't want to run CI twice, we can do this
```yaml
on:
  push:
    branches-ignore:
      - master
  pull_request:
    branches:
      - master
```

andyw8:
Let's not worry too much about running CI twice for the moment.

paulomarg (approved):
Agreed, let's run it twice for now and we can change it later.

PR #1205, opened 2021-03-05
Add a session tokens section to Troubleshooting guide

NabeelAhsen:
### What this PR does

This PR adds the following things to `/docs/Troubleshooting.md`:

* A table of contents
* A section on debugging common JWT session tokens issues

The best way to test this would be to view this file on the branch: https://github.com/Shopify/shopify_app/blob/update-troubleshooting-guide/docs/Troubleshooting.md

### Things to focus on

1. Are the links all working?
2. Is the information correct?

### Checklist

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -10,6 +10,7 @@
 
 [JWT session tokens](#jwt-session-tokens)
   * [My app is still using cookies to authenticate](#my-app-is-still-using-cookies-to-authenticate)
+  * [My app can't make requests to the Shopify API](#my-app-cant-make-requests-to-the-shopify-api)
 
 ## Generators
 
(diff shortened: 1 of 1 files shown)


NabeelAhsen:
The ruby build's finally passing ... no one touch anything! 😅 

(i'll try to capture an issue regarding intermittent test failures to investigate)

docs/Troubleshooting.md
```diff
@@ -1,7 +1,19 @@
-Troubleshooting Shopify App
-===========
+# Troubleshooting Shopify App
 
-### Generator shopify_app:install hangs
+#### Table of contents
```
paulomarg, line 3: Love this here, makes it a lot easier to know what's covered by this page.

docs/Troubleshooting.md
```diff
@@ -11,6 +23,52 @@ $ bundle exec spring stop
 
 Run shopify_app generator again.
 
-### App installation fails with 'The page you’re looking for could not be found' if the app was installed before
+## App installation
+
+## My app won't install
```
paulomarg, line 28: Nit: can we use ### for the subsection here? It looks like `App installation` is an empty section as it is.

docs/Troubleshooting.md
```diff
@@ -11,6 +23,52 @@ $ bundle exec spring stop
...
+#### `shopify_app` gem version
+
+Ensure the app is using shopify_app gem v13.x.x+. See [*Upgrading to `v13.0.0`*](/docs/Upgrading.md#upgrading-to-v1300).
+
+#### `shopify_app` gem Rails configuration
+
+Edit `config/initializer/shopify_app.rb` and ensure the following configurations are set:
+
+```diff
++ config.allow_jwt_authentication = true
```
paulomarg, line 47: Should we also reiterate about `embedded_app` and `allow_cookie_authentication` here?

paulomarg (approved):
Couple of nits, but LGTM otherwise!

docs/Troubleshooting.md
NabeelAhsen: Good catch!

docs/Troubleshooting.md
```diff
@@ -11,6 +23,55 @@ $ bundle exec spring stop
...
+```
+[ShopifyApp::JWT] Failed to validate JWT: [JWT::ImmatureSignature] Signature nbf has not been reached
+```
+
+**Note:** In a local development environment, you may want to temporarily update your `Gemfile` to point to a local instance of the `shopify_app` library instad of an installed gem. This will enable you to use a debugging tool like `byebug` to debug the library.
+
+```diff
+- gem 'shopify_app', '~> 14.2'
++ gem 'shopify_app', path: '/path/to/shopify_app'
+```
```
rezaansyed, line 77: Can we add this as well?
```
## How to handle invalid online access tokens for embedded apps

With session tokens, embedded apps using online tokens will need to login via OAuth from the frontend when after unsuccessful requests to the Shopify API.
```

NabeelAhsen: Let me know if the new wording + structure makes sense.

docs/Troubleshooting.md
```diff
@@ -11,6 +24,73 @@ $ bundle exec spring stop
...
+
+Converting your app to use session tokens means that your app will most likely not go through the OAuth flow as often as it did when relying on cookie sessions. Since the online access tokens stored in your app's database are refreshed during OAuth, this may cause your app's user session repository to use expired online access tokens.
+
+If the Shopify API  returns `401 Unauthorized`, handle this error on your app by redirecting the user to your login path to start the OAuth flow. As a result, your app will be given a new online access token for the current user.
+
+Add the following line to your app's unauthorized response handler:
+
+```diff
++ redirect_to(ShopifyApp.configuration.login_url, shop: current_shopify_domain)
```
rezaansyed, line 95: This will not work because the app cannot trigger redirects from the server for XHR requests. The redirect to login will need to be triggered from the frontend, which is why we pass in the `X-Shopify-API-Request-Failure-Unauthorized` header

rezaansyed (commented):
Looks good! Just need to modify the part about redirecting from the server as it should recommend a redirect from the frontend client.

docs/Troubleshooting.md
NabeelAhsen: Hmm, so this is the case where the frontend is using something like apollo-client that can't handle any status code like `302`, is that correct?

rezaansyed: yep!

rezaansyed (approved):

PR #1210, opened 2021-03-11
Fetch access scope via #dig in callback controller

kirillplatonov:
### Problem

After upgrading to v17.1.0 I noticed that controller tests are not passing. I use OmniAuth mock in controller tests to login as shop:
```ruby
class ActionDispatch::IntegrationTest
  def login(shop)
    OmniAuth.config.test_mode = true
    OmniAuth.config.add_mock(:shopify,
      provider: 'shopify',
      uid: shop.shopify_domain,
      credentials: { token: shop.shopify_token })
    Rails.application.env_config["omniauth.auth"] = OmniAuth.config.mock_auth[:shopify]

    get "/auth/shopify"
    follow_redirect!
  end
end
```

This code started raising an error:
CleanShot 2021-03-11 at 12 48 56@2x

### What this PR does

Changes the syntax for fetching access_scope in CallbackController to be tolerant to empty `auth_hash['extra']`.
`Hash#dig` is supported since Ruby 2.3 so it shouldn't break anything (since the library requires ruby >= 2.5).

### DIFF
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -129,8 +129,7 @@ module ShopifyApp
     end
 
     def access_scopes
-      return unless auth_hash['extra']['scope']
-      auth_hash['extra']['scope']
+      auth_hash.dig('extra', 'scope')
     end
 
     def reset_session_options


NabeelAhsen (approved):
Thank you for this! Can confirm this results in an error currently for Omniauth mocks


(Also can confirm that this change fixes the resulting errors)

rezaansyed (approved):

rezaansyed:
@kirillplatonov you'll need to sign the [CLA](https://cla.shopify.com/) in order to contribute to the gem.

kirillplatonov:
@rezaansyed Already signed CLA but the check not updated. Do I need to do anything else?
CleanShot 2021-03-12 at 00 19 01@2x

NabeelAhsen:
> @rezaansyed Already signed CLA but the check not updated. Do I need to do anything else?
> CleanShot 2021-03-12 at 00 19 01@2x

🤔  Hmm, try pushing up an empty commit to re-trigger the CLI?

kirillplatonov:
Done. CLA check is green now.

rezaansyed:
@kirillplatonov we pushed a fix so that the builds can run. Could you rebase your fork and push that up again?

kirillplatonov:
@rezaansyed Done. All green now.

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -129,7 +129,7 @@ def token
     end
 
     def access_scopes
-      return unless auth_hash['extra']['scope']
+      return unless auth_hash.dig('extra', 'scope')
```
andyw8, line 132: Couldn't both these lines be replaced by:

`auth_hash.dig('extra', 'scope')`

?

andyw8 (approved):

app/controllers/shopify_app/callback_controller.rb
kirillplatonov: Yeah, that's better. Updated.

rezaansyed:
@kirillplatonov thanks again for the fix! We've shipped this to the latest 17.1.1 version

Issue #1178, opened 2021-02-05
JSON response not rendered correctly with Ruby 3


andyw8:
See discussion in #1176

resistorsoftware:
I am curious about Rails for this. JS in the front-end expects some ActiveRecord data like Model -> User with an attribute first_name to be firstName, but does vanilla rails not just return first_name, assuming you get the JSON you want? 

ie) I was seeing my calls for data returning { "user": { "first_name": "AppleJuice" } } in my typical responses. What do people do about this? I use a cheesy gem called Olive Garden or something.

Issue #1163, opened 2021-01-26
Rails 6.1 support


kirillplatonov:
Since v17 `shopify_app` forces to use Rails < 6.1.
![CleanShot 2021-01-26 at 20 23 41@2x](https://user-images.githubusercontent.com/839922/105880516-7d232600-6014-11eb-9365-d1c43923b827.jpg)

It's stated in Readme that it's not compatible due to changes in `cookies_same_site_protection` setting in Rails 6.1. It's very easy to fix by restoring an old value of this setting (as a temporary workaround):
```ruby
# config/application.rb

config.load_defaults 6.1
config.action_dispatch.cookies_same_site_protection = nil
```

Can we allow installing `shopify_app` in Rails 6.1 and in Readme mention this workaround instead of forcing to use Rails 6.0? For now, the only solution is forking and regularly pulling the latest commits from upstream.

tanema:
There are more issues specifically around asset pipeline with 6.1 as well beyond just samesite. Our devs are looking into the best way to handle this.

oliwoodsuk:
I'm using Rails 6.1 and it's working fine (seemingly).
I really want to upgrade to the latest shopify_app version, as some customers are having issues with Safari and JWT. Think I'll hold off though until Shopify say 6.1 is all good.

awd:
What's the latest info on this? Still not supporting the latest Rails 6.1?

netwire88:
Hi @oliwoodsuk, @awd , we are having the same issues. Rails 6.1 supports disregarding arguments for webhooks, but we can't use it since shopify_app doesn't support Rails 6.1

oliwoodsuk:
> Hi @oliwoodsuk, @awd , we are having the same issues. Rails 6.1 supports disregarding arguments for webhooks, but we can't use it since shopify_app doesn't support Rails 6.1

Hi @netwire88 , we're using version 16.1 of the shopify_app gem with Rails 6.1 and it's working fine (well for all the things we use it for). I think this incompatibility issue is related to the more recent releases.

`gem 'shopify_app', '~> 16.1.0'`

itsniper:
Not ideal, but for now I've forked the project and [reverted](https://github.com/JDSoftwareCO/shopify_app/commit/54063698316bd6c310c41e212550f9b1d4617805) the Rails 6.1 limitation. As @kirillplatonov mentioned, I also added `config.action_dispatch.cookies_same_site_protection = nil` to `config/application.rb`.

[...]

PR #1221, opened 2021-03-18
Support rails 6.1

NabeelAhsen:
### What this PR does

> ~~Part 1 of 2: #1234~~

This PR introduces support for Rails 6.1. The key thing to support for Rails 6.1 is the following:

* Rails 6.1 configures `action_dispatch.cookies_same_site_protection = :lax` by default. This setting will not work with apps embedded in the Shopify Admin.

To fix this, embedded apps built with the next version of the Shopify App gem will have all their cookies set to `SameSite=None` using the `SameSiteCookieMiddleware`:

```diff
# same_site_cookie_middleware.rb
- cookie << '; SameSite=None' unless cookie =~ /;\s*samesite=/i
+ cookie << '; SameSite=None' if ShopifyApp.configuration.embedded_app?
```

For existing apps that wish to bump their Rails versions to `6.1`, but do not want to upgrade their shopify_app gem version, this PR provides a troubleshooting guide to configure the following in their `application.rb` config:

```diff
+ action_dispatch.cookies_same_site_protection = :none
```

To test this, I used this branch as the source of the shopify_app gem path using a new app built with Rails `v6.1.3`:

1. `rails _6.1.3_ new test-shopify-app`
2. `cd test-shopify-app`
3. Add `gem 'dotenv-rails'` and `gem 'shopify_app', path: 'local/path/to/shopify_app'` to the app's Gemfile
4. Create a `.env`
5. Generate a new shopify_app: `rails g shopify_app`
6. Register and run the app

<details><summary> ✅ Cookie behaviour using the current release of shopify_app `v8.3`</summary>

Screen Shot 2021-03-24 at 11 45 39 AM

_Rails 6.1 sets all cookies to `LAX` by default_
</details>

<details><summary> ✅ Cookie behaviour using this branch's `SameSiteCookieMiddleware`</summary>

Screen Shot 2021-03-24 at 11 46 24 AM

_Middleware sets all cookies to `None` by default if the app is embedded_
</details>

<details><summary> ✅ Cookie behaviour using Rails `application.rb` configurations</summary>
By adding the following:

```diff
# config/application.rb
require_relative "boot"
require "rails/all"

Bundler.require(*Rails.groups)

module TestShopifyApp
  class Application < Rails::Application
    config.load_defaults 6.1
+  config.action_dispatch.cookies_same_site_protection = :none
  end
end

```
Screen Shot 2021-03-24 at 11 46 24 AM

_Rails 6.1 sets all cookies to `None` by default_
</details>

### Things to focus on

1. I think this change is fair to ship as the next minor release (shopify_app `v17.2`). Is there any reason why this should be a major release?
2. Is there a way to mark future work for developers who are looking to support Rails 7+ to first handle changes made to the asset pipeline?
3. Should the release be part of a separate PR?

### Checklist


shopify_app.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.metadata['allowed_push_host'] = 'https://rubygems.org'
 
   s.add_runtime_dependency('browser_sniffer', '~> 1.2.2')
-  s.add_runtime_dependency('rails', '> 5.2.1', '< 6.1')
+  s.add_runtime_dependency('rails', '> 5.2.1', '~> 6.1')
```
NabeelAhsen, line 17: Is this followed by a `bundle update` the right way to do it?

andyw8: Not answering the `bundler update` question, but is that range valid? I think they are `AND`ed together. So this would prevent Rails 5 from working?

NabeelAhsen: Thanks for clarifying! I was definitely confused whether this was an `AND` or a range

paulomarg:
If I remember correctly, there were also issues with our asset pipeline (specifically with webpack, if I'm not mistaken). Have those issues been solved?

NabeelAhsen:
> If I remember correctly, there were also issues with our asset pipeline (specifically with webpack, if I'm not mistaken). Have those issues been solved?

Is there an issue/discussion for more context around this? I can see if this is still outstanding

andyw8:
https://github.com/Shopify/shopify_app/issues/1163

andyw8:
^ Ah, that issue is not so informative. I think there's more detail elsewhere...

Gemfile.lock
```diff
@@ -5,98 +5,103 @@ PATH
       browser_sniffer (~> 1.2.2)
       jwt (~> 2.2.1)
       omniauth-shopify-oauth2 (~> 2.2.2)
-      rails (> 5.2.1, < 6.1)
+      rails (> 5.2.1, < 6.2)
```
paulomarg, line 8: I wonder if we can just change this requirement to `~> 6.1` now that we know it'll be safe until v7.

shopify_app.gemspec
NabeelAhsen: re:

> I wonder if we can just change this requirement to ~> 6.1 now that we know it'll be safe until v7.

cc @paulomarg

I don't believe this'll be a valid range, so maybe `'< 7.0'`?

paulomarg: Oh, this was hidden in the current files, sorry! It seems a little strange that we previously supported `> 5.2.1` rather than `~>5.0` to begin with, but I guess we don't have to drop support for v5 if it still works. So yeah, let's go with the upper cap just to make sure we're not flipping over to breaking changes.

Gemfile.lock
```diff
@@ -209,8 +209,8 @@ GEM
       unicode-display_width (>= 1.4.0, < 3.0)
     rubocop-ast (1.4.1)
       parser (>= 2.7.1.5)
-    rubocop-shopify (1.0.7)
-      rubocop (~> 1.4)
+    rubocop-shopify (2.0.1)
+      rubocop (~> 1.11)
```
NabeelAhsen, line 213: More context: https://github.com/Shopify/ruby-style-guide/pull/238

This change enforces that all files use double quotes. The PR above suggests that this is the preference for Rails 6.1 as well. 

Making this change in this PR would mean **a lot** of files are changed, and not doing it would mean our rubocop build fails. How does everyone feel about making this change in this PR?

NabeelAhsen: For reference, running `bundle exec rubocop -a` makes 1200+ corrections.

NabeelAhsen: Alternatively, is it alright to just not jump to the latest `rubocop-shopify` release?

paulomarg: I would keep this one as is and set up a separate PR containing only the auto-updates.

Gemfile.lock
rezaansyed: Yeah same here, I think we can probably just keep the rails requirement to be: `rails (> 5.2.1, < 7)`

docs/Troubleshooting.md
```diff
@@ -24,6 +28,33 @@ $ bundle exec spring stop
...
+  class Application < Rails::Application
++    config.load_defaults 6.1
+
++    config.action_dispatch.cookies_same_site_protection = :none
+    ...
+  end
+end
+```
+
+As of Rails `v6.1`, the same-site cookie protection setting defaults  to `Lax`. This does not allow an embedded app to make cross-domain requests in the Shopify Admin.
```
paulomarg, line 56: Will the solution from `Upgrading` fix this? If so, we could also suggest that as a fix.

NabeelAhsen: That's a good point. I've added a line just below this that suggests an alternative to this is to upgrade to `v17.2.0` of the shopify_app gem. Unfortunately there's limited developer support for configuring the `SameSiteCookieMiddleware` at this time.

paulomarg (approved):

paulomarg:
I tested running the code for this PR with Ruby v3, and it works. We may want to tag that this closes https://github.com/Shopify/shopify_app/issues/1178 as well!

NabeelAhsen:
@andyw8 @paulomarg  it looks like there are a few Shopify teams waiting for this change. I'm confident to merge it, but I'd like to get your final opinions on the gemspec constraint I've added.

Is it better to restrict support to Rails version `<6.2` or `<7`?

rezaansyed (approved):
LGTM

spickermann:
The README.md still tells that _Rails 6.1 or above is not yet supported_.

Issue #1218, opened 2021-03-15
[tests] Intermittently failing tests: undefined method `shopify_app_path'


andyw8:
I've seen a few builds with this failure, e.g.:

https://github.com/Shopify/shopify_app/pull/1204/checks?check_run_id=2041576868

```
Error:
ShopifyApp::SessionsControllerTest#test_#new_redirects_to_the_enable_cookies_page_if_we_can't_set_cookies_and_the_user_agent_supports_cookie_partitioning:
NoMethodError: undefined method `shopify_app_path' for #<Module:0x00007fabee568398>
    /Users/runner/work/shopify_app/shopify_app/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.3.5/lib/action_dispatch/routing/mapper.rb:687:in `block in define_generate_prefix'
    /Users/runner/work/shopify_app/shopify_app/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.3.5/lib/action_dispatch/routing/mapper.rb:699:in `block (2 levels) in define_generate_prefix'
    /Users/runner/work/shopify_app/shopify_app/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.3.5/lib/action_dispatch/routing/route_set.rb:807:in `url_for'
    /Users/runner/work/shopify_app/shopify_app/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.3.5/lib/action_dispatch/routing/route_set.rb:269:in `call'
    /Users/runner/work/shopify_app/shopify_app/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.3.5/lib/action_dispatch/routing/route_set.rb:211:in `call'
    /Users/runner/work/shopify_app/shopify_app/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.3.5/lib/action_dispatch/routing/route_set.rb:326:in `block in define_url_helper'
    /Users/runner/work/shopify_app/shopify_app/app/controllers/shopify_app/sessions_controller.rb:146:in `enable_cookie_access'
    /Users/runner/work/shopify_app/shopify_app/app/controllers/shopify_app/sessions_controller.rb:91:in `authenticate_with_partitioning'
    /Users/runner/work/shopify_app/shopify_app/app/controllers/shopify_app/sessions_controller.rb:69:in `authenticate'
[...]

PR #1225, opened 2021-03-22
Fix intermittent engine controller test failures

NabeelAhsen:
### What this PR does

> Closes https://github.com/Shopify/shopify_app/issues/1218

**TL;DR** We've been seeing the following failures intermittently on builds for a few weeks now:

1. `NoMethodError: undefined method 'shopify_app_path' for #<Module:xyz>`
2. `NoMethodError: undefined method 'root_path' for #<Module:xyz>`

On closer investigation, the two Engine controllers that keep failing are `sessions_controllers` and `webhooks_controller`. 

**One fix that appears to work is to reload the Rails routes between each test.**

---

I'm not 100% sure what the order of operations are for engines when they mount, but I believe that these two controllers have difficulty finding the following routes mounted on the dummy app between tests:

https://github.com/Shopify/shopify_app/blob/129cea927861a21be2f5d22f6872fc1ff3658667/test/dummy/config/routes.rb#L1-L5

The routes above correspond to the dummy app paths `shopify_app_path` and `root_path` respectively. In Rails 6.x.y, the primary point of failure originates from [`ActionDispatch::Routing::Mapper`](https://www.rubydoc.info/docs/rails/4.1.7/ActionDispatch/Routing/Mapper).

The issue can be reproduced with the following command:

```sh
bundle exec rake test TESTOPTS="--seed=57116
```

To test this change, do the following:

1. Verify that you can run the command above and get intermittent failures on the `master` branch
2. Verify that you can run the command above and get consistently passing tests on this branch: `bug-fix/fix-intermittent-test-failures`

### Things to focus on

1. Is someone able to pinpoint what exactly it is about `Rails::Engine` that is non-deterministic? I don't think I fully understand why there was sudden inconsistent behaviour when testing engine-namespaced controllers (like `SessionsController` and `WebhooksController`)
2. Are there any other controllers defined by `ShopifyApp` engine that fail intermittently?

### Checklist

### DIFF
--- a/test/controllers/sessions_controller_test.rb
+++ b/test/controllers/sessions_controller_test.rb
@@ -10,6 +10,7 @@ end
 module ShopifyApp
   class SessionsControllerTest < ActionController::TestCase
     setup do
+      Rails.application.reload_routes!
       @routes = ShopifyApp::Engine.routes
       ShopifyApp::SessionRepository.shop_storage = ShopifyApp::InMemoryShopSessionStore
 
--- a/test/integration/webhooks_controller_test.rb
+++ b/test/integration/webhooks_controller_test.rb
@@ -10,6 +10,7 @@ module ShopifyApp
     include ActiveJob::TestHelper
 
     setup do
+      Rails.application.reload_routes!
       WebhooksController.any_instance.stubs(:verify_request).returns(true)
       WebhooksController.any_instance.stubs(:webhook_namespace).returns(nil)
     end


andyw8:
> Is someone able to pinpoint what exactly it is about Rails::Engine that is non-deterministic? I don't think I fully understand why there was sudden inconsistent behaviour when testing engine-namespaced controllers (like SessionsController and WebhooksController)
> Are there any other controllers defined by ShopifyApp engine that fail intermittently?

Even if we can't answer these yet, I think it's still worth shipping this.

andyw8 (approved):

PR #1227, opened 2021-03-23
Reload Rails routes before each engine test

NabeelAhsen:
### What this PR does

> Fixes (hopefully) https://github.com/Shopify/shopify_app/issues/1218

> This is a follow up to #1225 

There are still intermittently failing tests, but this time on different files than before. See the recent failing build: https://github.com/Shopify/shopify_app/pull/1221/checks?check_run_id=2175638755

The one thing all of these test files have in common is that they invoke `@routes = ShopifyApp::Engine.route` in their setup blocks, similar to the test files in #1225 

This PR adds logic to reload the Rails routes on all such test files.

Test `bundle exec rake test` on ruby 2.7, 2.6, and 2.5 repeatedly. I've noticed that the files changed in this PR are the ones consistently failing across each ruby version, except with different files failing in 2.6 vs 2.7. 

### Things to focus on

1. I have not yet analyzed how much time reloading routes adds to each test. I can't notice a significant difference on my local computer, but this may not be the case in a CI environment.

### Checklist

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -44,5 +44,6 @@ class ActiveSupport::TestCase
     WebMock.disable_net_connect!
     WebMock.stub_request(:get, "https://app.shopify.com/services/apis.json").to_return(body: API_META_TEST_RESPONSE)
     ShopifyAppConfigurer.call
+    Rails.application.reload_routes!
   end
 end
--- a/test/controllers/concerns/ensure_authenticated_links_test.rb
+++ b/test/controllers/concerns/ensure_authenticated_links_test.rb
@@ -36,10 +36,6 @@ class EnsureAuthenticatedLinksTest < ActionController::TestCase
     end
   end
 
-  teardown do
-    Rails.application.reload_routes!
-  end
-
   test 'redirects to splash page with a return_to and shop param if no session token is present' do
     get :some_link, params: { shop: @shop_domain }
 
--- a/test/controllers/concerns/require_known_shop_test.rb
+++ b/test/controllers/concerns/require_known_shop_test.rb
@@ -17,10 +17,6 @@ class RequireKnownShopTest < ActionController::TestCase
     end
   end
 
-  teardown do
-    Rails.application.reload_routes!
-  end
-
   test 'redirects to login if no shop param is present' do
     get :index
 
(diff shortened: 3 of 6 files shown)


rezaansyed (approved):

paulomarg (commented):
Would it make this more future-proof if we created a base `TestCase` class that all other ones inherit from, which sets up the routing?

NabeelAhsen:
> Would it make this more future-proof if we created a base `TestCase` class that all other ones inherit from, which sets up the routing?

I considered this until I realized that some engine tests inherit from `ActionController::TestCase` and some from `ActionDispatch::IntegrationTest`. I know that the former is deprecated in favour of the latter according to the first paragraph here: https://api.rubyonrails.org/v6.1.0/classes/ActionController/TestCase.html

Problem is I don't know how to reconcile the two - the functional-style `ActionController::TestCase` gives us the power to manipulate individual requests on a controller (including session objects). The integration-style only lets us observe request objects after the request has been made.

test/controllers/callback_controller_test.rb
```diff
@@ -17,6 +17,7 @@ class CallbackControllerTest < ActionController::TestCase
     TEST_SESSION = "this.is.a.user.session"
 
     setup do
+      Rails.application.reload_routes!
```
andyw8, line 20: I wonder if we should do this globally for every test, i.e. put it in `test_helper`, to reduce the chance of a similar regression being introduced later.

paulomarg: I like doing this in `test_helper` better than a base class, since our tests are extending multiple classes.

NabeelAhsen: I'll look into making this change!

test/test_helper.rb
```diff
@@ -44,5 +44,6 @@ def before_setup
     WebMock.disable_net_connect!
     WebMock.stub_request(:get, "https://app.shopify.com/services/apis.json").to_return(body: API_META_TEST_RESPONSE)
     ShopifyAppConfigurer.call
+    Rails.application.reload_routes!
```
NabeelAhsen, line 47: I've opted to call this in `ActiveSupport::TestCase#before_setup` to run before each test's setup block. 

I haven't noticed a considerable change in test runtime.

NabeelAhsen: This allows us to remove all other instances of `Rails.application.reload_routes!` across our tests

andyw8 (approved):

paulomarg (approved):
Makes sense to me!

PR #1234, opened 2021-03-29
Update rubocop to latest Shopify cops

NabeelAhsen:
### What this PR does

> ~~Part 2 of 2: #1221~~ 

This PR does the following things:

1. Run `bundle update`
    * This updates the `rubocop` dependencies to the latest
 2. Run `bundle exec rubocop -a`
     * This auto-fixes all files to conform to the latest Shopify cops
 3. Ensure `bundle exec rake test` passes
     * This is done by escaping all mocked out strings and expected assert values to reflect the new double-quoted strings requirement cop
     
Each step above is marked as its own commit

### Things to focus on

1. This is unlikely, but in the third commit, does escaping any of the expected string values change the core behaviour of this gem?

### Checklist


andyw8:
Is this tied into the Rails 6.1 update in some way? Not sure I understand the connection.

NabeelAhsen:
> Is this tied into the Rails 6.1 update in some way? Not sure I understand the connection.

Hmm on second thought this change isn't related to updating our gemspec to support Rails 6.1. 

For some context, when I ran `bundle update`, `rubocop` (and `rubocop-shopify`) automatically bumped their versions to the latest. The latest version of `rubocop-shopify` introduces a new cop to enforce double quotes on all strings.

1. Do you think this change is necessary?
2. Should this be a part of the bigger gemspec update to support Rails 6.1?

andyw8:
If you run `bundle update rails --conservative` then it should limit what gets updated. 

I think we _should_ update to follow the styleguide, but do it separately from the Rails 6.1 update.

NabeelAhsen:
> If you run `bundle update rails --conservative` then it should limit what gets updated.

TIL!

I'm going to leave this PR as a draft for now, and rename it to no longer be associated with the change to Rails 6.1. This change can be made at a later date!

rezaansyed (approved):

PR #1247, opened 2021-04-07
introduce slim app-bridge client for redirect

hannachen:
### What this PR does

This PR replaces EASDK redirect with an App Bridge redirect when attempting to break out of an iframe. This happens when an app is installed on a shop, but somehow the shop's record is lost on the app's side, and requires re-authentication.

Perks for this approach:
- Because this is an app-bridge action, it will be tracked
- With a special build of app-bridge, we can use this special package to identify apps using the `shopify_app` gem

1. Clone this branch
1. In a terminal, use [Shopify CLI](https://shopify.dev/tools/cli) to create and configure a new rails app: `shopify create rails`
1. In an editor, open the `Gemfile` inside the newly created app
1. Modify the entry to `shopify_app` to point to this branch in your local environment so that it looks similar to this:
    `gem 'shopify_app', path: '~/src/github.com/Shopify/shopify_app'`
1. In a terminal, start your app server with `shopify serve`, this will also output a ngrok url
1. In a browser, follow the ngrok url to install the app on a test store. It should look similar to this:
    `https://1587ef2f72d5.ngrok.io/login?shop=sabotender-shop.myshopify.com`
1. On your local environment, remove the `shop` entry for the store inside the db
1. In a browser, open the installed app. i.e. `https://sabotender-shop.myshopify.com/admin/apps/test-app-1`
1. Ensure that the entire redirect flow takes the browser back to the app. 

### Things to focus on

1. Focus on install/uninstall
2. Other areas that might kick off a top level redirect, or use the methods `fullpage_redirect_to`, `authenticate_at_top_level`, or `enable_cookie_access`

### Checklist


app/assets/javascripts/shopify_app/redirect.js
```diff
@@ -1,8 +1,11 @@
+//= require ./app_bridge_redirect_1.30.0.js
+
 (function() {
   function redirect() {
     var redirectTargetElement = document.getElementById("redirection-target");
+    var bodyElement = document.getElementsByTagName('body')[0];
```
vividviolet, line 6: I think this can be simplified to `document.body`

app/assets/javascripts/shopify_app/redirect.js
```diff
@@ -12,15 +15,23 @@
...
-        message: 'Shopify.API.remoteRedirect',
-        data: {location: normalizedLink.href}
+      var AppBridge = window['app-bridge'];
+      var actions = window['app-bridge'].actions;
+      var apiKey = bodyElement.dataset.apiKey;
+      var urlParams = new URLSearchParams(window.location.search);
+
+      var app = AppBridge.createApp({
+        apiKey: apiKey,
+        shopOrigin: urlParams.get('shop'),
```
vividviolet, line 30: The previous related PR grabbed the shop from the server which might be more reliable?

https://github.com/Shopify/shopify_app/pull/915/files#diff-d9b496aa95050a082c15327f130b749024b3afd9ea1a8c57b0ad915db38f439cR10

app/assets/javascripts/shopify_app/redirect.js
```diff
@@ -12,15 +15,23 @@
       // If the current window is the 'parent', change the URL by setting location.href
       window.top.location.href = targetInfo.url;
     } else {
-      // If the current window is the 'child', change the parent's URL with postMessage
+      // If the current window is the 'child' or embedded, change the parent's URL with App Bridge redirect
+      // This case happens when the app is installed but the app does not know that because there's no shop entry
```
greatestape, line 19: Not sure about this. I think the main reason we need to redirect out of the iframe is when the app wants to go through oauth to get new scopes.

app/assets/javascripts/shopify_app/redirect.js
```diff
@@ -12,15 +15,23 @@
...
 
-      data = JSON.stringify({
-        message: 'Shopify.API.remoteRedirect',
-        data: {location: normalizedLink.href}
+      var AppBridge = window['app-bridge'];
+      var actions = window['app-bridge'].actions;
+      var apiKey = bodyElement.dataset.apiKey;
+      var urlParams = new URLSearchParams(window.location.search);
+
+      var app = AppBridge.createApp({
```
greatestape, line 28: Should we also pass `forceRedirect: false` here?

app/assets/javascripts/shopify_app/redirect.js
hannachen: The new access scopes seem to only kick off re-auth when a backend call fails, but I will test this out some more. I usually delete the shop entry in the app's db to trigger this flow.

app/assets/javascripts/shopify_app/redirect.js
hannachen: Good call!

NabeelAhsen:
> Could end up with multiple versions of App Bridge

I'm unsure about the maintenance cost of this. How often will this need to be updated?

hannachen:
> > Could end up with multiple versions of App Bridge
> 
> I'm unsure about the maintenance cost of this. How often will this need to be updated?

It depends on the actions used here. Since we're only using a minimal number of actions, there shouldn't be a need to update often. We could potentially automate this process by pulling in the file from the CDN every time we update the library. The key thing is to download it and get the app itself to host the client. I think it would be nice to make it clear which version of the app bridge client is shipped different versions of the `shopify_app` gem.

lib/shopify_app/engine.rb
```diff
@@ -16,6 +16,7 @@ class Engine < Rails::Engine
 
     initializer "shopify_app.assets.precompile" do |app|
       app.config.assets.precompile += %w[
+        shopify_app/app_bridge_redirect_1.30.0.js
```
vividviolet, line 19: Is this needed when it's already included as part of `shopify_app/redirect.js` and the other files?

package-lock.json
```diff
@@ -0,0 +1,6961 @@
+{
```
vividviolet, line 1: Why do we have so many changes to this file?

vividviolet (commented):
Looks great - should we include instructions somewhere on how to replace the  `app/assets/javascripts/shopify_app/app_bridge_1.30.0.js` file with another if needed?

package-lock.json
hannachen: Good catch, this was added by mistake, deleted!

vividviolet (approved):
Code changes look good but I didn't have time to tophat

app/assets/javascripts/shopify_app/app_bridge_redirect.js
```diff
@@ -0,0 +1,23 @@
...
+    var app = createApp({
+      apiKey: shopifyData.apiKey,
+      shopOrigin: shopifyData.shopOrigin,
+      forceRedirect: false,
+    });
+
+    var normalizedLink = document.createElement('a');
+    normalizedLink.href = url;
+
+    Redirect.create(app).dispatch(Redirect.Action.REMOTE, normalizedLink.href);
```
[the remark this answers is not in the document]
hannachen: If I remember correctly, this was a trick to normalize the URL and make it relative to the current path. Not sure if it still applies tho, I will test it out!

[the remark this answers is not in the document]
joeyfreund: 👍 Thanks for the clarification.
Unless it's a super well-known standard trick, I would consider making this explicit. For example:

```js
if (!url.startsWith('http://') && !url.startsWith('https://')){
  var base = window.location.protocol + "//" + window.location.host;
  url = new URL(url, base).toString()
}
```

Or, keep using the trick, but extract it to a help function with a descriptive name (for example, `prependCurrentHostIfRelative`)

app/assets/javascripts/shopify_app/redirect.js
```diff
@@ -12,15 +14,10 @@
...
-
-      data = JSON.stringify({
-        message: 'Shopify.API.remoteRedirect',
-        data: {location: normalizedLink.href}
-      });
-      window.parent.postMessage(data, targetInfo.myshopifyUrl);
+      // If the current window is the 'child' or embedded, change the parent's URL with App Bridge redirect
+      // This case can happen when an app updates its access scopes, or when the shop thinks the app is installed, but
+      // the app does not have shop entry
+      window.appBridgeRedirect(targetInfo.url);
```
joeyfreund, line 20: Now that I understand [the "link trick"](https://github.com/Shopify/shopify_app/pull/1247/files#r789210446), I'm wondering if this line has a bug 🤔  ...

Can `targetInfo.url` be relative?
If so, the old code would have prepended the current host to it, and the new code doesn't.

PR #1255, opened 2021-04-21
Add Configuration for delayed webhook processing

### LINKED ISSUE #1194: Improved Configuration Request - configurable wait for background jobs
ActiveJob allows the consumer to apply some configuration before the job is pushed into the queue, one of those [being a `wait` value to process the job in at a later time](https://edgeguides.rubyonrails.org/active_job_basics.html#enqueue-the-job).

I'm running into some issues where my application and Shopify Flow are modifying the same resource at roughly the same time and losing data. If I could defer my jobs to run a little bit later when they first arrive this would make things a bit easier on my end.

It looks like all that would need to be done is making a small change to the [webhooks_controller](https://github.com/Shopify/shopify_app/blob/master/app/controllers/shopify_app/webhooks_controller.rb#L11) and adding a configuration variable.

Would there be interest in a PR of this feature?

jiblits:
Closes #1194

### What this PR does

This adds a configuration variable that allows developers to defer the processing of a webhook by a specified amount of time. This can be useful for developers who experience a lot of flux on items they are getting webhooks for.

Running `bundle exec rake test` should be all that is needed. Though a simple test app that receives webhooks from Shopify will also verify that this behaves as expected.

### Things to focus on

1. By using `ActiveJob#set` I needed to pass in a stub, which I'm not a huge fan of. A suggestion on how to stub this out better (or avoid the stubbing) would be appreciated.
2. Is the name of the configuration variable acceptable?
3. This is only the code portion, however if you want updates to the README about the new configuration I can also do that.

### Checklist


jiblits:
The CLA for Escarpment Labs was submitted last Friday. I don't know what is required in order to make that test run again, but we should be good on that front.

PR #1258, opened 2021-04-28
Upgrade to omniauth 2.0

greatestape:
### What this PR does

* Makes shopify_app compatible with omniauth 2.x

* Create a new app or install or
* Install the local checkout of the gem and `bundle install`

### Things to focus on

* [x] Apps created from scratch have omniauth 2.x as a dependency out of the box
* [x] Existing app can update to the new version of `shopify_app` and then update to omniauth 2.x
* [x] Oauth works
  * Actions
    * [x] App install
    * [x] Requesting new scopes
  * Scenarios
    * [x] In Chrome
    * [x] In Safari
    * [x] In an app using Session Tokens
    * [x] In an app using 3P session cookies
    * [x] In an app using only offline tokens
    * [x] In an app using offline and online tokens

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 Unreleased
 ----------
-* Support omniauth 2.x
+* Support OmniAuth 2.x 
+  * If your app has custom OmniAuth configuration, please refer to the [OmniAuth 2.0 upgrade guide](https://github.com/omniauth/omniauth/wiki/Upgrading-to-2.0).
 
 17.2.1 (April 1, 2021)
 ----------
--- a/app/views/shopify_app/shared/post_redirect_to_auth_shopify.html.erb
+++ b/app/views/shopify_app/shared/post_redirect_to_auth_shopify.html.erb
@@ -5,7 +5,15 @@
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   <base target="_top">
   <title>Redirecting…</title>
-  <%= javascript_include_tag('shopify_app/post_redirect', crossorigin: 'anonymous', integrity: true) %>
+  <script>
+    function redirect() {
+      var form = document.getElementById("redirect-form");
+      if (form) {
+        form.submit();
+      }
+    }
+    document.addEventListener("DOMContentLoaded", redirect);
+  </script>
 </head>
 <body>
   <%= form_tag '/auth/shopify', id: 'redirect-form' %>
--- a/test/shopify_app/session/jwt_test.rb
+++ b/test/shopify_app/session/jwt_test.rb
@@ -50,7 +50,7 @@ module ShopifyApp
 
     test "#shopify_domain and #shopify_user_id are nil if the jwt is unsigned" do
       # The library expects there to be a 3rd segment if we want to verify signature
-      expect_jwt_error(::JWT::DecodeError, 'Not enough or too many segments')
+      expect_jwt_error(::JWT::IncorrectAlgorithm, 'Expected a different algorithm')
 
       t = ::JWT.encode(payload, nil, 'none')
       jwt = JWT.new(t)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     shopify_app (17.2.1)
       browser_sniffer (~> 1.2.2)
-      jwt (~> 2.2.1)
+      jwt (>= 2.2.3)
       omniauth-rails_csrf_protection
       omniauth-shopify-oauth2 (~> 2.3)
       rails (> 5.2.1, < 6.2)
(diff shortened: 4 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 Unreleased
 ----------
+* Support omniauth 2.x
```
andyw8, line 3: I think we'll need some instructions here for people who are upgrading?

test/shopify_app/session/jwt_test.rb
```diff
@@ -50,7 +50,7 @@ class JWTTest < ActiveSupport::TestCase
 
     test "#shopify_domain and #shopify_user_id are nil if the jwt is unsigned" do
       # The library expects there to be a 3rd segment if we want to verify signature
-      expect_jwt_error(::JWT::DecodeError, 'Not enough or too many segments')
+      expect_jwt_error(::JWT::IncorrectAlgorithm, 'Expected a different algorithm')
```
greatestape, line 53: I wish the upgrade of `jwt` and this test change could be isolated in their own commit. If I ran across this change later and found that it was part of a commit upgrading omniauth I would be stumped. I realize it might be hard to tease apart the jwt upgrade from the rest of the upgrades, but I just wanted to put my hopes and dreams out there in the world.

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -223,7 +223,6 @@ def sanitized_params
     end
 
     def return_address
-      return base_return_address unless ShopifyApp.configuration.allow_jwt_authentication
```
ragalie, line 226: What's the rationale behind this change?

test/shopify_app/session/jwt_test.rb
rezaansyed: Yeah I'm trying to see if we can avoid it but because we're allowing patch releases of `jwt` and the bump from `2.2.2` to `2.2.3` included this change, we may have to support it.

app/views/shopify_app/shared/post_redirect_to_auth_shopify.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <title>Redirecting…</title>
+  <script>
+    function redirect() {
+      var form = document.getElementById("redirect-form");
+      if (form) {
+        form.submit();
+      }
+    }
+    window.addEventListener("load", redirect);
+  </script>
```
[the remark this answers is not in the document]
rezaansyed: I think listening for `DOMContentLoaded` event makes sense. With regards to moving this to the bottom of `<body>`, not sure if that approach is required. @hannachen do you have any thoughts on this?

lib/shopify_app/controller_concerns/login_protection.rb
rezaansyed: the return address previously didn't have the `shop` and `host` params when redirecting to the app after OAuth. Because of that, we're not able to use those params to load App Bridge to embed the app inside of an iframe. We can simply pass in the params in the return address now.

ragalie: Ok, can you help me understand how this is related to the omniauth upgrade?

rezaansyed: This isn't actually. I've gone ahead and created a [new PR](https://github.com/Shopify/shopify_app/pull/1259) for this. Could I get a review on that?

test/shopify_app/session/jwt_test.rb
ragalie: The gemspec still allows `2.2.2`, which I don't think will work properly after this change? https://github.com/Shopify/shopify_app/pull/1258/files#diff-d72fff1e948b092ced84db03fa37f7c2aa1774033d3aa250d41d847efb2a2264R21

You'll need to require `>= 2.2.3` for this change to be safe, I think.

app/views/shopify_app/shared/post_redirect_to_auth_shopify.html.erb
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <base target="_top">
+  <title>Redirecting…</title>
+  <script>
+    function redirect() {
```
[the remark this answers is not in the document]
rezaansyed: Done.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
...
```
andyw8, line 3:
```suggestion
* Support OmniAuth 2.x 
```

alexaitken (approved):
LGTM

Tested the following flows and configuration combinations
- Setup
  - New app
  - Upgrade existing app
- Configurations   
  - In an app using Session Tokens
  - In an app using 3P session cookies
  - In an app using only offline tokens
- Actions
     - App install
     - Requesting new scopes
- Browsers
   -  In latest Chrome
   -  In latest Safari
   -  In latest Firefox

PR #1259, opened 2021-04-29
Bugfix: Pass host param into return address regardless of apps's auth config

rezaansyed:
### What this PR does

This PR fixes a bug relating to recent changes to support App Bridge 2.0. For apps that rely on 3p cookies, the `host` parameter is not being passed in, which leads to failure to load App Bridge. This PR ensures that the host parameter is passed after completing the OAuth flow.

- [x] Ensure that this works for a newly generated app using 3p cookies

### Things to focus on

- The [return address](https://github.com/Shopify/shopify_app/pull/1259/files#diff-d73da7552f28e00cec3da61e695cfe70a8b49c4e2c403bc7062863bbf9fa185cR226) after completing OAuth

### Checklist

### DIFF
--- a/lib/generators/shopify_app/home_controller/templates/home_controller.rb
+++ b/lib/generators/shopify_app/home_controller/templates/home_controller.rb
@@ -3,8 +3,16 @@
 class HomeController < AuthenticatedController
   include ShopifyApp::ShopAccessScopesVerification
 
+  before_action :set_host
+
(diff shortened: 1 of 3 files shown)


hannachen (approved):
🎩 with a rails app generated using `shopify-cli`. The redirect back to the app's root no longer result in a missing host error.

PR #1265, opened 2021-05-05
Move js for POST redirect back to separate file

greatestape:
This change adopts an earlier implementation we'd explored for the POST redirect page. Rather than embed the JS in an inline `script` tag, we're shipping it in a separate file.

This is an immediate response to an issue that was flagged in 18.0.0. This fix introduces a race condition for local dev environments. We don't love that, but it's more important to us to make sure that:

1. Our latest release doesn't have a feature that might be incompatible with a consumer's CSP
1. We still have a release out there that unblocks devs who want to update to the latest omniauth

### What this PR does

* Move JS for POST redirect back to separate file
* Add file to asset manifest

* Install the gem on an existing app and verify that the oauth install flow still works
  * NOTE: If your local dev env uses the `cookie_store` session storage strategy, you may encounter 401 errors during oauth due to a race condition between asset requests and `/auth/shopify`. You should be able to work around for local testing by using a different browser or session storage strategy.
* Disabling source mapping in Chrome should eliminate the local dev race condition

### Things to focus on

* Does this change work for apps with Content Security Policies?
* Is the race condition described above acceptable? i.e. is it limited to local dev environments?

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 Unreleased
 ----------
-* Fix bug causing oauth flow to fail due to CSP violation
+* Fix bug causing OAuth flow to fail due to CSP violation. [#1265](https://github.com/Shopify/shopify_app/pull/1265)
 
 18.0.0 (May 3, 2021)
 ----------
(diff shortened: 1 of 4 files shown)


DAmatheson (approved):
Tested with an app and confirmed this unblocked the install flow when unsafe-inline is disallowed.

PR #1282, opened 2021-06-14
Fix webhooks handling in Ruby 3

kirillplatonov:
### What this PR does

Webhooks handling is currently broken on Ruby 3. It will incorrectly pass job params since keywords arguments changes in Ruby 3 (see [this](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/) for reference). Here's an error that we get:
CleanShot 2021-06-14 at 17 09 12@2x

To fix it we need to use double splat when passing the hash to treat it as keywords args in Ruby 3.

### Checklist


kirillplatonov:
Closing in flavor of https://github.com/Shopify/shopify_app/pull/1169

PR #1338, opened 2021-12-20
Update arbitrary domain names used in our documentation to a reserved second level domain name

fbeaudoin-shopify:
### What this PR does
Our documentation is using arbitrary domain names in its code snippets and to eliminate the risk of having an app developer miss-configuring their application with these, we should use invalid hosts in our examples.

I've used `example.com` which is a reserved domain: https://en.wikipedia.org/wiki/Example.com

### Things to focus on
1. Any typo?
2. Did I miss any other arbitrary domain names that we use in our `.md` files?

### DIFF
--- a/docs/shopify_app/engine.md
+++ b/docs/shopify_app/engine.md
@@ -15,7 +15,7 @@ While you can customize the login view by creating a `/app/views/shopify_app/ses
 
 ```ruby
 ShopifyApp.configure do |config|
-  config.login_url = 'https://my.domain.com/nested/login'
+  config.login_url = 'https://example.com/nested/login'
 end
 ```
 
@@ -77,6 +77,6 @@ class ReviewsController < ApplicationController
 end
 ```
 
-Create your app proxy URL in the [Shopify Partners dashboard](https://partners.shopify.com/organizations), making sure to point it to `https://your_app_website.com/app_proxy`.
+Create your app proxy URL in the [Shopify Partners dashboard](https://partners.shopify.com/organizations), making sure to point it to `https://example.com/app_proxy`.
 
 ![Creating an App Proxy](/images/app-proxy-screenshot.png)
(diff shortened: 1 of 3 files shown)


docs/shopify_app/engine.md
```diff
@@ -15,7 +15,7 @@ While you can customize the login view by creating a `/app/views/shopify_app/ses
 
 ```ruby
-  config.login_url = 'https://my.domain.com/nested/login'
+  config.login_url = 'https://my.domain.invalid/nested/login'
```
andyw8, line 18: There are a specially reserved domains for this purpose: https://en.wikipedia.org/wiki/Example.com

docs/shopify_app/script-tags.md
```diff
@@ -11,7 +11,7 @@ As with webhooks, ShopifyApp can manage your app's [ScriptTags](https://shopify-
 ```ruby
   config.scripttags = [
-    {event:'onload', src: 'https://my-shopifyapp.herokuapp.com/fancy.js'},
+    {event:'onload', src: 'https://my-shopfiyapp.herokuapp.invalid/fancy.js'},
```
andyw8, line 14: Typo in `shopfiyapp`

docs/shopify_app/webhooks.md
```diff
@@ -66,7 +66,7 @@ The WebhooksManager uses ActiveJob. If ActiveJob is not configured then by defau
 ShopifyApp can create webhooks for you using the `add_webhook` generator. This will add the new webhook to your config and create the required job class for you.
 
 ```
-rails g shopify_app:add_webhook -t carts/update -a https://example.com/webhooks/carts_update
+rails g shopify_app:add_webhook -t carts/update -a https://example-app.invalid/webhooks/carts_update
```
andyw8, line 69: example.com is reserved so this is safe.

docs/shopify_app/engine.md
fbeaudoin-shopify: TIl. Changed it to `example.com`

PR #1341, opened 2022-01-05
Fix config.hosts to use a regex that matches ngrok.io subdomains

sshaw:
### What this PR does

Allows one to use ngrok's free tier with embedded apps. 

- Create embedded app
- `rails s`
- `ngrok http 3000`

Without my change host validation fails. Per [the Rails guide (6.1)](https://guides.rubyonrails.org/v6.1/configuring.html#configuring-middleware): 

    The provided regexp will be wrapped with both anchors (\A and \z) so it must match the entire hostname. /product.com/, for example, once anchored, would fail to match www.product.com.

ngrok hosnames include hyphens, e.g., `1a07-138-51-91-7.ngrok.io`. Prior to my change `\w` was used which includes `[0-9a-zA-Z_]` **not hyphen** (`-`). 

### Checklist

(This should not be done by PR author but by maintainer IMHO)

### DIFF
--- a/lib/generators/shopify_app/install/install_generator.rb
+++ b/lib/generators/shopify_app/install/install_generator.rb
@@ -67,7 +67,7 @@ module ShopifyApp
       def insert_hosts_into_development_config
         inject_into_file(
           'config/environments/development.rb',
-          "  config.hosts = (config.hosts rescue []) << /\\w+\\.ngrok\\.io/\n",
+          "  config.hosts = (config.hosts rescue []) << /.+\\.ngrok\\.io/\n",
           after: "Rails.application.configure do\n"
         )
       end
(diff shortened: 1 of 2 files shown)


lib/generators/shopify_app/install/install_generator.rb
```diff
@@ -67,7 +67,7 @@ def mount_engine
       def insert_hosts_into_development_config
         inject_into_file(
           'config/environments/development.rb',
-          "  config.hosts = (config.hosts rescue []) << /\\w+\\.ngrok\\.io/\n",
+          "  config.hosts = (config.hosts rescue []) << /.+\\.ngrok\\.io/\n",
```
sshaw, line 70: Do we really want to do proper subdomain validation? I think not. We just care it ends with ngrok.io  🔚

sshaw: Also why `(config.hosts rescue []`)?

kirillplatonov: Why not just use `config.hosts.clear` in development env?

gonzaloriestra:
Thanks for the contribution! I hadn't seen this, but the regex has been fixed as part of https://github.com/Shopify/shopify_app/pull/1354, so I'm closing here.

PR #1342, opened 2022-01-07
Fix webhooks handling in Ruby 3

kirillplatonov:
Reopened https://github.com/Shopify/shopify_app/pull/1282

### What this PR does

Webhooks handling is currently broken on Ruby 3. It will incorrectly pass job params since keywords arguments changes in Ruby 3 (see this for reference). Here's an error that we get:
image

This PR updated webhooks controller to pass job params in a compatible way.


gonzaloriestra (approved):
@kirillplatonov sure! Thanks for the PR 👏

PR #1343, opened 2022-01-07
Update to Ruby 3

kirillplatonov:
- Updates default Ruby version as 2.5 is not maintained anymore
- Adds Ruby 3 to CI


kirillplatonov:
cc @NabeelAhsen

kirillplatonov:
@gonzaloriestra updated. Ready for merge.

.github/workflows/build.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     name: Ruby ${{ matrix.version }}
     strategy:
       matrix:
-        version: [2.5, 2.6, 2.7]
+        version: [2.5, 2.6, 2.7, '3.0']
```
gonzaloriestra, line 15: Let's better remove 2.5 to just keep supported versions. Also, does 3.0 require the quotes? 🤔 

```suggestion
        version: [2.6, 2.7, 3.0]
```

gonzaloriestra (approved):
Thanks for your help! ❤️

.github/workflows/build.yml
kirillplatonov: Removed 2.5. Quotes for `3.0` are required due to a bug in Github Actions runner: https://github.com/actions/runner/issues/849

gonzaloriestra:
I'm trying to merge this, but I cannot remove the Ruby 2.5 check. I've tried to create another PR, but it's still there... I guess we need to update the project settings, but I don't have permission, so I'll ask for help.

If you don't mind, we can close here and continue in https://github.com/Shopify/shopify_app/pull/1359, which includes a couple of more things (I've added you as a co-author).

PR #1345, opened 2022-01-07
Add AppBridgeMiddleware

kirillplatonov:
### What this PR does

When the app is installed via "Add app" button from the app store listing Shopify redirects the user to the login page without `host` param:
CleanShot 2022-01-07 at 18 47 36@2x

As a result, the app gets `AppBridge2 error` on the frontend and the user is stuck on the Splash screen. I had issues with this on production after AppBridge2.0 upgrade. I was talking with Shopify Partner support about this behavior and they said it's not a bug but the intended behavior 🤷‍♂️ Until this will change we need a middleware that will generate `host` from the `shop` param on the fly.

### DIFF
--- a/lib/shopify_app/middleware/app_bridge_middleware.rb
+++ b/lib/shopify_app/middleware/app_bridge_middleware.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+module ShopifyApp
+  class AppBridgeMiddleware
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      request = Rack::Request.new(env)
+
+      if request.params.has_key?("shop") && !request.params.has_key?("host")
+        shop = request.params["shop"]
+        host = Base64.urlsafe_encode64("#{shop}/admin", padding: false)
+        request.update_param("host", host)
+      end
+
+      @app.call(env)
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


kirillplatonov:
cc @NabeelAhsen

NabeelAhsen:
When the "Add App" button is clicked, the app is prompted to initiate OAuth by visiting the app's login URL: `https://<app>/login?shop=<shop>..`

The OAuth flow completes and correctly returns the `host` parameter to the app. Why does your app redirect to the splash page when sent to `/login`?

lib/shopify_app/middleware/app_bridge_middleware.rb
```diff
@@ -0,0 +1,20 @@
...
+      @app = app
+    end
+
+    def call(env)
+      request = Rack::Request.new(env)
+
+      if request.params.has_key?("shop") && !request.params.has_key?("host")
+        shop = request.params["shop"]
+        host = Base64.urlsafe_encode64("#{shop}/admin", padding: false)
+        request.update_param("host", host)
```
NabeelAhsen, line 14: From my understanding, the host parameter is subject to change; it's not _always_ going to be the shop Base64 encoded.

NabeelAhsen:
@jesalerno84 Could I get your input here? I'm not sure that this is a bug

kirillplatonov:
@NabeelAhsen I just double-checked and everything seems to work correctly without middleware. I definitely was able to reproduce this issue in November and had "APP::ERROR::INVALID_CONFIG: host must be provided" errors in Sentry after app install. Not sure if it was before or after login, to be honest. Let's close this for now then. I will keep an eye on Sentry and will let you know if the issue will happen again.

PR #1358, opened 2022-01-28
Check for `spin.dev` URLs in shopify domain

sainagasrikantham:
### What this PR does

Launching Shopify Core in spin results in a Shopify domain URL of the style `shop1.shopify.{spin-workspace-id}.{user-name}.us.spin.dev` which is different from the currently expected URLs of type `*.myshopify.com`.

### Things to focus on

1. Logic change in `lib/shopify_app/utils.rb`

### Checklist

### DIFF
--- a/lib/shopify_app/utils.rb
+++ b/lib/shopify_app/utils.rb
@@ -3,12 +3,17 @@ module ShopifyApp
   module Utils
     def self.sanitize_shop_domain(shop_domain)
       myshopify_domain = ShopifyApp.configuration.myshopify_domain
+      spin_domain = 'spin.dev'
       name = shop_domain.to_s.downcase.strip
       name += ".#{myshopify_domain}" if !name.include?(myshopify_domain.to_s) && !name.include?(".")
       name.sub!(%r|https?://|, '')
 
       u = URI("http://#{name}")
-      u.host if u.host&.match(/^[a-z0-9][a-z0-9\-]*[a-z0-9]\.#{Regexp.escape(myshopify_domain)}$/)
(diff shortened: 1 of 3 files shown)


lib/shopify_app/utils.rb
```diff
@@ -3,12 +3,17 @@ module ShopifyApp
   module Utils
     def self.sanitize_shop_domain(shop_domain)
       myshopify_domain = ShopifyApp.configuration.myshopify_domain
+      spin_domain = 'spin.dev'
```
t6d, line 6: Instead of treating `spin.dev` special, can we maybe expand the functionality of `myshopify_domain` and allow it to be an array? I'm thinking

```ruby
ShopifyApp.configuration.myshopify_domain = ["myshopify.com", "spin.dev"]

myshopify_domains = Array(Shopify.configuration.myshopify_domain)
u = URI("http://#{name}")
valid_domain = myshopify_domains.any? do |myshopify_domain_suffix|
	u.host&.match(/^[a-z0-9][a-z0-9\-]*[a-z0-9]\.#{Regexp.escape(myshopify_domain_suffix)}$/)
end

return u.host if valid_domain
nil
```

This will of course require some double checking of where `myshopify_domain` is used to ensure if somebody sets it to an Array that it still behaves correctly in all other places.

sainagasrikantham: @t6d 

@Bananajunk and I tried to use the array style you mentioned. We can't create a `URI` for `shop_domain` that end with `spin.dev` because we never know if it is a fully `URI` or not.

We simply want URLs ending with `spin.dev` be treated as a valid domain. We don't want to complicate the configuration piece of `myshopify_domain` because devs are used it to being a single value not an array.

PR #1360, opened 2022-01-31
Add ShopHost concern for fetching and saving host

kirillplatonov:
### The problem

Right now in AppBridge 2.0 apps, it's not possible to open any app link in a new tab (it won't load without `host` param). The only workaround is to pass `host` to every link in the app which is pretty hard to manage.

### What this PR does

This PR introduces `ShopifyApp::ShopHost` concern which handles `host` fetching and caches it in cookies. This allows avoiding passing host param manually to every link. I found this concern very convenient and use it in all of my apps.

### Checklist


kirillplatonov:
Hey @gonzaloriestra @NabeelAhsen! 
Any chance this PR could be merged? Happy to make adjustments if needed.
Right now opening any links in App Bridge 2.0 apps in the new tab is broken which is not great UX for merchants.

CodeSchneider:
Hi @kirillplatonov I recently posted an [issue](https://github.com/Shopify/shopify_app/issues/1387) for this gem. Do you think adding the `host` param would solve our problem. Currently in our Rails controller the redirect logic looks like:

`fullpage_redirect_to charge.confirmation_url`

Are you proposing something like this? 

`fullpage_redirect_to charge.confirmation_url + 'host=' + (@shop_domain + '/admin')`

I'm a little confused where to add the host param, does it need to be base64 encoded etc.

Thanks for your help.

kirillplatonov:
@gonzaloriestra sounds good. Resolved conflicts in this PR so it's ready to merge.

hoppergee:
Inspired by this PR, I came up with another way to fetch host without cookies. And it works well with an app on hand. https://github.com/Shopify/shopify_app/pull/1403

flavio-b:
Right now, AppBridge JWTs have a payload with lots of info. Wouldn't it make more sense to include the host parameter in this payload too, then build a backend helper to access it everywhere? @hoppergee, what do you think?



Screen Shot 2022-06-08 at 1 03 30 PM

PR #1362, opened 2022-02-01
Move app data back into body

hannachen:
### What this PR does

Follow-up to https://github.com/Shopify/shopify_app/pull/1247/

An unintentional side-effect was introduced while cleaning up `.erb` templates. Moving app information from the `body` to a `<script>` tag violates the `unsafe-inline` Content Security Policy.
 
1. Clone this branch
1. In a terminal, use [Shopify CLI](https://shopify.dev/tools/cli) to create and configure a new rails app: `shopify app create rails`
1. Using a text editor, open the `Gemfile` inside the newly created app
1. Modify the entry of `shopify_app` to point to this branch in your local environment, it should looks similar to this:
    `gem 'shopify_app', path: '~/src/github.com/Shopify/shopify_app'`
1. In a terminal, start your app server with `shopify app serve`, this will also output a **ngrok url**
1. In a browser, follow the ngrok url to install the app on a test store. It will look similar to this:
    `https://1587ef2f72d5.ngrok.io/login?shop=sabotender-shop.myshopify.com`
    1. Test 1: On your local environment, remove the `shop` entry for the store inside the app db (this will simulate one of the cases)
    2. Test 2: Update your generated app scope inside file `config/intializers/shopify_app.rb` to trigger a redirect from within the shop frame
1. In a browser, open the installed app. i.e. `https://sabotender-shop.myshopify.com/admin/apps/test-app-1`
1. Ensure that the entire redirect flow takes the browser back to the app. 

### Things to focus on

1. Focus on install/uninstall
    1. Test 1: Case of a possible incomplete install, where the shop thinks the app is installed, but the app does not have the shop data
    2. Test 2: Case of an app updating access scope, and requires re-auth by breaking out of the shop frame
2. Other areas that might kick off a top level redirect, or use the methods `fullpage_redirect_to`, `authenticate_at_top_level`, or `enable_cookie_access` (Safari browser with strict 3rd party cookie policy)

### Checklist

### DIFF
--- a/app/assets/javascripts/shopify_app/app_bridge_redirect.js
+++ b/app/assets/javascripts/shopify_app/app_bridge_redirect.js
@@ -5,7 +5,7 @@
     var AppBridge = window['app-bridge'];
     var createApp = AppBridge.default;
     var Redirect = AppBridge.actions.Redirect;
-    var shopifyData = window.shopifyData;
+    var shopifyData = document.body.dataset;
 
     var app = createApp({
       apiKey: shopifyData.apiKey,
--- a/app/assets/javascripts/shopify_app/storage_access.js
+++ b/app/assets/javascripts/shopify_app/storage_access.js
@@ -126,7 +126,7 @@
 
   /* ITP 2.0 solution: handles cookie partitioning */
   StorageAccessHelper.prototype.setUpHelper = function() {
-    var shopifyData = window.shopifyData;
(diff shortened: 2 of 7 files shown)


henrytao-me (approved):
🎩-ed. LGTM

Screen Shot 2022-02-01 at 14 47 13

Issue #1377, opened 2022-02-25
Implement default Content-Security-Policy that prevents clickjacking


fbeaudoin-shopify:
## Details
The [Shopify CLI](https://github.com/Shopify/shopify-cli) uses the **shopify_app** gem to create a simple rails application by running `shopify app create rails`. Per our [security requirement](https://shopify.dev/apps/store/security/iframe-protection), apps on the Shopify App Store must set the proper Content Security Policy `frame-ancestors` directive to avoid clickjacking attacks. 

Therefore, the generated app should have a defined `Content-Security-Policy` that at least sets the `frame-ancestors` directive. As for other directives (i.e. `script-src`, `img-src`, ...), they should be as strict as possible.

resistorsoftware:
It totally sucks that this issue remains open since February with zero resolution. A simple App from this gem will not pass Shopify robot on Clickjacking check, due to the CSP.

resistorsoftware:
No. The problem is as I pointed out back then, the CSP seems to be run as part of the Rack response regardless of serving up a page in an embedded app or not. So during CSP initialization Shopify App gem helpers are not instantiated. Hence the Shopify App check that sends GDPR tests fails.

resistorsoftware:
@fbeaudoin-shopify So what exact code should be in this gem. You talk around the issue, but no one has provided a solution to the actual issue. 

Thanks though for explaining it in more detail than anywhere else, by anyone else, for the most part. We learn piece by piece. Slowly.

fbeaudoin-shopify:
@resistorsoftware Unfortunately the team owning the gem has better context than me to provide the implement hence this issue that I created.

resistorsoftware:
It is so weird that such a small issue can take so many months to iron out. In the meantime, the crap shoot with the robot continues.

denoyaro:
@fbeaudoin-shopify Since according to https://shopify.dev/apps/store/security/iframe-protection, the embedded app should only be frameable in an authenticated shop domain, shouldn't the fallback be `Content-Security-Policy: frame-ancestors 'none';`

nelsonwittwer:
#1474 should address the concerns pointed out here, so we are going to close this issue.

If you still encounter this issue with the latest stable version, please reopen using the issue template. You can also contribute directly by submitting a pull request– see the CONTRIBUTING.md(.github/CONTRIBUTING.md) file for guidelines

PR #1389, opened 2022-03-14
Add custom CSP generator template to include frame_ancestors

molly-yu:
### What this PR does

For context: https://github.com/Shopify/shopify_app/issues/1377

Essentially, apps published in the App Store must set the proper Content Security Policy frame-ancestors directive to avoid clickjacking attacks. Currently, the CSP configuration template (`content_security_policy.rb`) is generated by `rails generate ...` as part of rails. These changes overwrite the basic CSP configuration template so that we can add custom policies such as frame-ancestors.

Note: A small change in the shopify-cli is required to force the overwrite of the `content_security_policy.rb` file. 
Changes found here: https://github.com/Shopify/shopify-cli/pull/2143

Second Note: The template has the CSP configurations commented out, since not all apps will be published to the App Store, and thus may not need to set these policies.

- create a rails app, either through the cli (with the required change noted above), or manually (following the steps in rails_service.rb in shopify-cli)
- the app should be created successfully.
- navigate to `config/initializers/content_security_policy.rb` in the app, and observe that the CSP frame_ancestors directive is there.

To test that the CSP frame-ancestors directive is set in the app, uncomment out the contents of `content_security_policy.rb`, install + load the app, and ensure that the proper frame_ancestors is set, as outlined in instructions here: https://shopify.dev/apps/store/security/iframe-protection

### Things to focus on

1. Changes do not break app creation.
2. The CSP file in the app is generated properly with the correct changes.
3. The frame_ancestors are set properly when loading the app.

### Checklist

### DIFF
--- a/lib/generators/shopify_app/install/templates/content_security_policy.rb
+++ b/lib/generators/shopify_app/install/templates/content_security_policy.rb
@@ -11,21 +11,26 @@
 #     policy.font_src    :self, :https, :data
 #     policy.img_src     :self, :https, :data
 #     policy.object_src  :none
-#     policy.script_src  :self, :https, :unsafe_inline
-#     policy.style_src   :self, :https, :unsafe_inline
+#     policy.script_src  :self, :https
+#     policy.style_src   :self, :https
 #     # Specify frame ancestors for embedded apps
 #     # For more information, see the following documentation
 #     # https://shopify.dev/apps/store/security/iframe-protection
 #     policy.frame_ancestors :https, -> { "https://#{current_shopify_domain} https://admin.shopify.com" }
+
 #     # Specify URI for violation reports
 #     # policy.report_uri "/csp-violation-report-endpoint"
 #   end
-#
-#   # Generate session nonces for permitted importmap and inline scripts
-#   config.content_security_policy_nonce_generator = ->(request) { request.session.id.to_s }
-#   config.content_security_policy_nonce_directives = %w(script-src)
-#
-#   # Report CSP violations to a specified URI. See:
-#   # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only
-#   # config.content_security_policy_report_only = true
+
+# If you are using UJS then enable automatic nonce generation
+# Rails.application.config.content_security_policy_nonce_generator = -> request { SecureRandom.base64(16) }
+
+# Set the nonce only to specific directives
+# Rails.application.config.content_security_policy_nonce_directives = %w(script-src)
+
+# Report CSP violations to a specified URI
+# For further information see the following documentation:
+# https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only
+# Rails.application.config.content_security_policy_report_only = true
+
 # end


lib/generators/shopify_app/install/templates/content_security_policy.rb
```diff
@@ -0,0 +1,28 @@
...
+
+# Rails.application.configure do
+#   config.content_security_policy do |policy|
+#     policy.default_src :self, :https
+#     policy.font_src    :self, :https, :data
+#     policy.img_src     :self, :https, :data
+#     policy.object_src  :none
+#     policy.script_src  :self, :https
+#     policy.style_src   :self, :https
+#     # Specify frame ancestors for embedded apps
```
hannachen, line 15: Maybe we can add a link to our public docs site for additional context? https://shopify.dev/apps/store/security/iframe-protection

lib/generators/shopify_app/install/templates/content_security_policy.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+# Be sure to restart your server when you modify this file.
+
+# Define an application-wide content security policy
+# For further information see the following documentation
+# https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
+
+# Rails.application.configure do
```
paulomarg, line 8: Are we sure we want to keep all of this stuff commented out in the generated file? I feel like since we require CSP to be set for all apps, we should be opinionated about actually applying the settings here. What do you think?

paulomarg: I don't have a strong opinion on uncommenting the file, since the solution will be there in case anyone runs into issues. It might make sense to have the settings from the start, just in case folks try to develop a scenario that wouldn't work.

molly-yu:
PR Note: The template uses inline script tags to generate the embedded app content. For this reason, the sample app fails to load, resulting in the error @paulomarg found above. In the CSP error, it suggests to add `unsafe-inline`, however this could expose the app to potential malicious attacks such as Cross-Site Scripting (XSS). 

Thanks @Abdulwahaab710 for helping with the following solution: 
Instead of adding `unsafe-inline`, we could separate the script contents into their own `scriptFile.js` files and instead include them using `<%= javascript_pack_tag "scriptFile" %>` to resolve this error. 

However, the `app/javascript` directory seems to be auto-generated by Webpack, so I'm not sure if there's a good way to do this without messing with the directory structure too much. So for now, I think it might be best to leave the CSP template commented out.

hannachen (approved):
This looks good to me, it's probably safer to have the file commented out and add a note to https://shopify.dev/apps/store/security/iframe-protection to instruct developers to enable it by removing the comment. Also based on this feedback, it looks like the other CSP might be too strict: https://github.com/Shopify/shopify_app/issues/1405#issuecomment-1095350940

lib/generators/shopify_app/install/templates/content_security_policy.rb
```diff
@@ -0,0 +1,36 @@
...
+#     policy.default_src :self, :https
+#     policy.font_src    :self, :https, :data
+#     policy.img_src     :self, :https, :data
+#     policy.object_src  :none
+#     policy.script_src  :self, :https
+#     policy.style_src   :self, :https
+#     # Specify frame ancestors for embedded apps
+#     # For more information, see the following documentation
+#     # https://shopify.dev/apps/store/security/iframe-protection
+#     policy.frame_ancestors :https, -> { "https://#{current_shopify_domain} https://admin.shopify.com" }
```
fbeaudoin-shopify, line 19: How is the `current_shopify_domain` retrieve from the session? We should provide the example. 

Also, in a case that `current_shopify_domain` is `nil` (pre-auth) we should fallback to `[:https, "*.myshopify.com", "admin.shopify.com".

luigi-zarrelli-latori: Instead of checking for `nil` you will need to check for `defined?` and I totally agree with you that there has to be a fallback.

luigi-zarrelli-latori: Maybe I'm wrong with this and please correct me if I am as this issue seems to concern a lot of public app developers using this gem.

In the Rails documentation it says (see: https://guides.rubyonrails.org/security.html#content-security-policy-header) 

> Use lambdas to inject per-request values, such as account subdomains in a multi-tenant application:

with an example of

```ruby
class PostsController < ApplicationController
    policy.base_uri :self, -> { "https://#{current_user.domain}.example.com" }
  end
end
```

shouldn't that be the approach of this PR? I just see a lot of problems by having it as an initialized file where the context is not always present and it will also be called on requests to the application that might not need this policy at all (e.g. if you make a request to the application from the storefront instead it will break since `current_shopify_domain` will never be defined).

Looking at how this gem operates currently I would assume a controller concern that uses a similar way as this https://github.com/Shopify/shopify_app/blob/e5350a587594169b3506c8e63356df0844dabc7b/app/controllers/concerns/shopify_app/require_known_shop.rb#L12-L15 to fetch the shop domain would be more appropriate (and also not dependent on the `RequireKnownShop` concern).

I would really appreciate some communication on this.

luigi-zarrelli-latori:
Hey @hannachen @molly-yu @fbeaudoin-shopify @paulomarg I would really appreciate if you could please respond to this topic.

PR #1413, opened 2022-04-13
Fixing online token creation

paulomarg:
### What this PR does

We unintentionally dropped some important functionality from this gem as part of our updates of `shopify_api`: the OAuth flow was expected to create _both_ an offline and an online tokens when using user storage, but the updated flow was only doing one of them.

This PR walks that back by adding back the behaviour of first getting an offline token, then an online token, then completing the OAuth process, but only in cases where online tokens are required - offline flows remain the same (only a shop is added to the DB).

1. Create a new app with the CLI
2. Run `rails generate shopify_app:user_model`
3. Add `config.user_session_repository = 'User'` in shopify_app.rb
4. `shopify app serve` => go through login

### Things to focus on

1. Does oauth work for online tokens?
2. Does oauth work for offline tokens?

### Checklist

- [X] Update `CHANGELOG.md` if the changes would impact users

### DIFF
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -27,6 +27,12 @@ module ShopifyApp
         value: auth_result[:cookie].value,
       }
 
+      session[:shopify_user_id] = auth_result[:session].associated_user.id if auth_result[:session].online?
+
+      if start_user_token_flow?(auth_result[:session])
+        return respond_with_user_token_flow
+      end
+
       perform_post_authenticate_jobs(auth_result[:session])
 
       respond_successfully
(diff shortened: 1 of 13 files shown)


hannachen:
When I add the line `config.user_session_repository = 'User'` into `shopify_app.rb`, initial install takes me back to the app listing page with this error message after a few redirects:
Screen Shot 2022-04-14 at 12 00 32 PM

And the app doesn't seem to load any products once installed 🤔 
Screen Shot 2022-04-14 at 12 07 08 PM

I'm installing the app using the ngrok URL displayed in the `serve` command: `https://6764-2607-fea8-a71f-f5fc-54cd-5d00-d13f-9c69.ngrok.io/login?shop=sabotender-shop.myshopify.com`

hannachen (approved):
🎩 👍
I created an app with this template using the CLI, it's now working and no more redirect loops!

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -43,6 +49,25 @@ def respond_with_error
       redirect_to(login_url_with_optional_shop)
     end
 
+    def respond_with_user_token_flow
+      redirect_to(login_url_with_optional_shop)
+    end
+
+    def start_user_token_flow?(shopify_session)
+      return false unless ShopifyApp::SessionRepository.user_storage.present?
+      return false if shopify_session.online?
```
rezaansyed, line 58: how does `shopify_session.online?` check whether the response if for an online token?

hannachen: I think this is part of *Shopify API*, maybe this should be `is_online` instead? https://github.com/Shopify/shopify_api/blob/5a4785a40be012f7f3ec93ecb037508efff0c292/lib/shopify_api/auth/session.rb#L47

paulomarg:
@hannachen I noticed we weren't properly cleaning up all of the configuration values when setting up for tests, which led to this inconsistent behaviour. I've now changed the setup config so that it always resets all values (that I could find).

I also isolated a bit of code for mocking a session which was what actually broke in the tests as part of my changes.

app/controllers/shopify_app/callback_controller.rb
paulomarg: Yes, I believe that is the original behaviour here - if we want online tokens, we go through OAuth twice in a row to get both of them.

PR #1432, opened 2022-05-19
Bump browser_sniffer

rtymchyk:
### What this PR does

Bumps `browser_sniffer` to `2.0.0` to receive an update on [POS user agent detection](https://github.com/Shopify/browser_sniffer/pull/38) I've recently published.

Open to suggestions on how I can add a test case for this. Previous gem bumps here added a test to `sessions_controller` [but the test passes regardless of what user-agent is passed through](https://github.com/Shopify/shopify_app/pull/1331/files#diff-0ebf4c65ae927de1d02650f4caa7f2810e14eb64201b7570cb30939f0ee38719). 

A method that checks for POS browser name is [also marked private and not used anywhere](https://github.com/Shopify/shopify_app/blob/main/lib/shopify_app/controller_concerns/itp.rb#L29).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     shopify_app (19.0.2)
       activeresource
-      browser_sniffer (~> 1.4.0)
+      browser_sniffer (~> 2.0)
       jwt (>= 2.2.3)
       rails (> 5.2.1)
       redirect_safely (~> 1.0)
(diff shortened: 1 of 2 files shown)


EsterKais (approved):

AnthonyRobertson17 (approved):

rtymchyk:
This change has also been back-ported to v18 here https://github.com/Shopify/shopify_app/pull/1434

PR #1445, opened 2022-06-09
Allow custom OAuth URLs

paulomarg:
### What this PR does

In some cases, the Shopify routes might be mounted under a custom `root_url`, when the `Engine` is not mounted at `/`. Assuming that `root_url` is expected to match the root URL for the `Engine`, this PR adds support for customizing both the OAuth login and callback routes so that they can be arbitrarily determined by the app.

The goal here is for this to preserve the current behaviour by default, so I just wanted to make sure that assumption about `root_url` is actually correct.

Tested via unit tests and by using the gem locally on a CLI-generated app.

### Things to focus on

1. Is this changing any existing behaviour?
2. Are the test cases covering enough scenarios?
3. Is it the right thing to strip out `root_url` from the URLs when setting up the routes? That is where the assumption about `root_url` comes into play.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased
 ----------
 
+* Add the `login_callback_url` config to allow overwriting that route as well, and mount the engine routes based on the configurations. [#1445](https://github.com/Shopify/shopify_app/pull/1445)
+
 19.0.2 (April 27, 2022)
 ----------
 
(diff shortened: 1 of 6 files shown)


gonzaloriestra (approved):
I haven't tested it, but it LGTM!

hannachen (approved):
This looks good to me 👍
🎩 with CLI 2.0 and did a quick smoke test

PR #1450, opened 2022-06-13
Set special redirection headers on 401s

paulomarg:
### What this PR does

To make it easier for embedded apps to redirect to OAuth after a token expires and a request reaches the server, we set the special `X-Shopify-API-Request-Failure-Reauthorize-Url` header, so that the client can know when to trigger an App Bridge redirect to the login action.

This was mostly tested via unit tests.

### Checklist


hrdwdmrbl:
Where is the accompanying client code which recognizes and acts on the special header?

PR #1455, opened 2022-06-17
Add support for billing for apps

paulomarg:
### What this PR does

Partners might want to charge for their apps, but this gem currently doesn't really add any specific support for that scenario.

This PR adds that support by appending a concern to `Authenticated` to check if there is an active payment for the merchant, or requesting one if there isn't, and redirecting the merchant to it.

If the new `billing` configuration setting isn't present, this does nothing.

This was mostly tested by unit tests, and by activating the configuration in a new app.

### Things to focus on

1. Does this break any pre-existing flows? As in, are there cases where we might try to apply billing when we shouldn't?
2. Is the test coverage good?

### Checklist


mkevinosullivan (approved):

immakdas:
The main scenario is not covered so it is useless
1. merchant user clicks to Install the app
1inst
2. on the payment page customer click cancel
2cancels
3. as a result App is installed but not paid and the user can Open it for free
3free

tested with the shopify_app 20.0.2
I believe the Trial scenario is also not covered
```
config.billing = ShopifyApp::BillingConfiguration.new(
     charge_name: "My app billing charge",
     amount: 5,
     interval: ShopifyApp::BillingConfiguration::INTERVAL_EVERY_30_DAYS,
     currency_code: "USD",
     trial: 1,
  )```

return
`gems/shopify_app-20.0.2/lib/shopify_app/configuration.rb:119:in `initialize': unknown keyword: :trial (ArgumentError)`

PR #1474, opened 2022-07-12
Add CSP headers to unauthenticated controller

paulomarg:
### What this PR does

As per https://github.com/Shopify/shopify_app/pull/1389#issuecomment-1139841072 in a different attempt at adding CSP headers, we should make sure to only include the `frame-ancestors` directive on requests that are actually for embedded apps.

This PR attempts to solve that problem using the suggested lambda approach from the Rails docs and creating a controller concern specifically for this purpose, which is also included in the existing `EmbeddedApp` concern, so that controllers marked as embedded app ones are always setting the appropriate value for that directive.

If you use a local clone of this gem in an app created with the CLI's ruby template, you should see this header in the document request for `/` when loading the app within the admin:

```
content-security-policy: frame-ancestors https://<shop>.myshopify.com https://admin.shopify.com;
```

### Things to focus on

1. Is setting the directive on this concern enough? Are there other places we should be doing that?

### Checklist


luigi-zarrelli-latori:
@paulomarg did you already test if this PR passes the app submission security robot? I tried a similar approach and it didn't pass it, sadly there's no further information or details provided in the rejection email.

bernardoamc:
@luigi-zarrelli-latori Do you mind sharing the name of the app used? We would also like to double check to guarantee that the checks are passing. And thank you for testing! 🙏

luigi-zarrelli-latori:
> @luigi-zarrelli-latori Do you mind sharing the name of the app used? We would also like to double check to guarantee that the checks are passing. And thank you for testing! 🙏

```ruby
# frozen_string_literal: true

module EmbeddedAppFrameAncestors
  extend ActiveSupport::Concern

  included do
      policy.frame_ancestors(-> { "https://#{current_shopify_domain} https://admin.shopify.com;" })
    end
  end
end
```

`Version: shopify_app (19.1.0)` SPA and only included in the home controller. So far no rejection 🤞

PR #1483, opened 2022-08-02
Process embedded parameter to optimize OAuth

mkevinosullivan:
### What this PR does

We want to improve OAuth experience by:

- Making it faster
- Reducing the amount of screen flicker.

Note, this PR will require https://github.com/Shopify/shopify-api-ruby/pull/1002 to be merged first, the TODO's updated accordingly, before it can be merged itself

Fixes https://github.com/Shopify/first-party-library-planning/issues/393

To be documented

### Things to focus on

1. Top-level: App installation from the app store 
1. Top-level: Token refresh scopes change (grant screen present)
1. Top-level: Token refresh no scopes change (grant screen skipped)
1. Iframe: Token refresh scopes change (grant screen present)
1. Iframe:Token refresh no scopes change (grant screen skipped)

### Checklist

- Update `README.md`, if appropriate - not applicable
- For security fixes, the [Disclosure Policy](https://github.com/Shopify/shopify_app/blob/master/SECURITY.md#disclosure-policy) must be followed - not applicable

### DIFF
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -44,9 +44,9 @@ module ShopifyApp
 
     def respond_successfully
       if ShopifyAPI::Context.embedded?
-        # TODO: replace this param with ShopifyAPI::Utils::AppUrl.get_embedded or whatever the new
+        # TODO: replace this param with ShopifyAPI::Auth.embedded_app_url or whatever the new
         # method name will be, once https://github.com/Shopify/shopify-api-ruby/pull/1002 is merged
-        redirect_to(embedded_app_url, allow_other_host: true)
+        redirect_to(embedded_app_url(params[:host]), allow_other_host: true)
       else
         redirect_to(return_address)
       end
@@ -54,7 +54,7 @@ module ShopifyApp
 
     # TODO: remove this once https://github.com/Shopify/shopify-api-ruby/pull/1002
     # is merged and released
-    def embedded_app_url
+    def embedded_app_url(host)
       decoded_host = Base64.decode64(host)
       "https://#{decoded_host}/apps/#{ShopifyAPI::Context.api_key}"
     end
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -46,7 +46,7 @@ module ShopifyApp
       if ShopifyAPI::Context.embedded?
         # TODO: replace this param with ShopifyAPI::Utils::AppUrl.get_embedded or whatever the new
         # method name will be, once https://github.com/Shopify/shopify-api-ruby/pull/1001 is merged
-        redirect_to(embedded_app_url)
+        redirect_to(embedded_app_url, allow_other_host: true)
       else
         redirect_to(return_address)
       end
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -46,7 +46,8 @@ module ShopifyApp
       if ShopifyAPI::Context.embedded?
         # TODO: replace this param with ShopifyAPI::Auth.embedded_app_url or whatever the new
         # method name will be, once https://github.com/Shopify/shopify-api-ruby/pull/1002 is merged
-        redirect_to(embedded_app_url(params[:host]), allow_other_host: true)
+        return_to = session.delete(:return_to) || ""
+        redirect_to(embedded_app_url(params[:host]) + return_to, allow_other_host: true)
       else
         redirect_to(return_address)
       end
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -43,7 +43,7 @@ module ShopifyApp
     private
 
     def respond_successfully
-      if ShopifyAPI::Context.is_embedded?
+      if ShopifyAPI::Context.embedded?
         # TODO: replace this param with ShopifyAPI::Utils::AppUrl.get_embedded or whatever the new
         # method name will be, once https://github.com/Shopify/shopify-api-ruby/pull/1001 is merged
         redirect_to(embedded_app_url)
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -44,20 +44,11 @@ module ShopifyApp
 
     def respond_successfully
       if ShopifyAPI::Context.embedded?
-        # TODO: replace this param with ShopifyAPI::Auth.embedded_app_url or whatever the new
-        # method name will be, once https://github.com/Shopify/shopify-api-ruby/pull/1002 is merged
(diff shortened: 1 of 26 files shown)


lib/shopify_app/controller_concerns/sanitized_params.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    def sanitize_shop_param(params)
+      return unless params[:shop].present?
+      ShopifyApp::Utils.sanitize_shop_domain(params[:shop])
+    end
+
+    def sanitized_params
+      request.query_parameters.clone.tap do |query_params|
+        if params[:shop].is_a?(String)
```
paulomarg, line 29: Should this check be in `sanitize_shop_param`?

lib/shopify_app/controller_concerns/sanitized_params.rb
```diff
@@ -0,0 +1,35 @@
...
+        referer_uri = URI(request.referer)
+        query_params = Rack::Utils.parse_query(referer_uri.query)
+
+        sanitize_shop_param(query_params.with_indifferent_access)
+      end
+    end
+
+    def sanitize_shop_param(params)
+      return unless params[:shop].present?
+      ShopifyApp::Utils.sanitize_shop_domain(params[:shop])
```
paulomarg, line 24: I wonder if this would be better placed in the library since it might be useful for non-rails apps too.

test/controllers/callback_controller_test.rb
```diff
@@ -25,6 +26,10 @@ class CallbackControllerTest < ActionController::TestCase
     setup do
       @routes = ShopifyApp::Engine.routes
       ShopifyApp.configuration = nil
+      ShopifyApp.configuration.api_key = "api_key"
+      ShopifyApp.configuration.secret = "secret"
+      ShopifyApp.configuration.api_version = "2022-07"
+      ShopifyApp.configuration.scope = "read_orders, write_orders"
```
paulomarg, line 32: I think the test configurer should load these values with reasonable values, so we wouldn't need to do it here.

test/controllers/callback_controller_test.rb
```diff
@@ -141,14 +146,25 @@ class CallbackControllerTest < ActionController::TestCase
       end
     end
 
-    test "#callback redirects to the root_url with shop and host parameter" do
+    test "#callback redirects to the root_url with shop and host parameter for non-embedded" do
+      ShopifyApp.configuration.embedded_app = false
+      setup_context
```
paulomarg, line 151: Same here, the context is set up for tests automatically based on the configs set in `setup`, so we wouldn't need to have a separate method to do that here.

test/controllers/sessions_controller_test.rb
```diff
@@ -19,6 +19,13 @@ class SessionsControllerTest < ActionController::TestCase
 
       request.env["HTTP_USER_AGENT"] = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML,
                  like Gecko) Chrome/69.0.3497.100 Safari/537.36'
+
+      ShopifyApp.configuration = nil
```
paulomarg, line 23: Same comments about configs / setting up context here.

test/controllers/sessions_controller_test.rb
```diff
@@ -196,6 +210,79 @@ class SessionsControllerTest < ActionController::TestCase
       end
     end
 
+    ["my-shop", "my-shop.myshopify.com", "https://my-shop.myshopify.com",
+     "http://my-shop.myshopify.com",].each do |good_url|
```
paulomarg, line 214: Absolute nit, and completely my personal preference, so I apologize in advance for being pedantic, but what do you think about listing one of these per line? Makes it easier to grok that it's multiple test cases IMO.

paulomarg (approved):
Good test coverage, mostly just nits!

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -43,7 +43,21 @@ def callback
     private
 
     def respond_successfully
-      redirect_to(return_address)
+      if ShopifyAPI::Context.embedded?
+        # TODO: replace this param with ShopifyAPI::Auth.embedded_app_url or whatever the new
```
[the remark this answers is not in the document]
mkevinosullivan: Will need a new release of the Ruby Shopify API lib first

lib/shopify_app/controller_concerns/sanitized_params.rb
mkevinosullivan: This has been like this for ~4 years (https://github.com/Shopify/shopify_app/pull/571/files) ... I'm just moving it out of `login_protection.rb` to be accessible elsewhere.

lib/shopify_app/controller_concerns/sanitized_params.rb
mkevinosullivan: Let's do that as a separate PR/refactoring

test/controllers/sessions_controller_test.rb
```diff
@@ -254,5 +386,32 @@ def assert_redirected_to_top_level(shop_domain, expected_url = nil)
...
+        is_private: false,
+        is_embedded: ShopifyApp.configuration.embedded_app,
+        session_storage: ShopifyApp::SessionRepository,
+      )
+    end
+
+    def with_embedded_redirect_url
+      ShopifyApp.configuration.embedded_redirect_url = "/a-redirect-page"
+      yield
+      ShopifyApp.configuration.embedded_redirect_url = nil
```
paulomarg, line 414: Should we just always set this value to `nil` in the `test_helper` instead of resetting it here?

paulomarg (approved):
Just a nit

shopify_app.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency("jwt", ">= 2.2.3")
   s.add_runtime_dependency("rails", "> 5.2.1")
   s.add_runtime_dependency("redirect_safely", "~> 1.0")
-  s.add_runtime_dependency("shopify_api", "~> 11.0")
+  s.add_runtime_dependency("shopify_api", "~> 11.1.0")
```
paulomarg, line 22: I think we shouldn't be locking down the minor version here, right?

```suggestion
  s.add_runtime_dependency("shopify_api", "~> 11.1")
```

lib/generators/shopify_app/home_controller/templates/unauthenticated_home_controller.rb
```diff
@@ -6,7 +6,11 @@ class HomeController < ApplicationController
   include ShopifyApp::ShopAccessScopesVerification
 
   def index
-    @shop_origin = current_shopify_domain
-    @host = params[:host]
+    if ShopifyAPI::Context.embedded? && !params[:embedded].present? || params[:embedded] != "1"
```
andyw8, line 9: (nit) can we add parens here to make the precedence clear?

PR #1485, opened 2022-08-09
Move setup_context to ShopifyAppConfigurer

mkevinosullivan:
### What this PR does

Moves call to `setup_context` from tests to `ShopifyAppConfigurer`, therefore globally set before each test case.

`setup` in some test suites need to issue a call to `setup_context` if the `ShopifyApp` configuration changes.


paulomarg (approved):
This all makes sense to me, I wonder if in a future PR we want to revisit the way we set up the `Context` - it's awkward having to set the config and then still have to set up the context, I wonder if there's a way to trigger that automatically from the configuration class itself so that apps don't need to worry about it. That would really clean up the initializer file.

PR #1492, opened 2022-08-23
Webhooks generator with path instead of address

nelsonwittwer:
### What this PR does

The [Ruby API](https://github.com/Shopify/shopify-api-ruby/blob/main/lib/shopify_api/webhooks/registrations/http.rb#L12) handles enforcing of HTTPS and the host and is expecting `shopify_app` consumers to provide a path instead of an address.

This will fix the generators as well as a few others bugs found in the process.

### Reviewer's guide to testing

1. Create a new rails app
2. Install this gem in Gemfile:

```ruby
gem "shopify_app", git: 'https://github.com/shopify/shopify_app.git', branch: 'nelsonwittwer/webhook_fixes'
```
3. Bundle install
4. run the generator `rails g shopify_app` in the directory of your new app
5. run the generator `rails g shopify_app:add_webhook --topic orders/create --path webhooks/create_orders` to generate a webhook
6. Boot up rails `rails s`

### Things to focus on

We should no longer be seeing [an error raise complaining about not being able to find the job handler generated to handle webhook](https://github.com/Shopify/shopify_app/issues/1452). It should just work ™️ on rails boot.

### Checklist

Before submitting the PR, please consider if any of the following are needed:

- [x] Update `CHANGELOG.md` if the changes would impact users
- [x] Update `README.md`, if appropriate.
- [x] Update any relevant pages in `/docs`, if necessary
- [x] For security fixes, the [Disclosure Policy](https://github.com/Shopify/shopify_app/blob/master/SECURITY.md#disclosure-policy) must be followed.

PR #1495, opened 2022-09-01
Deduce webhook path from address

nelsonwittwer:
### What this PR does

As discussed in #1492 , the Ruby API is [expecting the webhook path](https://github.com/Shopify/shopify-api-ruby/blob/7a08ae9d96a7a85abd0113dae4eb76398cba8c64/lib/shopify_api/webhooks/registrations/http.rb#L12) instead of webhook address.

This will deduce the webhook's path if the engine is configured with the deprecated `address` attribute.

To make the `:address` -> `:path` webhook 

1. Create a new rails app
2. Install this gem in Gemfile:

```ruby
gem "shopify_app", git: 'https://github.com/shopify/shopify_app.git', branch: 'nelsonwittwer/deduce_path'
```
3. Bundle install
4. run the generator `rails g shopify_app` in the directory of your new app
5. run the generator `rails g shopify_app:add_webhook --topic orders/create --path webhooks/create_orders` to generate a webhook
6. Modify the `config/initializers/shopify_app.rb` webhooks section to use the webhook address instead of path.
7. Boot up rails `rails s`

### Checklist

### DIFF
--- a/lib/shopify_app/managers/webhooks_manager.rb
+++ b/lib/shopify_app/managers/webhooks_manager.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "uri"
+
 module ShopifyApp
   class WebhooksManager
     class CreationFailed < StandardError; end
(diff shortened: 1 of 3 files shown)


lib/shopify_app/managers/webhooks_manager.rb
```diff
@@ -38,18 +40,37 @@ def add_registrations
...
+            handler: webhook_job_klass(webhook_path),
             fields: attributes[:fields]
           )
         end
       end
 
       private
 
+      def path(webhook_attributes)
+        path = webhook_attributes.dig(:path)
```
andyw8, line 58: (nit) why `dig` here if it's not nested?

lib/shopify_app/managers/webhooks_manager.rb
```diff
@@ -38,18 +40,37 @@ def add_registrations
...
+        address = webhook_attributes.dig(:address)
+        uri = URI(address) if address
+
+        # Fail unless path or address are present
+        if path.nil? && address.nil?
+          raise ShopifyApp::MissingWebhookJobError("The :path attribute is required for webhook registration.")
```
andyw8, line 64:
```suggestion
          raise ShopifyApp::MissingWebhookJobError, "The :path attribute is required for webhook registration.")
```
I'm not sure if they are exactly equivalent but I think this is the usual form?

lib/shopify_app/managers/webhooks_manager.rb
```diff
@@ -38,18 +40,37 @@ def add_registrations
...
+        elsif path.present?
+          path
+        elsif uri&.path&.present?
+          uri.path
+        else
+          raise ShopifyApp::MissingWebhookJobError("The :path attribute is required for webhook registration.")
```
andyw8, line 70: Is it right that both clauses raise the same error?

andyw8 (approved):

PR #1503, opened 2022-09-12
Improve OAuth tunnel docs

thisdotrob:
### What this PR does

This commit removes duplicate sections on how to set up a tunnel to allow OAuth redirect during development, makes the one remaining section clearer and fixes the link it contains to a shopify.dev tutorial.

This gem's docs previously gave instructions to setup a Cloudflared tunnel, but then directed users to follow a section of a shopify.dev tutorial which started with setting up an Ngrok tunnel instead before moving on to adding the redirect URI to the partners dashboard.

After this change, the docs link first to the shopify.dev tutorial section and include a note after for using Cloudflared should the user wish.

### Checklist

### DIFF
--- a/docs/Quickstart.md
+++ b/docs/Quickstart.md
@@ -4,24 +4,31 @@ This guide assumes you have completed the steps to create a new Rails app using 
 
 #### Table of contents
 
-[Make your app available to the internet](#make-your-app-available-to-the-internet)
+[Setup SSH tunnel for development](#setup-ssh-tunnel-for-development)
 
 [Use Shopify App Bridge to embed your app in the Shopify Admin](#use-shopify-app-bridge-to-embed-your-app-in-the-shopify-admin)
 
-## Make your app available to the internet
+## Setup SSH tunnel for development
 
 Your local app needs to be accessible from the public Internet in order to install it on a Shopify store, to use the [App Proxy Controller](/lib/generators/shopify_app/app_proxy_controller/templates/app_proxy_controller.rb) or receive [webhooks](/docs/shopify_app/webhooks.md).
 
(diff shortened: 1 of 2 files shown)


docs/Quickstart.md
```diff
@@ -12,16 +12,16 @@ This guide assumes you have completed the steps to create a new Rails app using
 
 Your local app needs to be accessible from the public Internet in order to install it on a Shopify store, to use the [App Proxy Controller](/lib/generators/shopify_app/app_proxy_controller/templates/app_proxy_controller.rb) or receive [webhooks](/docs/shopify_app/webhooks.md).
 
-Use a tunneling service like [Cloudflare](https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/run-tunnel/trycloudflare/) to make your development environment accessible to the internet.
+Follow the [*Embed the app in Shopify*](https://shopify.dev/tutorials/build-rails-react-app-that-uses-app-bridge-authentication#step-3-embed-the-app-in-shopify) section of [*Build a Shopify app with Rails, React, and App Bridge*](https://shopify.dev/tutorials/build-rails-react-app-that-uses-app-bridge-authentication) to make your development environment accessible to the internet using Ngrok.
```
andyw8, line 15: @nelsonwittwer Should we be intentionally steering people towards Cloudflare rather than Ngrok now?

nelsonwittwer: Yes, we last updated the docs to use Cloudflare instead of Ngrok because Ngrok accounts kept getting blocked from Chrome. From what I could tell OAuth redirects to Ngrok hosts looked like phishing and Google would ban accounts. Cloudfare hasn't had that behavior so we have been pointing people that direction.

thisdotrob: I've updated the changes to list the steps to use Cloudflared and not link to the shopify.dev docs based on the above

nelsonwittwer (approved):
Thanks for taking the time to make the docs better :tada:

PR #1516, opened 2022-09-22
Fix user session storing without scopes

paulomarg:
### What this PR does

We had updated the `UserSessionStorageWithScopes` class to fix an issue where the type of the `user` param was incorrect, but the same issue was present in `UserSessionStorage`. This PR fixes that to properly call `user.id` to access the `AssociatedUser` property.

You'll need to create an app with a few conditions:
- It has User sessions (`rails g shopify_app:user_model`)
  - Answer `n` when prompted if you want to add user scopes
- Make sure the `User` model is using the `include ShopifyApp::UserSessionStorage` concern (not `ShopifyApp::UserSessionStorageWithScopes`
- Set `config.reauth_on_access_scope_changes = false` in the `shopify_app` initializer
- Set `config.user_session_repository = "User"`

### Things to focus on

Before this branch, that would lead into an infinite auth loop. With this fix, the app should authenticate normally, and there should be an entry in both the `users` and `shops` tables.

### Checklist


lilypustovyk (approved):
😍 
Thank you for addressing this!

PR #1518, opened 2022-09-23
use ShopifyAPI::Context.host instead of assume https://

nelsonwittwer:
### Things to focus on

1. 
2. 
3. 

### Checklist

### DIFF
--- a/app/views/shopify_app/sessions/enable_cookies.html.erb
+++ b/app/views/shopify_app/sessions/enable_cookies.html.erb
@@ -24,7 +24,7 @@
       id: 'redirection-target',
       data: {
         target: {
-          myshopifyUrl: "https://#{current_shopify_domain}",
+          myshopifyUrl: current_shopify_domain,
           hasStorageAccessUrl: "#{has_storage_access_url}",
           doesNotHaveStorageAccessUrl: "#{does_not_have_storage_access_url}",
           appTargetUrl: "#{app_target_url}"
--- a/app/views/shopify_app/shared/redirect.html.erb
+++ b/app/views/shopify_app/shared/redirect.html.erb
@@ -13,7 +13,7 @@
     id: 'redirection-target',
     data: {
       target: {
-        myshopifyUrl: "https://#{current_shopify_domain}",
+        myshopifyUrl: current_shopify_domain,
         url: url,
       },
     },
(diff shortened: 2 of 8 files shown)


app/views/shopify_app/sessions/enable_cookies.html.erb
```diff
@@ -24,7 +24,7 @@
       id: 'redirection-target',
       data: {
         target: {
-          myshopifyUrl: "https://#{current_shopify_domain}",
+          myshopifyUrl: "#{ShopifyAPI::Context.host_scheme}://#{current_shopify_domain},
```
paulomarg, line 27: I think when we're pointing to `current_shopify_domain`, we should always be using https - what is http in this scenario is the app only, right?

app/views/shopify_app/shared/redirect.html.erb
```diff
@@ -13,7 +13,7 @@
     id: 'redirection-target',
     data: {
       target: {
-        myshopifyUrl: "https://#{current_shopify_domain}",
+        myshopifyUrl: ShopifyAPI::Context.host_scheme + "://" + current_shopify_domain,
```
andyw8, line 16: Could we have a method like `current_shopify_uri`?

app/views/shopify_app/sessions/enable_cookies.html.erb
nelsonwittwer: that's right! Great call out

andyw8 (approved):

lib/generators/shopify_app/install/templates/shopify_app.rb.tt
```diff
@@ -38,7 +38,7 @@ Rails.application.config.after_initialize do
       api_key: ShopifyApp.configuration.api_key,
       api_secret_key: ShopifyApp.configuration.secret,
       api_version: ShopifyApp.configuration.api_version,
-      host_name: URI(ENV.fetch('HOST', '')).host || '',
+      host: ENV['HOST'],
```
alan-pie, line 41: Either this or https://github.com/Shopify/shopify-api-ruby/pull/1017 is incorrect.  `host_name` is still required by the `setup` function. I'm not sure if it should still be required or should be optional now. If it should be optional maybe there should be some error checking to ensure that host and host_name don't conflict?

Eianaftyio:
.patch

PR #1530, opened 2022-10-11
Avoid overly-general rescue in callback controller

### LINKED ISSUE #1527: Avoid overly-general `rescue` of errors in `CallbackController`
In `CallbackController`, we [rescue from _any_ kind of `StandardError`.](https://github.com/Shopify/shopify_app/blob/6f8080da5e765859091ec133dff0370999917cc1/app/controllers/shopify_app/callback_controller.rb#L20-L22) This can mask coding errors such as calling an undefined method, which may later result in an infinite redirect or other problem.

Instead, we should only rescue from error swhich are _expected_ as part of the OAuth flow, and allow other kinds of errors to bubble up.

andyw8:
Closes #1527

~Still to test, just looking for feedback on the approach.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Unreleased
 * Fixes a bug with `EnsureAuthenticatedLinks` causing deep links to not work [#1549](https://github.com/Shopify/shopify_app/pull/1549)
 * Ensure online token is properly used when using `current_shopify_session` [#1566](https://github.com/Shopify/shopify_app/pull/1566)
 * Emit a deprecation notice for wrongly-rescued exceptions [#1530](https://github.com/Shopify/shopify_app/pull/1530)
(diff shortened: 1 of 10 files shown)


test/controllers/callback_controller_test.rb
```diff
@@ -35,13 +35,13 @@ class CallbackControllerTest < ActionController::TestCase
     end
 
     test "#callback flashes error when omniauth is not present" do
-      get :callback, params: { shop: "shop" }
```
andyw8, line 38: The `rescue` was masking these failing tests.

andyw8:
I plan to change this to emit a deprecation warning instead of raising.

nelsonwittwer (approved):

andyw8:
@nelsonwittwer ready for a final review.

PR #1540, opened 2022-10-19
Pass access scopes on query string

andyw8:
### What this PR does

See explanation in related PR: https://github.com/Shopify/shopify-api-ruby/pull/1023

Use dev version of https://shopify-graphiql-app.shopifycloud.com/

### Checklist

- [x] ~Update `README.md`, if appropriate.~
- [x] ~Update any relevant pages in `/docs`, if necessary~
- [x] ~For security fixes, the [Disclosure Policy](https://github.com/Shopify/shopify_app/blob/master/SECURITY.md#disclosure-policy) must be followed.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,14 @@
 Unreleased
 ----------
 
+* Pass access scopes on query string [#1540](https://github.com/Shopify/shopify_app/pull/1540)
+
 21.1.1 (Oct 20, 2022)
 ----------
 * Updates dependency to `shopify_api` to 12.2 to fix error with host_name argument.
 
 21.1.0 (Oct 17, 2022)
 ----------
-
 * Removes assumed `https` required to run locally. Support both `http` and `https` in backward compatible way. [#1518](https://github.com/Shopify/shopify_app/pull/1518)
 
 21.0.0 (Oct 3, 2022)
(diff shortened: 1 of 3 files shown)


lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -167,6 +167,10 @@ def login_url_params(top_level:)
         query_params[:host] ||= host
       end
 
+      if params[:access_scopes].present?
+        query_params[:scope] = params[:access_scopes].join(",")
```
andyw8, line 171: Other places in this method use `||=` but I don't see a need for that here.

CHANGELOG.md
```diff
@@ -1,13 +1,14 @@
...
 
+* Pass access scopes on query string [#1540](https://github.com/Shopify/shopify_app/pull/1540)
+
 21.1.1 (Oct 20, 2022)
 ----------
 * Updates dependency to `shopify_api` to 12.2 to fix error with host_name argument.
 
 21.1.0 (Oct 17, 2022)
 ----------
-
```
andyw8, line 10: Removed these for consistency.

lib/shopify_app/controller_concerns/login_protection.rb
paulomarg: Makes sense to me, if we get query args we should override them, assuming we can trust the source - is this secured in any way, like with an HMAC?

paulomarg (approved):
Just a general question, but if there is nothing to change then I'm happy with this.

nelsonwittwer (approved):

lib/shopify_app/controller_concerns/login_protection.rb
andyw8: (Discussed internally, we do not see any issue here)

PR #1545, opened 2022-10-21
Adding Debug Logs

klenotiw:
### What this PR does
This is intended to add logs to critical areas of the code to ease the pain of debugging. 

closes https://github.com/Shopify/first-party-library-planning/issues/482

Read through all of the logs. If you are really cool you can test some logs out by using a testing app.

### Things to focus on

1. Does this PR follow our logging approach?
2. Read logging messages to makes sure they sound clear and make sense.
3. Should we add logs anywhere else?
4. Do the logs fall in the correct level?

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ Unreleased
 ----------
 * Fixes a bug with `EnsureAuthenticatedLinks` causing deep links to not work [#1549](https://github.com/Shopify/shopify_app/pull/1549)
 * Ensure online token is properly used when using `current_shopify_session` [#1566](https://github.com/Shopify/shopify_app/pull/1566)
+* Added debug logs, you can read more about logging (here)[./docs/logging.md]. [#1545](https://github.com/Shopify/shopify_app/pull/1545)
 * Emit a deprecation notice for wrongly-rescued exceptions [#1530](https://github.com/Shopify/shopify_app/pull/1530)
 * Log a deprecation warning for the use of incompatible controller concerns [#1560](https://github.com/Shopify/shopify_app/pull/1560)
 
--- a/app/controllers/concerns/shopify_app/ensure_authenticated_links.rb
+++ b/app/controllers/concerns/shopify_app/ensure_authenticated_links.rb
@@ -28,8 +28,8 @@ module ShopifyApp
     def redirect_to_splash_page
       redirect_to(splash_page)
     rescue ::ShopifyApp::ShopifyDomainNotFound => error
-      Rails.logger.warn("[ShopifyApp::EnsureAuthenticatedLinks] Redirecting to login: [#{error.class}] "\
-        "Could not determine current shop domain")
+      ShopifyApp::Logger.warn("Redirecting to login: [#{error.class}]"\
+        " Could not determine current shop domain")
       redirect_to(ShopifyApp.configuration.login_url)
     end
 
--- a/app/controllers/shopify_app/sessions_controller.rb
+++ b/app/controllers/shopify_app/sessions_controller.rb
@@ -27,6 +27,8 @@ module ShopifyApp
     def destroy
       reset_session
       flash[:notice] = I18n.t(".logged_out")
+      ShopifyApp::Logger.debug("Session destroyed")
+      ShopifyApp::Logger.debug("Redirecting to #{login_url_with_optional_shop}")
       redirect_to(login_url_with_optional_shop)
     end
 
@@ -38,6 +40,7 @@ module ShopifyApp
       copy_return_to_param_to_session
 
       if embedded_redirect_url?
+        ShopifyApp::Logger.debug("Embedded URL within / authenticate")
         if embedded_param?
           redirect_for_embedded
         else
@@ -52,6 +55,7 @@ module ShopifyApp
 
     def start_oauth
       callback_url = ShopifyApp.configuration.login_callback_url.gsub(%r{^/}, "")
+      ShopifyApp::Logger.debug("Starting OAuth with the following callback URL: #{callback_url}")
 
       auth_attributes = ShopifyAPI::Auth::Oauth.begin_auth(
         shop: sanitized_shop_name,
@@ -65,7 +69,10 @@ module ShopifyApp
         value: auth_attributes[:cookie].value,
       }
 
-      redirect_to(auth_attributes[:auth_route], allow_other_host: true)
+      auth_route = auth_attributes[:auth_route]
+
+      ShopifyApp::Logger.debug("Redirecting to auth_route - #{auth_route}")
+      redirect_to(auth_route, allow_other_host: true)
     end
 
     def validate_shop_presence
@@ -94,7 +101,9 @@ module ShopifyApp
     end
 
     def redirect_auth_to_top_level
-      fullpage_redirect_to(login_url_with_optional_shop(top_level: true))
+      url = login_url_with_optional_shop(top_level: true)
+      ShopifyApp::Logger.debug("Redirecting to top level - #{url}")
+      fullpage_redirect_to(url)
     end
   end
 end
--- a/docs/shopify_app/logging.md
+++ b/docs/shopify_app/logging.md
@@ -1,0 +1,21 @@
+# Logging
+
+## Log Levels
+
+There are four log levels with `error` being the most severe.
+
+1. Debug
+2. Info
+3. Warn
+4. Error
+
+## Configuration
+
+The logging is controlled by the `log_level` configuration setting.
+The default log level is `:info`.
+You can disable all logs by setting this to `:off`.
+
+## Upgrading
+
+For a newly-generated app, the `shopify_app` initializer will contain the `log_level` setting.
+If you are upgrading from a previous version of the `shopify_app` gem then you will need to add this manually, otherwise it will default to `:info`.
--- a/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
+++ b/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
@@ -7,7 +7,7 @@ ShopifyApp.configure do |config|
   config.after_authenticate_job = false
   config.api_version = "<%= @api_version %>"
   config.shop_session_repository = 'Shop'
-
+  config.log_level = :info
   config.reauth_on_access_scope_changes = true
 
   config.api_key = ENV.fetch('SHOPIFY_API_KEY', '').presence
--- a/lib/shopify_app/controller_concerns/ensure_billing.rb
+++ b/lib/shopify_app/controller_concerns/ensure_billing.rb
@@ -28,6 +28,7 @@ module ShopifyApp
       unless has_payment
         if request.xhr?
           add_top_level_redirection_headers(url: confirmation_url, ignore_response_code: true)
+          ShopifyApp::Logger.debug("Responding with 401 unauthorized")
           head(:unauthorized)
         else
           redirect_to(confirmation_url, allow_other_host: true)
@@ -55,6 +56,7 @@ module ShopifyApp
     end
 
     def has_subscription?(session)
+      ShopifyApp::Logger.debug("Checking if shop has subscription")
       response = run_query(session: session, query: RECURRING_PURCHASES_QUERY)
       subscriptions = response.body["data"]["currentAppInstallation"]["activeSubscriptions"]
 
@@ -70,6 +72,7 @@ module ShopifyApp
     end
 
     def has_one_time_payment?(session)
+      ShopifyApp::Logger.debug("Checking if has one time payment")
       purchases = nil
       end_cursor = nil
 
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -25,6 +25,7 @@ module ShopifyApp
     def activate_shopify_session
       if current_shopify_session.blank?
         signal_access_token_required
+        ShopifyApp::Logger.debug("No session found, redirecting to login")
         return redirect_to_login
       end
 
@@ -36,9 +37,11 @@ module ShopifyApp
       end
 
       begin
+        ShopifyApp::Logger.debug("Activating Shopify session")
         ShopifyAPI::Context.activate_session(current_shopify_session)
         yield
       ensure
+        ShopifyApp::Logger.debug("Deactivating session")
         ShopifyAPI::Context.deactivate_session
       end
     end
@@ -52,8 +55,10 @@ module ShopifyApp
           is_online: online_token_configured?,
         )
       rescue ShopifyAPI::Errors::CookieNotFoundError
+        ShopifyApp::Logger.info("No cookies have been found - cookie name: #{cookie_name}")
         nil
       rescue ShopifyAPI::Errors::InvalidJwtTokenError
+        ShopifyApp::Logger.info("Invalid JWT token for current Shopify session")
         nil
       end
     end
@@ -61,6 +66,7 @@ module ShopifyApp
     def login_again_if_different_user_or_shop
       return unless session_id_conflicts_with_params || session_shop_conflicts_with_params
 
+      ShopifyApp::Logger.debug("Clearing session and redirecting to login")
       clear_shopify_session
       redirect_to_login
     end
@@ -78,8 +84,10 @@ module ShopifyApp
 
     def add_top_level_redirection_headers(url: nil, ignore_response_code: false)
       if request.xhr? && (ignore_response_code || response.code.to_i == 401)
+        ShopifyApp::Logger.debug("Adding top level redirection headers")
         # Make sure the shop is set in the redirection URL
         unless params[:shop]
+          ShopifyApp::Logger.debug("Setting current shop session")
           params[:shop] = if current_shopify_session
             current_shopify_session.shop
           elsif (matches = request.headers["HTTP_AUTHORIZATION"]&.match(/^Bearer (.+)$/))
@@ -90,6 +98,7 @@ module ShopifyApp
 
         url ||= login_url_with_optional_shop
 
+        ShopifyApp::Logger.debug("Setting Reauthorize-Url to #{url}")
         response.set_header("X-Shopify-API-Request-Failure-Reauthorize", "1")
         response.set_header("X-Shopify-API-Request-Failure-Reauthorize-Url", url)
       end
@@ -112,6 +121,7 @@ module ShopifyApp
     def redirect_to_login
       if request.xhr?
         add_top_level_redirection_headers(ignore_response_code: true)
+        ShopifyApp::Logger.debug("Login redirect request is a XHR")
         head(:unauthorized)
       else
         if request.get?
@@ -124,12 +134,15 @@ module ShopifyApp
           query = query.merge(sanitized_params).to_query
         end
         session[:return_to] = query.blank? ? path.to_s : "#{path}?#{query}"
+        ShopifyApp::Logger.debug("Redirecting to #{login_url_with_optional_shop}")
         redirect_to(login_url_with_optional_shop)
       end
     end
 
     def close_session
       clear_shopify_session
+      ShopifyApp::Logger.debug("Closing session")
+      ShopifyApp::Logger.debug("Redirecting to #{login_url_with_optional_shop}")
       redirect_to(login_url_with_optional_shop)
     end
 
--- a/lib/shopify_app/controller_concerns/redirect_for_embedded.rb
+++ b/lib/shopify_app/controller_concerns/redirect_for_embedded.rb
@@ -16,7 +16,10 @@ module ShopifyApp
 
     def redirect_for_embedded
       # Don't actually redirect if we're already in the redirect route - we want the request to reach the FE
-      redirect_to(redirect_uri_for_embedded) unless request.path == ShopifyApp.configuration.embedded_redirect_url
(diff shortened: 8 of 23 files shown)


lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -44,9 +47,12 @@ def current_shopify_session
           cookies: { cookie_name => cookies.encrypted[cookie_name] },
           is_online: user_session_expected?,
         )
+        # Rails.logger.debug("ShopifyApp - Loaded current shopify session")
```
klenotiw, line 50: This was causing an error when loading the products. Not really sure how that relates to a log message.

```
ShopifyApp - Loaded current shopify session
Completed 500 Internal Server Error in 10ms (ActiveRecord: 1.9ms | Allocations: 5484)


  
NoMethodError (undefined method `shopify_session_id' for 44:Integer):
  
/Users/klenotiw/src/github.com/Shopify/shopify_app/lib/shopify_app/controller_concerns/login_protection.rb:223:in `session_id_conflicts_with_params'
/Users/klenotiw/src/github.com/Shopify/shopify_app/lib/shopify_app/controller_concerns/login_protection.rb:58:in `login_again_if_different_user_or_shop'
activesupport (7.0.3.1) lib/active_support/callbacks.rb:400:in `block in make_lambda'
```

klenotiw:
I probably missed somethings but Im still trying to read through the gem and see if there is anything that jumps out to me. Feel free to make comments.

lib/shopify_app/managers/webhooks_manager.rb
```diff
@@ -15,6 +15,8 @@ def queue(shop_domain, shop_token)
       def create_webhooks(session:)
         return unless ShopifyApp.configuration.has_webhooks?
 
+        Rails.logger.debug("ShopifyApp - Creating Webhooks #{ShopifyApp.configuration.webhooks}")
```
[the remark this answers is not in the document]
nelsonwittwer: That's great as is, thanks for that example! I've just been bit my logging hashes in the past

klenotiw:
@nelsonwittwer I just realized with how we have it set up if someone changes the log levels in their app doing something like this:
``` ruby
config.log_level = :warn
``` 

This would not only remove our logs but other logs they have. Should we make something where they can change the log levels of the shopify_app gem specifically or is this not a big deal?

app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -27,6 +27,7 @@ def top_level_interaction
     def destroy
       reset_session
       flash[:notice] = I18n.t(".logged_out")
+      Rails.logger.debug("ShopifyApp - Destroy Session and redirecting to login")
```
andyw8, line 30: I wonder if we can use Tagged Logging so we don't need to repeat `ShopifyApp` everywhere: https://api.rubyonrails.org/classes/ActiveSupport/TaggedLogging.html

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -45,8 +49,10 @@ def current_shopify_session
           is_online: user_session_expected?,
         )
       rescue ShopifyAPI::Errors::CookieNotFoundError
+        ShopifyApp::Utils::Logger.warn("CookiesNotFound for current shopify session")
         nil
       rescue ShopifyAPI::Errors::InvalidJwtTokenError
+        ShopifyApp::Utils::Logger.warn("Invalid Jwt token for current shopify session")
```
klenotiw, line 55: Should rescued errors be warns?

lib/shopify_app/session/jwt.rb
```diff
@@ -35,7 +35,7 @@ def set_payload
       payload, _ = parse_token_data(ShopifyApp.configuration&.secret, ShopifyApp.configuration&.old_secret)
       @payload = validate_payload(payload)
     rescue *WARN_EXCEPTIONS => error
-      Rails.logger.warn("[ShopifyApp::JWT] Failed to validate JWT: [#{error.class}] #{error}")
+      ShopifyApp::Utils::Logger.warn("[ShopifyApp::JWT] Failed to validate JWT: [#{error.class}] #{error}")
```
klenotiw, line 38: This is an example where I just found a Rails logger output, so I figured I would change it to our new one.

lib/shopify_app/controller_concerns/redirect_for_embedded.rb
```diff
@@ -16,7 +16,10 @@ def embedded_param?
 
     def redirect_for_embedded
       # Don't actually redirect if we're already in the redirect route - we want the request to reach the FE
-      redirect_to(redirect_uri_for_embedded) unless request.path == ShopifyApp.configuration.embedded_redirect_url
+      unless request.path == ShopifyApp.configuration.embedded_redirect_url
+        ShopifyApp::Utils::Logger.debug("Redirecting to #{redirect_uri_for_embedded}")
+        redirect_to(redirect_uri_for_embedded)
+      end
```
klenotiw, line 22: I kind of started going a little wild and logging every redirect, then stopped and figured I would see what other people think. Should I just log every redirect? Does it matter if the redirect is in a private method?

klenotiw:
Hopefully this isn't too rough but lmk what you guys think. Rubocop is working fine on my machine :P i'll figure this out later.

docs/shopify_app/logging.md
```diff
@@ -0,0 +1,8 @@
+# Log Levels
```
andyw8, line 1:
```suggestion
# Logging

## Log Levels
```

lib/shopify_app/utils.rb
```diff
@@ -27,5 +27,41 @@ def self.shop_login_url(shop:, host:, return_to:)
 
       url.to_s
     end
+
+    module Logger
```
andyw8, line 31: I think 'Utils' is not a helpful grouping so I would suggest moving this up a level to `ShopifyApp::Logger`.

docs/shopify_app/logging.md
```diff
@@ -0,0 +1,8 @@
...
+
+1. Debug
+2. Info
+3. Warn
+4. Error
+
+We have four log levels with `error` being the most strict. You can configure your log level by changing your `SHOPIFY_LOG_LEVEL` environment variable. If you don't have a `SHOPIFY_LOG_LEVEL` set your `shopify_app.rb` configuration file with default to `info`. To turn off all shopify_app logs you can change this environment variable to `off`.
```
andyw8, line 10:
```suggestion
We have four log levels with `error` being the most severe. You can configure your log level by changing your `SHOPIFY_LOG_LEVEL` environment variable. If you don't have a `SHOPIFY_LOG_LEVEL` set your `shopify_app.rb` configuration file with default to `info`. To turn off all shopify_app logs you can change this environment variable to `off`.
```

andyw8: What takes precedence if both `SHOPIFY_LOG_LEVEL` and the `shopify_app.rb` configs are set?

lib/shopify_app/session/session_repository.rb
```diff
@@ -66,8 +71,10 @@ def delete_session(id)
         match = id.match(/^offline_(.*)/)
 
         record = if match
+          ShopifyApp::Utils::Logger.debug("Destroying Session by domain - domain: #{match[1]}")
           Shop.find_by(shopify_domain: match[1])
         else
+          ShopifyApp::Utils::Logger.debug("Destroying Session by user - user_id: #{id.split("_").last}")
```
andyw8, line 77: I would avoid doing any manipulation in the log calls, since it could later get out of sync with the implementation. Instead we could extract `id.split("_").last` to a variable and use that in both places.

lib/shopify_app/utils.rb
```diff
@@ -27,5 +27,41 @@ def self.shop_login_url(shop:, host:, return_to:)
...
+      end
+
+      def self.error(message)
+        send_to_logger(:error, message)
+      end
+
+      private
+
+      def enabled_for_log_level?(log_level)
+        LOG_LEVELS[log_level] <= LOG_LEVELS[ShopifyApp.configuration.log_level]
```
andyw8, line 63: Will this work if the log_level is set to `"off"`?

lib/shopify_app/controller_concerns/login_protection.rb
andyw8: Since these are 'handled' errors I would treat them as `info` or `debug`.

andyw8:
Also, I wonder if we should make a 'quiet' release first, and enable it on some internal apps, before we publicize this in the README?

app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -38,12 +40,15 @@ def authenticate
       copy_return_to_param_to_session
 
       if embedded_redirect_url?
+        ShopifyApp::Utils::Logger.debug("Embedded URL within / authenticate")
         if embedded_param?
+          ShopifyApp::Utils::Logger.debug("Embedded param")
```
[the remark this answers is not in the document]
klenotiw: Oh I think this is telling the session_controller if the app is embedded or not? At one point we wanted to log all of the conditionals in the `authenticate` method. Do you think we should just remove these logs?

test/test_helper.rb
```diff
@@ -47,6 +47,7 @@ def before_setup
     ShopifyApp::InMemorySessionStore.clear
     ShopifyAppConfigurer.call
     Rails.application.reload_routes!
+    ShopifyApp.configuration.log_level = :warn
```
klenotiw, line 50: I had to set a log_level for our tests. I decided on `warn` instead of off because some tests are expecting to see some logs come through. All of those tests are rescued errors so I figured `:warn` was the best level. LMK what you think

lib/generators/shopify_app/install/templates/shopify_app.rb.tt
```diff
@@ -13,6 +13,8 @@ ShopifyApp.configure do |config|
   config.api_key = ENV.fetch('SHOPIFY_API_KEY', '').presence
   config.secret = ENV.fetch('SHOPIFY_API_SECRET', '').presence
 
+  config.log_level = ENV["SHOPIFY_LOG_LEVEL"]&.to_sym || :info
```
klenotiw, line 16: Should we just do something like
`config.log_level = :info` 
and let them edit the `shopify_app.rb` file directly?

lib/shopify_app/logger.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module Logger
+    LOG_LEVELS = { debug: 0, info: 1, warn: 2, error: 3, off: 4 }
```
andyw8, line 5: I think it would good to match the Rails level numbers:

https://guides.rubyonrails.org/debugging_rails_applications.html#log-levels

I'm a bit unsure about having 'off' as a level, that feels like overloading this setting. I'm looking into how else we might handle that.

lib/generators/shopify_app/install/templates/shopify_app.rb.tt
andyw8: Yeah, that seems better, maybe we don't need the `SHOPIFY_LOG_LEVEL` setting at all.

lib/shopify_app/logger.rb
```diff
@@ -0,0 +1,37 @@
...
+    def self.warn(message)
+      send_to_logger(:warn, message)
+    end
+
+    def self.error(message)
+      send_to_logger(:error, message)
+    end
+
+    def self.enabled_for_log_level?(log_level)
+      LOG_LEVELS[log_level] >= LOG_LEVELS[ShopifyApp.configuration.log_level]
```
andyw8, line 34: If the `shopify_app.rb` is missing the `log_level` setting (e.g. for someone upgrading) then this fails with:

```
Completed 500 Internal Server Error in 1ms (ActiveRecord: 0.0ms | Allocations: 1214)


  
ArgumentError (comparison of Integer with nil failed):
  
/Users/andyw8/src/github.com/Shopify/shopify_app/lib/shopify_app/logger.rb:38:in `>='
/Users/andyw8/src/github.com/Shopify/shopify_app/lib/shopify_app/logger.rb:38:in `enabled_for_log_level?'
/Users/andyw8/src/github.com/Shopify/shopify_app/lib/shopify_app/logger.rb:9:in `send_to_logger'
/Users/andyw8/src/github.com/Shopify/shopify_app/lib/shopify_app/logger.rb:18:i
```

lib/shopify_app/logger.rb
klenotiw: Is this not following the Rails level number, just not including `fatal` and `unknown`? Yeah thats just what popped in my head. We could make a separate variable in the config? Or we could just manually check for `:off` in the `enabled_for_log_level?` method?

docs/shopify_app/logging.md
```diff
@@ -0,0 +1,18 @@
...
+
+We have four log levels with `error` being the most severe.
+
+## Configuration
+
+Your log level is set to `:info` by default. You can change this by going into your configuration file and changing the `log_level` setting. You can remove all logs by setting this to `:off`.
+
+## Upgrading
+
+Make sure to add the `config.log_level` setting to your configuration file so you can start changing the log level to your preference. If `log_level` isn't found if your configuration settings then it will default to the `:info` level.
```
klenotiw, line 18: Also added this upgrading section. (wasn't sure if this is the best place)

lib/shopify_app/logger.rb
andyw8: My concern was the `4` here means `off`, but `4` for Rails means 'fatal`. We could map `off` to 6 to avoid any confusion there.

lib/shopify_app/controller_concerns/ensure_billing.rb
```diff
@@ -55,6 +56,7 @@ def has_active_payment?(session)
     end
 
     def has_subscription?(session)
+      ShopifyApp::Logger.debug("Checking if has subscription")
```
teddyhwang, line 59: checking if what has subscription?

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -16,8 +16,11 @@ module LoginProtection
     ACCESS_TOKEN_REQUIRED_HEADER = "X-Shopify-API-Request-Failure-Unauthorized"
 
     def activate_shopify_session
+      ShopifyApp::Logger.debug("Activating Shopify Session")
+
       if current_shopify_session.blank?
         signal_access_token_required
+        ShopifyApp::Logger.debug("No session found")
```
teddyhwang, line 23:
```suggestion
        ShopifyApp::Logger.debug("No session found, redirecting to login")
```

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -32,6 +35,7 @@ def activate_shopify_session
         ShopifyAPI::Context.activate_session(current_shopify_session)
         yield
       ensure
+        ShopifyApp::Logger.debug("Deactivating Session")
```
teddyhwang, line 38: How come for some logs we capitalize each word like it's a title? Is this a format that we're doing in the node library as well /cc @paulomarg

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -45,15 +49,18 @@ def current_shopify_session
           is_online: online_token_configured?,
         )
       rescue ShopifyAPI::Errors::CookieNotFoundError
+        ShopifyApp::Logger.info("CookiesNotFound for current shopify session")
         nil
       rescue ShopifyAPI::Errors::InvalidJwtTokenError
+        ShopifyApp::Logger.info("Invalid Jwt token for current shopify session")
```
teddyhwang, line 55:
```suggestion
        ShopifyApp::Logger.info("Invalid JWT token for current shopify session")
```

teddyhwang (commented):
I'm not sure if it's necessary but it would be nice to have the same log language as we do across the libraries so that we can have the same tone and content

lib/shopify_app/controller_concerns/login_protection.rb
paulomarg: No, we've used sentence case pretty much everywhere for node!

klenotiw: ok just went through and tried to clean that up

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@ Unreleased
 ----------
 * Fixes a bug with `EnsureAuthenticatedLinks` causing deep links to not work [#1549](https://github.com/Shopify/shopify_app/pull/1549)
 * Ensure online token is properly used when using `current_shopify_session` [#1566](https://github.com/Shopify/shopify_app/pull/1566)
-* Added Debug Logs [#1545](https://github.com/Shopify/shopify_app/pull/1545)
+* Added debug logs, you can read more about our logging strategy (here)[https://github.com/Shopify/shopify_app/blob/main/docs/logging.md]. [#1545](https://github.com/Shopify/shopify_app/pull/1545)
```
paulomarg, line 5: I'm wondering if this is the right approach - I would argue that we should document the logger settings that we make available, but not necessarily the strategy behind it (as in, do developers care _why_ as opposed to _how_ here?). No strong opinion here!

BTW this should probably be a relative link (untested):
```suggestion
* Added debug logs, you can read more about our logging strategy (here)[./docs/logging.md]. [#1545](https://github.com/Shopify/shopify_app/pull/1545)
```

app/controllers/concerns/shopify_app/ensure_authenticated_links.rb
```diff
@@ -28,8 +28,8 @@ def splash_page_with_params(params)
     def redirect_to_splash_page
       redirect_to(splash_page)
     rescue ::ShopifyApp::ShopifyDomainNotFound => error
-      Rails.logger.warn("[ShopifyApp::EnsureAuthenticatedLinks] Redirecting to login: [#{error.class}] "\
-        "Could not determine current shop domain")
+      ShopifyApp::Logger.warn("[ShopifyApp::EnsureAuthenticatedLinks] Redirecting to login: [#{error.class}]"\
```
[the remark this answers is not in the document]
andyw8: It's only this way because this was an existing log entry. I think including the class/module name everytime is a little too noisy - perhaps we only include it if there's some similar code paths and we need to differentiate them?

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -45,15 +49,18 @@ def current_shopify_session
           is_online: online_token_configured?,
         )
       rescue ShopifyAPI::Errors::CookieNotFoundError
+        ShopifyApp::Logger.info("CookiesNotFound for current Shopify session")
         nil
       rescue ShopifyAPI::Errors::InvalidJwtTokenError
+        ShopifyApp::Logger.info("Invalid JWT token for current shopify session")
```
andyw8, line 55:
```suggestion
        ShopifyApp::Logger.info("Invalid JWT token for current Shopify session")
```
(same for any other places where we mention Shopify).

lib/shopify_app/session/jwt.rb
```diff
@@ -35,7 +35,7 @@ def set_payload
       payload, _ = parse_token_data(ShopifyApp.configuration&.secret, ShopifyApp.configuration&.old_secret)
       @payload = validate_payload(payload)
     rescue *WARN_EXCEPTIONS => error
-      Rails.logger.warn("[ShopifyApp::JWT] Failed to validate JWT: [#{error.class}] #{error}")
+      ShopifyApp::Logger.warn("[ShopifyApp::JWT] Failed to validate JWT: [#{error.class}] #{error}")
```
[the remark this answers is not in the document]
klenotiw: That log was already there and is apart of a test, so I just left it how it was but Ill just remove that now.

lib/shopify_app/session/session_repository.rb
```diff
@@ -45,8 +45,12 @@ def user_storage
       # ShopifyAPI::Auth::SessionStorage override
       def store_session(session)
         if session.online?
-          user_storage.store(session, session.associated_user)
+          user = session.associated_user
+          ShopifyApp::Logger.debug("Storing online user session - session: #{session},"\
```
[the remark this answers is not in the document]
andyw8: We should probably avoid logging any PII, including the email.

app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -38,12 +40,15 @@ def authenticate
       copy_return_to_param_to_session
 
       if embedded_redirect_url?
+        ShopifyApp::Logger.debug("Embedded URL within / authenticate")
         if embedded_param?
+          ShopifyApp::Logger.debug("Embedded param")
```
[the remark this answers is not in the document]
klenotiw: I think we can just remove this since we already have a log in `redirect_for_embedded`. 

I think the original idea was to separate out all of the conditionals in the `authenticate` method.

lib/shopify_app/session/session_repository.rb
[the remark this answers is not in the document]
andyw8: Session ID is fine.

lib/shopify_app/controller_concerns/ensure_billing.rb
klenotiw: Yeah as far as I can tell thats whats happening, "Checking if shop has subscription" would probably be more clear.

nelsonwittwer (approved):

PR #1563, opened 2022-11-02
Move ownership for session persistence from library to this gem

nelsonwittwer:
### What this PR does

Exploring the idea of moving session persistence calls to this gem instead of passing `ShopifyApp::SessionRepository` to the API library and having the API library be responsible for persisting the session. Having the API library invoke Rails' persistence is backward from a responsibility perspective.

The goal with this proposal is to use the API library that already returns a [session payload](https://github.com/Shopify/shopify-api-ruby/blob/main/lib/shopify_api/auth/oauth.rb#L101) to the `shopify_app` and use that payload to persist and load session data using the `ShopifyApp::SessionRepository` exclusively and remove any concept of session persistence from the API library. 

### TODO

- [x] Should we copy session storage errors from API library over?
- [x] Should we keep session ID deconstruction helpers in the API library? 
- [x] Can we remove `ShopifyAPI::Auth::SessionStorage` from `SessionRepository` now?

This is a risky change in that all API calls and authentication code has been modified. We'll want to perform a full regression of the features of the gem and ensure we can make rest / graphql calls with the new session interface.

### Checklist

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -29,3 +29,4 @@ jobs:
           days-before-pr-close: -1
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           exempt-issue-labels: "feature request"
+          close-issue-reason: "not_planned"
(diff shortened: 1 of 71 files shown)


klenotiw (approved):
Looks good to me. Just to be sure I didn't miss out somewhere, I tested this by generating a new app. Then I installed it to a store and saw products. This should be enough right?

Issue #1502, opened 2022-09-09
Consider renaming `RequireKnownShop` and `Authenticated` modules


andyw8:
As per a suggestion by @paulomarg, we might want give these clearer names, e.g.:

* RequireKnownShop => EnsureInstalled
* Authenticated => EnsureHasSession

But we should retain the old names so as not to break existing apps.

cc @mkevinosullivan

PR #1565, opened 2022-11-03
Renaming some Controller Concerns

klenotiw:
### What this PR does

We had two controllers concerns that we wanted to rename, but still have our old names be backwards compatible for the time being.

RequireKnownShop => EnsureInstalled
Authenticated => EnsureHasSession

I merged in the commit from[ the logging PR ](https://github.com/Shopify/shopify_app/pull/1545)because I wanted to add some deprecation logs. So sorry for the messed up commit history.

closes https://github.com/Shopify/shopify_app/issues/1502

### Things to focus on

1. Do we want to rename any other concerns?
2. Are we happy with the new names?
3. Are we happy with the log formatting?

### Checklist

### DIFF
--- a/app/controllers/concerns/shopify_app/authenticated.rb
+++ b/app/controllers/concerns/shopify_app/authenticated.rb
@@ -2,11 +2,12 @@
 
 module ShopifyApp
   module Authenticated
-    extend ShopifyApp::EnsureHasSession
     extend ActiveSupport::Concern
 
     included do
       ShopifyApp::Logger.deprecated("Authenticated has been renamed to EnsureHasSession")
     end
+
+    include ShopifyApp::EnsureHasSession
   end
 end
--- a/app/controllers/concerns/shopify_app/require_known_shop.rb
+++ b/app/controllers/concerns/shopify_app/require_known_shop.rb
@@ -3,47 +3,11 @@
 module ShopifyApp
   module RequireKnownShop
     extend ActiveSupport::Concern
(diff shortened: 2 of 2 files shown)


lib/shopify_app/utils.rb
```diff
@@ -27,5 +27,45 @@ def self.shop_login_url(shop:, host:, return_to:)
...
+      def self.warn(message)
+        send_to_logger(:warn, message)
+      end
+
+      def self.error(message)
+        send_to_logger(:error, message)
+      end
+
+      def self.deprecated(message)
+        send_to_logger(:warn, "DEPRECATED - #{message}")
```
andyw8, line 61: I think we should be using the ActiveSupport Deprecation API for this:
https://andycroll.com/ruby/use-a-deprecation-message/

app/controllers/concerns/shopify_app/ensure_has_session.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module ShopifyApp
+  module EnsureHasSession
```
andyw8, line 4: Rather than duplicating this, I think inheriting would be preferrable for maintenance. But I'm unsure which way round, `EnsureHasSession < Authenticated` or `Authenticated < EnsureHasSession`...

andyw8: Ignore, I see that's what we're doing already.

PR #1479, opened 2022-07-18
Tests to ensure `current_shopify_session` validates session token using old and new secrets

tatsuya:
### What this PR does

Related #1459 and https://github.com/Shopify/shopify-api-ruby/issues/978

Adds tests to ensure that API key rotation #1459 is supported by passing both old and new secrets in the unit tests. This PR itself has no logic changes as the issue itself was addressed in `shopify-api-ruby` repo (`shopify_api` gem) by https://github.com/Shopify/shopify-api-ruby/issues/978.

### Reviewer's guide to testing

Run `bundle exec rake test TEST=test/shopify_app/controller_concerns/login_protection_test.rb`.

### Things to focus on

Any concern on adding `ShopifyAPI::Context.setup` call in the unit tests? I believe passing a *real* JWT through `ShopifyAPI::Utils::SessionUtils.load_current_session` gives us extra confidence when upgrading `shopify_api` gem in the future.

### Checklist

Before submitting the PR, please consider if any of the following are needed:

- [x] Update `CHANGELOG.md` if the changes would impact users - No changelog is needed as it only adds test cases
- [x] Update `README.md`, if appropriate.
- [x] Update any relevant pages in `/docs`, if necessary
- [x] For security fixes, the [Disclosure Policy](https://github.com/Shopify/shopify_app/blob/master/SECURITY.md#disclosure-policy) must be followed.

PR #1566, opened 2022-11-03
use online tokens when available

teddyhwang:
### What this PR does

Related to https://github.com/Shopify/shopify-graphiql-app/issues/1069

There is currently logic on when to retrieve an online or offline session token where `current_shopify_session` in the `LoginProtection` concern makes an incorrect assumption on checking if a shop session exists when determining to pass in `is_online` to load the session from the API gem. This check is required for the original OAuth request but any subsequent requests after OAuth should just check if a User session is configured.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 Unreleased
 ----------
 * Fixes a bug with `EnsureAuthenticatedLinks` causing deep links to not work [#1549](https://github.com/Shopify/shopify_app/pull/1549)
+* Ensure online token is properly used when using `current_shopify_session` [#1566](https://github.com/Shopify/shopify_app/pull/1566)
 
 21.2.0 (Oct 25, 2022)
 ----------
--- a/test/shopify_app/controller_concerns/login_protection_test.rb
+++ b/test/shopify_app/controller_concerns/login_protection_test.rb
@@ -79,10 +79,6 @@ class LoginProtectionControllerTest < ActionController::TestCase
   end
 
   test "#current_shopify_session loads online session if user session expected" do
-    ShopifyApp::SessionRepository.shop_storage.stubs(:retrieve_by_shopify_domain)
-      .with(@shop)
-      .returns(mock_session(shop: @shop))
-
     request.headers["HTTP_AUTHORIZATION"] = "Bearer token"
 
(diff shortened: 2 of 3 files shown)


test/shopify_app/controller_concerns/login_protection_test.rb
```diff
@@ -116,42 +112,6 @@ class LoginProtectionControllerTest < ActionController::TestCase
     end
   end
 
-  test "#current_shopify_session loads offline session if token is signed with new secret" do
```
teddyhwang, line 119: @tatsuya I noticed you added this test - can you clarify what you thought this test was trying to cover since `secret` and `old_secret` in the configuration object does not change the behavior of whether an online or offline token is used

paulomarg (approved):
Makes sense to me!

nelsonwittwer (approved):
Glad we chased this down! 🎉

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 Unreleased
 ----------
+* Ensure online tokens is properly used when using `current_shopify_session` [#1566](https://github.com/Shopify/shopify_app/pull/1566)
```
andyw8, line 3:
```suggestion
* Ensure online token is properly used when using `current_shopify_session` [#1566](https://github.com/Shopify/shopify_app/pull/1566)
```

andyw8 (approved):

test/shopify_app/controller_concerns/login_protection_test.rb
tatsuya: @teddyhwang The intention here was to ensure that `LoginProtection#current_shopify_session` works properly with session tokens signed with real `secret` and `old_secret`. It didn't mean to test the differences between online or offline tokens.

In the past, this method was returning a `nil` when a token was signed with an old secret. After [the fix](https://github.com/Shopify/shopify-api-ruby/issues/978) was made in the shopify_api lib, that was no longer the case though, I thought having a test case like this will help to ensure that future upgrades to shopify_app lib do not break the current behaviour. Full context https://github.com/Shopify/shopify_app/pull/1479

tatsuya: Having said that, no strong opinion for it. Feel free to keep it or remove it 👍

Issue #1536, opened 2022-10-16
Un-rescuable `ShopifyApp::LoginProtection::ShopifyDomainNotFound` error every time there is an invalid JWT token error


dansundy:
### Description

After upgrading to version `20.1.1` we are seeing a `ShopifyApp::LoginProtection::ShopifyDomainNotFound` error every time there is an invalid JWT token error: `ShopifyAPI::Errors::InvalidJwtTokenError`. The former error can't seem to be rescued or handled in any way and the request returns a 500 level error even if the latter 401 is handled and returned to the front end. I've sifted through the Shopify App gem's login protection concern without seeing any way to avoid this with my configuration.

### Steps to Reproduce

Send a request with an invalid (e.g. expired) JWT token.

**Expected behavior:**

The server returns a 401 level error to the front end.

**Actual behavior:**

The server return a 500 level error to the front end.

**Reproduces how often:**

The `ShopifyApp::LoginProtection::ShopifyDomainNotFound` error happens every time a `ShopifyAPI::Errors::InvalidJwtTokenError` happens. I can't figure out if it is happening after or simultaneously somehow.

### Browsers

- Firefox

### Gem versions

- Shopify app gem `v20.1.1` in production
- Also tested locally with `v20.2.0`

### Additional Information

Stack trace for `ShopifyApp::LoginProtection::ShopifyDomainNotFound`:

```
Stack trace[Hide framework code]
ShopifyApp::LoginProtection::ShopifyDomainNotFound: ShopifyApp::LoginProtection::ShopifyDomainNotFound
…ib/shopify_app/controller_concerns/login_protection.rb:  197:in `current_shopify_domain'
…lib/shopify_app/controller_concerns/frame_ancestors.rb:   10:in `block (3 levels) in <module
….1/lib/action_dispatch/http/content_security_policy.rb:  341:in `instance_exec'
….1/lib/action_dispatch/http/content_security_policy.rb:  341:in `resolve_source'
….1/lib/action_dispatch/http/content_security_policy.rb:  328:in `block in build_directive'
….1/lib/action_dispatch/http/content_security_policy.rb:  328:in `map'
….1/lib/action_dispatch/http/content_security_policy.rb:  328:in `build_directive'
….1/lib/action_dispatch/http/content_security_policy.rb:  317:in `block in build_directives'
….1/lib/action_dispatch/http/content_security_policy.rb:  312:in `each'
[...]

PR #1580, opened 2022-11-15
Allow for `current_shopify_domain` to be nil with authenticated requests

nelsonwittwer:
### What this PR does

There is a bug [with expired tokens](https://github.com/Shopify/shopify_app/issues/1536) that causes our CSP header to throw an error because `current_shopify_domain` isn't able to be retrieved because there is no current active session (it's expired).

Because `LoginProtection#current_shopify_domain` previously has been throwing and error when the domain couldn't be found from the session, this was throwing and exception before it could return a `401 - unauthorized` response.

#### `LoginProtection#current_shopify_domain`

This changes a few things that I'll want to make sure are acceptable design changes:

1. `current_shopify_domain` is allowed to be nil within `LoginProtection`
2. When `current_shopify_domain` is required for `302 - redirect`s in `LoginProtection` we will move responsibility for raising and error to those redirect calls.

#### Fetch requests don't evaluate to true with the `requirest.hr?` helper

The other problem preventing us from returning `401 - unauthorized` was that we were doing a full page redirect when the current session wasn't found. Our generated `home/index.html.erb` used the fetch API and `LoginProtection` was only looking for `xhr` requests to handle JS requests 🤦 . 

1. Added `requseted_by_javascript?` conditional to better determine API requests vs browser requests. [MDN indicates](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#textjavascript) the official MIME type is text/javascript though we'll support the legacy application/javascript as well.
2. Updated our `home_controller/index.html.erb` to add the correct content type header to act as a better example.

1. Clone [test repo that reproduces expired token](https://github.com/dansundy/app_gem_error_test)
2. Add `Content-Type: "text/javascript"` header to `/fetch` request in `home_controller/index.html.erb`
3. Create test app with that test repo
4. Ensure server responds with expected `401` instead of `500` with bad tokens.

### Things to focus on

Is changing the raise behavior in `current_shopify_domain` acceptable? Or is this too big of a breaking change? If so, we can rescue in `FrameAncestors` and use our wildcard fallback domain. 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ Unreleased
 ----------
 * Fixes a bug with `EnsureAuthenticatedLinks` causing deep links to not work [#1549](https://github.com/Shopify/shopify_app/pull/1549)
 * Ensure online token is properly used when using `current_shopify_session` [#1566](https://github.com/Shopify/shopify_app/pull/1566)
+* Emit a deprecation notice for wrongly-rescued exceptions [#1530](https://github.com/Shopify/shopify_app/pull/1530)
+* Log a deprecation warning for the use of incompatible controller concerns [#1560](https://github.com/Shopify/shopify_app/pull/1560)
 * Fixes bug with expired sessions for embedded apps returning a 500 instead of 401 [#1580](https://github.com/Shopify/shopify_app/pull/1580)
 
 21.2.0 (Oct 25, 2022)
--- a/README.md
+++ b/README.md
@@ -37,21 +37,21 @@ This gem requires that you have the following credentials:
 rails new my_shopify_app
 ```
 
-2. Add the Shopify App gem to `my_shopify_app`'s Gemfile.
+2. Add the Shopify App gem to the app's Gemfile:
 
 ```sh
 bundle add shopify_app
 ```
 
-3. Create a `.env` file in the root of `my_shopify_app` to specify your full host and Shopify API credentials:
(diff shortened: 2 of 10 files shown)


ifyouseewendy:
Hey guys, I'm working on upgrading shopify_app for our app and running into the similar issue. More context here https://github.com/Shopify/script-editor/pull/3145#issuecomment-1319053421 . 

image

After some debugging, I realized that the error comes from [frame_ancestors.rb](https://github.com/Shopify/shopify_app/blob/main/lib/shopify_app/controller_concerns/frame_ancestors.rb#L8-L12)

```rb
        policy.frame_ancestors(-> do
          domain_host = current_shopify_domain || "*.#{::ShopifyApp.configuration.myshopify_domain}"
          "#{ShopifyAPI::Context.host_scheme}://#{domain_host} https://admin.shopify.com"
        end)
```

that [`current_shopify_domain`](https://github.com/Shopify/shopify_app/blob/cc3b66fb8ceb567e2f630f61e543221e4a2c52bf/lib/shopify_app/controller_concerns/login_protection.rb#L212) raises an `::ShopifyApp::ShopifyDomainNotFound` when there is no shopify domain found.

By searching in the shopify_app repo, I noticed this PR happens to be the right fix.

FYI, I pulled the PR and tested. It works as expected. Thank you for the fix 👍

andyw8 (approved):

PR #1581, opened 2022-11-15
Adding Deprecation Logs

klenotiw:
### What this PR does

We had two controllers concerns that we wanted to rename, but still have our old names to be backwards compatible for the time being.

`RequireKnownShop => `EnsureInstalled`
`Authenticated` => `EnsureHasSession`

closes https://github.com/Shopify/shopify_app/issues/1502

---

We also added a bunch of deprecation messages for the following features.
- Marketing extensions/ verification
- Script tag manager
- ITP (Intelligent Tracking Prevention)

### Things to focus on

1. Are all of the messages good?
2. Are all of the versions numbers good? This version number is when the feature will no longer be supported.
3. Am I missing any logs anywhere? Or did I not fully deprecate something?
4. Do our tests catch if the deprecation goes over the version number?

### Checklist


test/controllers/concerns/ensure_installed_test.rb
```diff
@@ -0,0 +1,92 @@
...
+    ShopifyApp::SessionRepository.expects(:retrieve_shop_session_by_shopify_domain).returns(true)
+
+    shopify_domain = "shop1.myshopify.com"
+
+    get :index, params: { shop: shopify_domain }
+
+    assert_response :ok
+  end
+
+  test "detects incompatible controller concerns" do
```
klenotiw, line 64: I have no idea why this works on `RequireKnownShop` but not the new name.

app/controllers/concerns/shopify_app/ensure_installed.rb
```diff
@@ -0,0 +1,58 @@
...
+    include ShopifyApp::RedirectForEmbedded
+
+    included do
+      if ancestors.include?(ShopifyApp::LoginProtection)
+        message = <<~EOS
+        We detected the use of incompatible concerns (EnsureInstalled and LoginProtection) in #{name},
+        which may lead to unpredictable behavior. In a future release of this library this will raise an error.
+        EOS
+
+        ShopifyApp::Logger.deprecated(message ,"22.0.0")
```
klenotiw, line 15: If we want to do minimum runway of 90 days, I think we need to bump this to "23.0.0", since we do a major version update every ruby-api update.

app/controllers/concerns/shopify_app/require_known_shop.rb
```diff
@@ -3,54 +3,16 @@
...
-
-      url.query = URI.encode_www_form(
-        shop: params[:shop],
-        host: params[:host],
-        return_to: request.fullpath,
-      )
-
-      url.to_s
+    def self.included(mod)
+      ShopifyApp::Logger.deprecated("RequireKnownShop has been replaced by to EnsureInstalled."\
```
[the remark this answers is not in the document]
klenotiw: Oh yeah I think thats a really cool idea. Sounds like it would be very similar to the migration POC you made.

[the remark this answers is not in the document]
nelsonwittwer: I'm a fan of iterative improvement. This is the first time we've really used deprecation logging in our api / app gem. Can we ship this as is for now and if we want to invest in more feature rich IDE tooling in the future, we can do so then?

lib/shopify_app/managers/scripttags_manager.rb
```diff
@@ -24,6 +24,7 @@ def self.build_src(scripttags, domain)
     attr_reader :required_scripttags, :shop_domain
 
     def initialize(scripttags, shop_domain)
+      ShopifyApp::Logger.deprecated("The ScripttagsManager will become deprecated in an upcoming version", "22.0.0")
```
klenotiw, line 27: Is on initialize enough? Since this is a class we could use [this](https://api.rubyonrails.org/classes/ActiveSupport/Deprecation/MethodWrapper.html#method-i-deprecate_methods) to send a message every time one of its methods gets called?

app/controllers/concerns/shopify_app/ensure_installed.rb
```diff
@@ -0,0 +1,59 @@
...
+
+        ShopifyApp::Logger.deprecated(message, "22.0.0")
+      end
+
+      before_action :check_shop_domain
+      before_action :check_shop_known
+    end
+
+    def current_shopify_domain
+      return if params[:shop].blank?
```
nelsonwittwer, line 23:
```suggestion
    if params[:shop].blank?
      ShopifyApp::Logger.info("Could not identify installed shop from current_shopify_domain")
      return
    end
```

app/controllers/concerns/shopify_app/ensure_installed.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      @shopify_domain ||= ShopifyApp::Utils.sanitize_shop_domain(params[:shop])
+      ShopifyApp::Logger.info("Current Shopify Domain - #{@shopify_domain}")
```
nelsonwittwer, line 26:
```suggestion
      ShopifyApp::Logger.info("Installed store:  #{@shopify_domain} - deduced from Shopify Admin params")
```

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -210,7 +212,9 @@ def fullpage_redirect_to(url)
     end
 
     def current_shopify_domain
-      sanitized_shop_name || current_shopify_session&.shop
+      shopify_domain = sanitized_shop_name || current_shopify_session&.shop
+      ShopifyApp::Logger.info("Current Shopify Domain - #{shopify_domain}")
```
nelsonwittwer, line 216:
```suggestion
      ShopifyApp::Logger.info("Installed store  - #{shopify_domain} deduced from user session")
```

app/controllers/concerns/shopify_app/ensure_installed.rb
```diff
@@ -0,0 +1,57 @@
...
+    include ShopifyApp::RedirectForEmbedded
+
+    included do
+      if ancestors.include?(ShopifyApp::LoginProtection)
+        message = <<~EOS
+          We detected the use of incompatible concerns (EnsureInstalled and LoginProtection) in #{name},
+          which may lead to unpredictable behavior. In a future release of this library this will raise an error.
+        EOS
+
+        ShopifyApp::Logger.deprecated(message, "22.0.0")
```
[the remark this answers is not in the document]
nelsonwittwer: Works for me! This is a simple rename for users so I don't see a point in dragging this out

lib/shopify_app/managers/scripttags_manager.rb
nelsonwittwer: I really like that idea to use `ActiveSupport::Deprecation::MethodWrapper`, however I'd vote we do that after we merge what we have here.

test/controllers/concerns/ensure_installed_test.rb
```diff
@@ -62,34 +62,29 @@ def index
...
-    ActiveSupport::Deprecation.silenced = false
-    ShopifyAPI::Context.stubs(:log_level).returns(:warn)
-
-    assert_deprecated(/incompatible concerns/) do
-      Class.new(ApplicationController) do
-        include ShopifyApp::LoginProtection
-        include ShopifyApp::EnsureInstalled
-      end
+    version = "22.0.0"
+    ShopifyApp::Logger.expects(:deprecated).with(regexp_matches(/incompatible concerns/), version)
```
nelsonwittwer, line 66: 🆒 ! Today I learned `regexp_matches` can be used here!

test/test_helper.rb
```diff
@@ -58,4 +59,8 @@ def mock_session(shop: "my-shop.myshopify.com", scope: ShopifyApp.configuration.
 
     mock_session
   end
+
```
nelsonwittwer, line 62:
```suggestion
##
# If a test fails with this assertion it means the behavior should now be removed from the codebase. 
# The deprecation schedule gives users time to upgrade before the functionality can safely removed.
```

PR #1593, opened 2022-11-22
WIP: Add reauth option

andyw8:
### Things to focus on

1. 
2. 
3. 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 Unreleased
 ----------
+* reauth
 * Fixes a bug with `EnsureAuthenticatedLinks` causing deep links to not work [#1549](https://github.com/Shopify/shopify_app/pull/1549)
 * Ensure online token is properly used when using `current_shopify_session` [#1566](https://github.com/Shopify/shopify_app/pull/1566)
 * Added debug logs, you can read more about logging (here)[./docs/logging.md]. [#1545](https://github.com/Shopify/shopify_app/pull/1545)
--- a/app/controllers/concerns/shopify_app/shop_access_scopes_verification.rb
+++ b/app/controllers/concerns/shopify_app/shop_access_scopes_verification.rb
@@ -34,7 +34,7 @@ module ShopifyApp
     end
 
     def shop_login
-      ShopifyApp::Utils.shop_login_url(shop: params[:shop], host: params[:host], return_to: request.fullpath)
+      ShopifyApp::Utils.shop_login_url(shop: params[:shop], host: params[:host], return_to: params[:return_to], reauthorize: true)
     end
   end
 end
--- a/app/controllers/shopify_app/sessions_controller.rb
+++ b/app/controllers/shopify_app/sessions_controller.rb
@@ -12,7 +12,11 @@ module ShopifyApp
     end
 
     def new
-      authenticate if sanitized_shop_name.present?
+      if params[:reauthorize].present?
+        flash[:error] = "Due to a change in permissions for this app it needs to be reauthorized."
+      elsif sanitized_shop_name.present?
+        authenticate
+      end
     end
 
     def create
@@ -35,6 +39,7 @@ module ShopifyApp
     private
 
     def authenticate
+      ShopifyApp::Logger.debug("# authenticate")
       return render_invalid_shop_error unless sanitized_shop_name.present?
 
       copy_return_to_param_to_session
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -96,6 +96,7 @@ module ShopifyApp
           end
         end
 
+        #
         url ||= login_url_with_optional_shop
 
         ShopifyApp::Logger.debug("Setting Reauthorize-Url to #{url}")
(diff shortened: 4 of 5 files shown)


shopify_app.gemspec
```diff
@@ -25,6 +25,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency("byebug")
   s.add_development_dependency("minitest")
   s.add_development_dependency("mocha")
+  s.add_development_dependency("nokogiri")
```
andyw8, line 28: `nokogiri` was already in the `Gemfile.lock` as a transative dependency, but now that I'm using it in a test I want to make it explicit.

app/controllers/concerns/shopify_app/shop_access_scopes_verification.rb
```diff
@@ -33,8 +29,14 @@ def current_shopify_domain
...
 
-    def shop_login
-      ShopifyApp::Utils.shop_login_url(shop: params[:shop], host: params[:host], return_to: request.fullpath)
+    def reauthorized_shop_login_url
+      ShopifyApp::Utils.shop_login_url(shop: params[:shop], host: params[:host], return_to: request.fullpath,
+        reauthorize: true)
+    end
+
+    def fullpage_redirect_to(url)
+      render("shopify_app/shared/redirect", layout: false,
```
andyw8, line 38: This is very similar to the method in `LoginProtection` but I didn't think it was worth extracting out.

app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -12,7 +12,11 @@ class SessionsController < ActionController::Base
     end
 
     def new
-      authenticate if sanitized_shop_name.present?
+      if params[:reauthorize].present?
+        flash[:error] = "Due to a change in required access scopes this app will need to be reauthorized."
```
andyw8, line 16: TODO: translation?

app/controllers/concerns/shopify_app/shop_access_scopes_verification.rb
```diff
@@ -13,11 +13,7 @@ module ShopAccessScopesVerification
 
     def login_on_scope_changes
       if scopes_mismatch?
-        if embedded_param?
-          redirect_for_embedded
-        else
-          redirect_to(shop_login)
-        end
+        fullpage_redirect_to(reauthorized_shop_login_url)
```
andyw8, line 16: I'm still considering if this would be a suitable approach for all apps.

andyw8:
(this branch is already running in prod for shopify-graphql-app)

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 Unreleased
 ----------
+* Add reauth option [#1593](https://github.com/Shopify/shopify_app/pull/1593)
```
andyw8, line 3: TODO: add better summary

app/controllers/concerns/shopify_app/shop_access_scopes_verification.rb
```diff
@@ -13,11 +13,8 @@ module ShopAccessScopesVerification
 
     def login_on_scope_changes
       if scopes_mismatch?
-        if embedded_param?
```
andyw8, line 16: The existing logic here may be wrong. `embedded_param?` is defined as:

```
def embedded_param?
  embedded_redirect_url? && params[:embedded].present? && params[:embedded] == "1"
end
```

and `embedded_redirect_url?` is checking for the presence of the `embedded_redirect_url` configuration value.

But the upgrade guide says this value is optional:
https://github.com/Shopify/shopify_app/blob/main/docs/Upgrading.md#upgrading-to-v2010

paulomarg: I think it might still happen on embedded requests, so we might want to handle it. AFAIK `embedded_redirect_url` is optional, but it has a default value so it should still work.

There should be a route in the app that handles the `/exitiframe` path that it defaults to, if this is the setting I think it is.

app/controllers/concerns/shopify_app/shop_access_scopes_verification.rb
paulomarg: I think it is, but it relies on the "old" CDN-based app bridge to do the actual redirecting.

lib/shopify_app/access_scopes/shop_strategy.rb
```diff
@@ -6,7 +6,9 @@ class ShopStrategy
       class << self
         def update_access_scopes?(shop_domain)
           shop_access_scopes = shop_access_scopes(shop_domain)
-          configuration_access_scopes != shop_access_scopes
+          result = configuration_access_scopes != shop_access_scopes
```
paulomarg, line 9: Should we do something similar for the user strategy?

app/controllers/shopify_app/sessions_controller.rb
```diff
@@ -12,7 +12,11 @@ class SessionsController < ActionController::Base
     end
 
     def new
-      authenticate if sanitized_shop_name.present?
+      if params[:reauthorize].present?
+        flash[:error] = I18n.t("reauth_needed")
```
paulomarg, line 16: I'm not sure I understand the idea here - will we just show a message showing re-auth is needed, rather than actually trigger it?

PR #1599, opened 2022-11-30
Add configuration option for user access scopes strategy

andyw8:
### What this PR does

The gem currently provides two strategies for verifying user scopes:

* ShopifyApp::AccessScopes::NoopStrategy 
* ShopifyApp::AccessScopes::UserStrategy

Some apps (likely only first party apps made by Shopify) may have special requirements and need to use a different strategy. This PR adds a configuration option to allow that.

Reach out to me on Slack to discuss.

### Things to focus on

1. Is there any way this could break an existing app?

### Checklist

### DIFF
--- a/lib/shopify_app/configuration.rb
+++ b/lib/shopify_app/configuration.rb
@@ -83,7 +83,17 @@ module ShopifyApp
       ShopifyApp::AccessScopes::ShopStrategy
     end
 
+    def user_access_scopes_strategy=(class_name)
+      unless class_name.is_a?(String)
+        raise ConfigurationError, "Invalid user access scopes strategy - expected a string"
+      end
+
+      @user_access_scopes_strategy = class_name.safe_constantize
+    end
+
     def user_access_scopes_strategy
+      return @user_access_scopes_strategy if @user_access_scopes_strategy
+
       return ShopifyApp::AccessScopes::NoopStrategy unless reauth_on_access_scope_changes
 
       ShopifyApp::AccessScopes::UserStrategy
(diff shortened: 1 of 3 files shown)


test/shopify_app/configuration_test.rb
```diff
@@ -209,4 +209,14 @@ class ConfigurationTest < ActiveSupport::TestCase
     assert_equal ShopifyApp::AccessScopes::ShopStrategy, ShopifyApp.configuration.shop_access_scopes_strategy
     assert_equal ShopifyApp::AccessScopes::UserStrategy, ShopifyApp.configuration.user_access_scopes_strategy
   end
+
+  test "user access scopes strategy is configurable" do
+    my_strategy = Object
+    ShopifyApp.configure do |config|
+      config.user_access_scopes_strategy = my_strategy
```
andyw8, line 216: We set values such as `user_session_repository` using strings rather than constants, I wonder if need to that here too... 🤔

andyw8: Updated to use a string.

lib/shopify_app/configuration.rb
```diff
@@ -83,7 +83,17 @@ def shop_access_scopes_strategy
       ShopifyApp::AccessScopes::ShopStrategy
     end
 
+    def user_access_scopes_strategy=(class_name)
+      unless class_name.is_a?(String)
```
andyw8, line 87: Using a string to be consistent with `user_session_repository` and `shop_session_repository`.

PR #1600, opened 2022-11-30
Move covered scopes check into strategy

andyw8:
### What this PR does

There are two places that we check the user's scopes when determining if a re-auth is needed:

* `update_access_scopes?` in the access scopes user strategy.
* in `LoginProtection#activate_shopify_session` 

In #1599 we added an option to set the user strategy. But this has a gap, since only one of the above listed places is within the strategy. To have proper control over these checks, we need both of them to be in the strategy.

This PR aims to not alter any existing behaviour, it only moves code around.

### Things to focus on

1. 
2. 
3. 

### Checklist

### DIFF
--- a/lib/shopify_app/access_scopes/user_strategy.rb
+++ b/lib/shopify_app/access_scopes/user_strategy.rb
@@ -12,6 +12,11 @@ module ShopifyApp
             "#update_access_scopes? requires user_id or shopify_user_id parameter inputs")
         end
 
+        def covered_scopes?(current_shopify_session)
+          # NOTE: this not Ruby's `covers?` method, it is defined in ShopifyAPI::Auth::AuthScopes
+          current_shopify_session.scope.to_a.empty? || current_shopify_session.scope.covers?(ShopifyAPI::Context.scope)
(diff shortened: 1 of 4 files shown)


lib/shopify_app/access_scopes/user_strategy.rb
```diff
@@ -12,6 +12,11 @@ def update_access_scopes?(user_id: nil, shopify_user_id: nil)
             "#update_access_scopes? requires user_id or shopify_user_id parameter inputs")
         end
 
+        def covered_scopes?(current_shopify_session)
```
andyw8, line 15: Maybe `covers_scopes?`?

andyw8:
@paulomarg @nelsonwittwer sharing as a draft, any thoughts so far?

lib/shopify_app/access_scopes/user_strategy.rb
andyw8: Done.

paulomarg (approved):
Just a non-blocking comment.

PR #1609, opened 2022-12-12
packaging for release 21.3.1

nelsonwittwer:


klenotiw (changes requested):
Oh wait don't we need to change `package.json`?

klenotiw:
Looks like we haven't changed it in a few releases. Im not completely sure what it does tho? 
https://github.com/Shopify/shopify_app/blob/main/package.json

Issue #1591, opened 2022-11-21
Verify Access Token


asecondwill:
When the app review team uninstall and reinstall my app with their automated tools, they are getting an error because the uninstall webhook has not yet triggered the delayed job to delete the Shop record.  

I've been advised over on Shopify's slack page

> You should not be relying on an uninstalled webhook to determine a shop's status with your app. instead, each time a shop hits your app, you should make a request to your server to first determine if the access token you have for them is valid (if so, they have your app installed - if not, redirect to OAuth) and then from there check for a valid charge ID using the billing API (if not, redirect to billing page). 

Should this be implmented by this gem or the Shopify demo implementations:

https://github.com/Shopify/next-gen-auth-app-demo/tree/master/app
https://github.com/Shopify/turbolinks-jwt-sample-app

If not, how should my app do that?

nelsonwittwer:
👋 We need to do a better job of that here in the gem. I'll see what I can do!

wisniewski94:
Hi @nelsonwittwer, thanks for your action on this. In #1612 if I understood correctly, you are attempting to create a Rest Client with the stored in DB session. 

1. Can you share when it should happen so the check passes?
2. I was dead sure that if an app is embedded and Shopify's test tool attempts to open it, it will see "There's no page at the given address" and the test should pass. I agree that this is the only way to check access token validity however I have a feeling that even if it's a lightweight call it's still a call that must be made and there is *some* weight in the end. Isn't it the problem on the test tool end? I mean, if accessToken is not valid anymore and webhooks cannot be received, why do we bother in the first place?

PR #1612, opened 2022-12-13
WIP - Validate store sessions 

nelsonwittwer:
### What this PR does

When a store uninstalls an app we are currently dependent upon the uninstall webhook to tell our apps that a store should no longer be able to access the API / app resources. This leads to apps getting rejected using our templates as reported in #1591 .  

This PR will do a lightweight API call after it has found the shop session in storage to ensure it is still valid. If it is not, it will redirect to the embedded login url.

Still figuring out how to test this locally 😅 

### Checklist

### DIFF
--- a/app/controllers/concerns/shopify_app/ensure_installed.rb
+++ b/app/controllers/concerns/shopify_app/ensure_installed.rb
@@ -17,6 +17,7 @@ module ShopifyApp
 
       before_action :check_shop_domain
       before_action :check_shop_known
+      before_action :validate_emebedded_session_is_active, if: :embedded_param?
     end
 
     def current_shopify_domain
@@ -28,6 +29,10 @@ module ShopifyApp
       @shopify_domain ||= ShopifyApp::Utils.sanitize_shop_domain(params[:shop])
       ShopifyApp::Logger.info("Installed store:  #{@shopify_domain} - deduced from Shopify Admin params")
       @shopify_domain
+    end
+
+    def installed_shop_session
+      @installed_shop_session ||= @shop
     end
 
     private
@@ -57,6 +62,13 @@ module ShopifyApp
       )
 
       url.to_s
+    end
+
+    def validate_emebedded_session_is_active
+      client = ShopifyAPI::Clients::Rest::Admin.new(session: installed_shop_session)
+      client.get(path: "shop")
+    rescue ShopifyAPI::Errors::HttpResponseError
+      redirect_for_embedded
     end
   end
 end
--- a/test/controllers/concerns/ensure_installed_test.rb
+++ b/test/controllers/concerns/ensure_installed_test.rb
@@ -87,4 +87,26 @@ class EnsureInstalledTest < ActionController::TestCase
 
     assert_within_deprecation_schedule(version)
   end
+
+  test "redirects to redirect_for_embedded if app is no longer installed according to the API but not the session repository" do
+    session = mock
+    ShopifyApp::SessionRepository.stubs(:retrieve_shop_session_by_shopify_domain).returns(session)
+
+    client = mock
+    ShopifyAPI::Clients::Rest::Admin.expects(:new).with(session: session).returns(client)
+    uninstalled_http_error = ShopifyAPI::Errors::HttpResponseError.new(
+      response: ShopifyAPI::Clients::HttpResponse.new(
+        code: 404,
+        headers: {},
+        body: "Invalid API key or access token (unrecognized login or wrong password)",
+      ),
+    )
+    client.expects(:get).with(path: "shop").raises(uninstalled_http_error)
+
+    embedded_redirect_url = "reverse_card_on_reverse_card.embedded.joke"
+    ShopifyApp.configuration.stubs(:embedded_redirect_url).returns(embedded_redirect_url)
+
+    @controller.expects(:redirect_for_embedded)
+    get :index, params: { shop: "shop1.myshopify.com", embedded: "1" }
+  end
 end


app/controllers/concerns/shopify_app/ensure_installed.rb
```diff
@@ -58,5 +63,12 @@ def shop_login
 
       url.to_s
     end
+
+    def validate_emebedded_session_is_active
+      client = ShopifyAPI::Clients::Rest::Admin.new(session: installed_shop_session)
+      client.get(path: "shop")
+    rescue ShopifyAPI::Errors::HttpResponseError
```
nelsonwittwer, line 70: I wish the error coming back from API were more descriptive, but I get why they do it. Shouldn't give bad guys clues to bad session states. It does make rescuing from this and handling it more delicate.

I'm assuming we wouldn't want to drop the shop record from storage here? 🤔

app/controllers/concerns/shopify_app/ensure_installed.rb
```diff
@@ -17,6 +17,7 @@ module EnsureInstalled
 
       before_action :check_shop_domain
       before_action :check_shop_known
+      before_action :validate_emebedded_session_is_active, if: :embedded_param?
```
nelsonwittwer, line 20: @paulomarg - I know you mentioned you wanted to use the `params[:embedded] == 1` flag here. I'm assuming here we want to use [all these conditionals from embedded_param as well?](https://github.com/Shopify/shopify_app/blob/main/lib/shopify_app/controller_concerns/redirect_for_embedded.rb#L14)

paulomarg: Yeah, the idea is that we only need to run this check IF we're planning on redirecting to the admin (`embedded !== 1`).

This keeps the number of requests we need to make to a minimum because normally the app would just be loaded straight from the admin (`embedded == 1`), and for that to happen we know it's installed (Admin still references it).

app/controllers/concerns/shopify_app/ensure_installed.rb
paulomarg: We could further test this for a `401` code, which would indicate the token is invalid.

paulomarg: I would consider keeping the app_uninstalled webhook around for actually deleting stuff from the db to minimize the amount of work we do here / make sure we get all sessions for a shop.

app/controllers/concerns/shopify_app/ensure_installed.rb
```diff
@@ -58,5 +63,12 @@ def shop_login
 
       url.to_s
     end
+
+    def validate_emebedded_session_is_active
+      client = ShopifyAPI::Clients::Rest::Admin.new(session: installed_shop_session)
+      client.get(path: "shop")
+    rescue ShopifyAPI::Errors::HttpResponseError => error
+      redirect_for_embedded if error.code == 401 # unauthorized due to uninstall
```
paulomarg, line 71: Would this be a `redirected_for_embedded` situation? I think we would call this when NOT embedded, right?

nelsonwittwer: ah ya, that's not clear because the conditional for this method is listed in the `before_action` declaration. I'll fix

nelsonwittwer: As I understand it, we should only be doing this session sanity check if it is embedded, right?

nelsonwittwer: We should redirect to login, which we sadly call `login_url` for reasons 🤦

app/controllers/concerns/shopify_app/ensure_installed.rb
```diff
@@ -58,5 +63,14 @@ def shop_login
 
       url.to_s
     end
+
+    def validate_non_embedded_session
+      return if loaded_directly_from_admin?
+
+      client = ShopifyAPI::Clients::Rest::Admin.new(session: installed_shop_session)
+      client.get(path: "shop")
+    rescue ShopifyAPI::Errors::HttpResponseError => error
```
paulomarg, line 72: Should we re-throw if the error isn't a 401 here? Unlikely, but may happen 🤷

app/controllers/concerns/shopify_app/ensure_installed.rb
```diff
@@ -58,5 +63,14 @@ def shop_login
 
       url.to_s
     end
+
+    def validate_non_embedded_session
```
paulomarg, line 67: Do you think calling this `validate_session_before_embedding` would make it easier to understand this method?

test/controllers/concerns/ensure_installed_test.rb
```diff
@@ -61,6 +66,36 @@ def index
     assert_response :ok
   end
 
+  test "redirects to login_url (oauth path) to reinstall the app if the store's session token is no longer valid" do
+    ShopifyApp.configuration.stubs(:embedded_app?).returns(true)
```
paulomarg, line 70: I think you can actually change the configuration value as part of the tests, I'm pretty sure we reset them for each test.

test/controllers/concerns/ensure_installed_test.rb
```diff
@@ -61,6 +66,36 @@ def index
...
+        headers: {},
+        body: "Invalid API key or access token (unrecognized login or wrong password)",
+      ),
+    )
+    client.expects(:get).with(path: "shop").raises(uninstalled_http_error)
+
+    shopify_domain = "shop1.myshopify.com"
+    get :index, params: { shop: shopify_domain }
+
+    assert_response :redirect
```
paulomarg, line 89: Should we assert that the location we go to is `/login`?

paulomarg (approved):
LGTM, only had non-blocking comments!

uurcank:
@nelsonwittwer  @paulomarg  this change checks whether the token is valid but callback_controller does not update the offline token. From what I see if store_session_storage is not empty, same values stay and database not updated with new token value.

Could you please confirm whether this is the case?

netwire88:
Will all the documentation that refers to `RequireKnownShop` need to be changed to `EnsureInstalled`?

Also, does `EnsureInstalled` work with `EmbeddedApp` and `ShopAccessScopesVerification` in an unauthenticated controller?

```
  include ShopifyApp::EnsureInstalled
  include ShopifyApp::ShopAccessScopesVerification
```

PR #1622, opened 2022-12-21
Improve github templates

paulomarg:
Now that we provide better logging capabilities in the library package, we should encourage folks to increase log verboseness before opening an issue, so that we can get as much information as possible upfront.

This PR also tweaks the templates' general wording so they're simpler, and makes the comments actual comments so they don't show up in created issues.


cquemin (approved):
Lets do this!

PR #1633, opened 2023-01-05
Update shopify_api to 12.4.0

paulomarg:
This PR just bumps the dependency on `shopify_api` to the latest version 12.4.0.


paulomarg:
> Wait shouldn't the gemspec file be changed?

Yes, but technically the new version still matches the old `~> 12.3` constraint. Probably good to update it anyway, so I did it 🙂

PR #1636, opened 2023-01-06
Set `access_scopes` column to string by default

### LINKED ISSUE #1411: V19.0.1 - sorbet error?
Attempting to re-install my dev app, I foolishly uninstalled (knowing this is always a huge pain to debug on version bumps). 

But for once a new error has appeared which was not seen before v19 releases.

```ruby
Parameter 'scope': Expected type T.any(String, T::Array[String]), got type NilClass
Caller: /usr/local/bundle/gems/sorbet-runtime-0.5.9878/lib/types/private/methods/call_validation.rb:117
Definition: /usr/local/bundle/gems/shopify_api-10.0.2/lib/shopify_api/auth/session.rb:52
```

Screen Shot 2022-04-12 at 6 34 27 PM

kirillplatonov:
### What this PR does

Access scopes can't be nil anymore as it will trigger a sorbet error in ShopifyAPI 10+.
```rb
Parameter 'scope': Expected type T.any(String, T::Array[String]), got type NilClass
Caller: /usr/local/bundle/gems/sorbet-runtime-0.5.9878/lib/types/private/methods/call_validation.rb:117
Definition: /usr/local/bundle/gems/shopify_api-10.0.2/lib/shopify_api/auth/session.rb:52
```

Resolves #1411

### Checklist


nelsonwittwer (approved):

vfonic:
I understand I'm half a year late to the party. Just wanted to understand what's the general guidance on making such changes that are not backwards compatible?
For example, if I had an app which already had the access scopes migration (without default `""` and `null: true`), how can I upgrade to the new gem version? Do I need to manually create a new migration adding default `""` and `null: false`?

Is there any reason why this isn't the default guidance/workflow?

Also, unrelated to this particular issue, I strongly believe that @kirillplatonov should be paid for his work on Shopify projects. @nelsonwittwer, you're getting paid for your work, correct?
Why would @kirillplatonov volunteer to upgrade your infrastructure and the only benefit he'd get is to not have to fork the repo to have his fixes in? ...but I guess that's not the status quo, it's expected from outside developers to contribute for free... (that's why I stopped contributing)
https://github.com/Shopify/shopify_app/pulls?q=author%3Akirillplatonov

vfonic:
Not sure what the issue was, but `bundle update shopify_app` would not update `shopify_app` gem from 21.2.0 to 21.5.0. I tried specifying version restriction in several different ways, even removing `'~> 21.2'`. The only thing that worked was explicitly setting `gem 'shopify_app', '~> 21.5'`.

PR #1639, opened 2023-01-11
Don't redirect on mismatching scopes if configuration is disabled

minichate:
### What this PR does

This skips the redirect-to-reauthenticate block in `LoginProtection` 

We had an incident earlier this week where we were doing a scope upgrade and despite us disabling [`reauth_on_access_scope_changes`](https://github.com/Shopify/shopify_app/blob/main/lib/shopify_app/configuration.rb#L22) ahead of time.

The scopes were updated/backfilled on the Shopify Core side, but since the cached scopes in `user` and `shop` models were a subset of the actually-approved scopes in Core requests were rejected with a `401` response without even sending them to Core.

This was surprising behaviour to us since we'd disabled the feature intentionally ahead of time.

1. Have an existing set of scopes cached on the `user` and `shop` models
2. Add a new scope c (ie: in `config.scope`)
3. Set `reauth_on_access_scope_changes = false` in your app's `shopify_app` initializer
4. Try and use the app; you'll get a 401 response from XHR requests, and a redirect for regular document requests.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased
 ----------
 
+* Fixed bug where authentication redirect could still happen even though `reauth_on_access_scope_changes` is set to `false` [#1639](https://github.com/Shopify/shopify_app/pull/1639)
+
 21.4.0 (Jan 5, 2023)
 ----------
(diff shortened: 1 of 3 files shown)


jeremywrowe (approved):

jeremywrowe:
I approved, the code looks good. In addition to the change there needs to be updates to documentation and the changelog.

PR #1688, opened 2023-05-05
Add ability to control test charges flag when creating application charges

gcaprio:
### What this PR does

The gem is currently hard coded to only support test charges in non-Rails production environments. However, there are valid reasons why you would want to set the Rails.env to production and still test charges. For example testing a staging Shopify app w/ a development shop (which cannot support real charges"

I added a flag to the `BillingConfiguration` object and defaulted is value to the same !Rails.env.production? check so default behavior is not changed or impacted.

### Things to focus on

1. Is the default behavior correctly maintained
2. Is the environment variable named correctly and are the generator updates correctly pulling in the environment variable?
3. There are no tests to around creating charges, so I simply added one to verify the flag is being defaulted and toggled correctly.

### Checklist

- [X] Update `CHANGELOG.md` if the changes would impact users
- [-] Update `README.md`, if appropriate.
- [-] Update any relevant pages in `/docs`, if necessary
- [-] For security fixes, the [Disclosure Policy](https://github.com/Shopify/shopify_app/blob/master/SECURITY.md#disclosure-policy) must be followed.


gcaprio:
Just bumping this  @nelsonwittwer @Al-Campuzano or @gonzaloriestra (looking at the last people to merge PRs) any chance you can check this out?

gonzaloriestra (approved):
Thanks for your contribution!

PR #1691, opened 2023-05-18
Use new CLI env variables

paulomarg:
### What this PR does

With https://github.com/Shopify/shopify-dev/pull/33718, the CLI will deprecate some env var names. While they will still be present so previous apps won't break, we should migrate new apps as soon as possible to the new format.

This should run both on the latest version of the CLI or with a current one.

### Checklist


test/app_templates/config/initializers/shopify_app.rb
```diff
@@ -2,8 +2,8 @@
 
 ShopifyApp.configure do |config|
   config.application_name = "name"
-  config.api_key = ENV["SHOPIFY_API_KEY"]
-  config.secret = ENV["SHOPIFY_API_SECRET"]
+  config.api_key = ENV["SHOPIFY_APP_API_KEY"] || ENV["SHOPIFY_API_KEY"]
+  config.secret = ENV["SHOPIFY_APP_API_SECRET"] || ENV["SHOPIFY_API_SECRET"]
   config.scope = "read_orders read_products"
```
mkevinosullivan, line 7: Should we also start using the `SHOPIFY_APP_SCOPES` env var?

mkevinosullivan (approved):

PR #1695, opened 2023-05-29
docs: fix inline job syntax error

shafy:
### What this PR does

In the authentication docs, there's a syntax error when defining an line job. The correct line should be:

```ruby
config.after_authenticate_job = { job: "Shopify::AfterAuthenticateJob", inline: true }
```

`Shopify::AfterAuthenticateJob` needs to be in quotes.

### Checklist


jagthedrummer:
not stale

nelsonwittwer:
Curious if you were running into an error with the constant instead of string. [It looks like it should work with either a string or a constant](https://github.com/Shopify/shopify_app/blob/main/app/controllers/shopify_app/callback_controller.rb#L149). Either way, I like being consistent :)

nelsonwittwer (approved):

PR #1704, opened 2023-07-02
Fix typo in customers/redact webhook topic

flavio-b:
Fix typo. Webhook topic should be "customers/redact".

### What this PR does

Just changes a webhook type in `shopify_app.rb` initializer template. 

Create a brand new app using the generators. Submit a "customers/redact" webhook to the app.


nelsonwittwer (approved):
Thanks for fixing this!

PR #1716, opened 2023-08-22
Fix InMemoryUserSessionStore - user.shopify_user_id does not exist

zzooeeyy:
### What this PR does
Using `ShopifyApp` gem's `InMemoryUserSessionStore` doesn't work out of the box 
nomethoderror

`shopify_user_id` does not belong in `ShopifyAPI::Auth::AssociatedUser` , using `id` instead.

Added test around `InMemoryUserSessionStore` as well.

Set configuration user session repository to use `InMemoryUserSessionStore`, see that it fails before this fix, and can store sessions after this fix.
```ruby
# my-app/web/config/initializers/shopify_app.rb
#...
  config.shop_session_repository = ShopifyApp::InMemoryShopSessionStore
  config.user_session_repository= ShopifyApp::InMemoryUserSessionStore
#...
end
```
### Checklist


mllemango (approved):
❤️

PR #1718, opened 2023-08-24
WebhooksManager#recreate_webhooks! should creates webhooks

tomriley:
### What this PR does

``#recreate_webhooks!`` doesn't seem to attempt to create webhooks after destroying them. This patch fixes that.

I have also removed the suggestion of including manual registration of mandatory webhooks from the generator template. [Shopify graphql API documentation](https://shopify.dev/docs/api/admin-graphql/2023-07/enums/WebhookSubscriptionTopic) states that these shouldn't be registered manually via the API and instead need be set up on the partner dashboard (in App Settings). From the docs:

"You don't create webhook subscriptions to [mandatory webhooks](https://shopify.dev/apps/webhooks/configuration/mandatory-webhooks). Instead, you configure mandatory webhooks in your Partner Dashboard as part of your app setup."

To avoid errors, existing users should remove the three mandatory webhooks from their ``ShopifyApp.configuration.webhooks``.

Mandatory webhook configuration should be removed before testing on a real app.

### Things to focus on

1. Is Shopify documentation correct? Are all these manually webhook API subscriptions just being ignored? Why are the subscription not failing? So many questions...

### Checklist

### DIFF
--- a/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
+++ b/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
@@ -10,10 +10,7 @@ ShopifyApp.configure do |config|
   config.log_level = :info
   config.reauth_on_access_scope_changes = true
   config.webhooks = [
-    { topic: "app/uninstalled", address: "webhooks/app_uninstalled"},
-    { topic: "customers/data_request", address: "webhooks/customers_data_request" },
-    { topic: "customers/redact", address: "webhooks/customers_redact"},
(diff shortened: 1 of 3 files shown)


tomriley:
I have signed the CLA!

nelsonwittwer:
Thanks for your help finding the solution here. I stumbled across this same solution with #1743 while trying to track down another issue. Totally agree that mandatory webhooks are goofy here within the app gem.  I'll get that working before I merge in that PR. Closing this and consolidating with #1743

Issue #1684, opened 2023-04-14
Confusion EnsureHasSession vs. Authenticated


netwire88:
In shopify_app 21.4.0, there's this deprecation warning:

```
DEPRECATION WARNING: [22.0.0] [ ShopifyApp | WARN | Shop Not Found ] Authenticated has been replaced by EnsureHasSession. Please use the EnsureHasSession controller concern for the same behavior
```

However, in https://github.com/Shopify/shopify_app/blob/main/docs/shopify_app/authentication.md#shopifyappauthenticated, the example is this. Which is correct?

```
class AuthenticatedController < ApplicationController
  include ShopifyApp::Authenticated
end
```

Trimakas:
would be great to get an answer here.. any answer at all.. ever

djkgamc:
lol

PR #1720, opened 2023-08-25
update controller concern docs

nelsonwittwer:
### What this PR does

Updates controller concerns docs to 
1. Update deprecated naming
2. Clarify public vs private concerns

Closes https://github.com/Shopify/shopify_app/issues/1684


zzooeeyy (approved):
:yay:

PR #1746, opened 2023-11-09
Redirect embedded apps to the provided return_to if it's a fully-formed URL

laynekoftinowmikan:
### Things to focus on

1. 
2. 
3. 

### Checklist

### DIFF
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -69,13 +69,18 @@ module ShopifyApp
     end
 
     def redirect_to_app
-      if ShopifyAPI::Context.embedded?
-        return_to = "#{decoded_host}#{session.delete(:return_to)}"
-        return_to = ShopifyApp.configuration.root_url if deduced_phishing_attack?
-        redirect_to(return_to, allow_other_host: true)
+      if ShopifyAPI::Context.embedded? && !fully_formed_url?(session[:return_to])
+          return_to = "#{decoded_host}#{session.delete(:return_to)}"
+          return_to = ShopifyApp.configuration.root_url if deduced_phishing_attack?
+          redirect_to(return_to, allow_other_host: true)
       else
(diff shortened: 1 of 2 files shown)


app/controllers/shopify_app/callback_controller.rb
```diff
@@ -69,7 +69,7 @@ def update_rails_cookie(api_session, cookie)
     end
 
     def redirect_to_app
-      if ShopifyAPI::Context.embedded?
+      if ShopifyAPI::Context.embedded? && !fully_formed_url?(session[:return_to])
         return_to = "#{decoded_host}#{session.delete(:return_to)}"
         return_to = ShopifyApp.configuration.root_url if deduced_phishing_attack?
```
nelsonwittwer, line 74: We'll still need this `deduced_phishing_attack?` check here as well. The `return_to` param is used with phishing attacks to make links seem legit because they linked back from shopify.

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -69,7 +69,7 @@ def update_rails_cookie(api_session, cookie)
     end
 
     def redirect_to_app
-      if ShopifyAPI::Context.embedded?
+      if ShopifyAPI::Context.embedded? && !fully_formed_url?(session[:return_to])
```
nelsonwittwer, line 72: While I see what this is doing now because I have the context, it isn't super clear as to the intent of the feature. Maybe we add another `return_to` local override that says `return_to = return_url if full_formed_url?`. Something similar to line 74's `return_to` override.

laynekoftinowmikan: Makes sense, I've refactored a bit so it's hopefully more clear. Let me know what you think

rondanino (approved):
Nice. How do we build and use it in shop-server?

nelsonwittwer (approved):

PR #1751, opened 2023-11-13
Packaging for release 21.8.0

nelsonwittwer:


paulomarg (approved):
I think we're missing `bundle install`, as per https://github.com/Shopify/shopify_app/blob/main/docs/Releasing.md?

PR #1767, opened 2023-12-16
Allow more control over the application layout

stefanvermaas:
> **TLDR:** The PR gives an app developer more granular control over what layout to render when using the `shopify_app` gem to build an embedded Shopify app.

When running a Shopify app in embedded mode, every controller that requires an authenticated session will use the embedded app layout by default. This is enforced by the `ShopifyApp::EmbeddedApp` module.

As a default, this is great. It helps the developer to make fewer choices when they start to develop their new Shopify app. However, sometimes one wants to run an embedded app in non-embedded mode.

Some examples of running a normally embedded app in non-embedded mode:
- When the remote environment is a CI;
- When the remote environment is a preview/PR app;
- When the developer wants to run the app in a non-embedded mode for testing.

The static implementation of the embedded app layout causes duplication and maintainability issues when running in both modes.

This PR introduces a more flexible system whereby the developer has more control over what layout is being rendered during the process and allows the developer to ignore the embedded app layout.

Instead, the developer can implement the embedded app layout (as provided by the gem) once in the application.html.erb and set the

Now, when running the app both in embedded and non-embedded mode, the developer only has to maintain a single layout file for all controllers in its app.

To use the same application layout for every application controller, a developer can now overwrite the `#use_embedded_app_layout?` method.

```ruby
class ApplicationController
  # Ensures every controller is using the standard layouts/application.html.erb layout.
  # 
  # @return [true, false]
  def use_embedded_app_layout?
    false
  end
end
```

### Checklist

- [ ] ~~For security fixes, the [Disclosure Policy](https://github.com/Shopify/shopify_app/blob/master/SECURITY.md#disclosure-policy) must be followed.~~

### DIFF
--- a/docs/shopify_app/controller-concerns.md
+++ b/docs/shopify_app/controller-concerns.md
@@ -64,5 +64,26 @@ Implements Rails' [protect_from_forgery](https://api.rubyonrails.org/classes/Act
 #### EmbeddedApp
 If your ShopifyApp configuration has the `embedded_app` config set to true, [P3P header](https://www.w3.org/P3P/) and [content security policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) are handled for you.
 
+By default, the `EmbeddedApp` concern also sets the layout file to be `app/views/layouts/embedded_app.html.erb`.
+
+Sometimes one wants to run an embedded app in non-embedded mode. For example:
+
+- When the remote environment is a CI;
+- When the remote environment is a preview/PR app;
+- When the developer wants to run the app in a non-embedded mode for testing.
+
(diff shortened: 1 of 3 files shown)


matteodepalo:
Hi @stefanvermaas, thank you so much for your contribution! Would it be possible to add something to the documentation that mentions this new option?

stefanvermaas:
> Would it be possible to add something to the documentation that mentions this new option?

Yes, of course! I refrained from adding documentation initially as I wanted to see if there was an appetite for the Shopify team for this proposed change. I'll update the PR with accompanying documentation.

matteodepalo:
@stefanvermaas looks good! Could you rebase the branch?

matteodepalo (approved):

stefanvermaas:
Thanks @matteodepalo!

PR #1778, opened 2024-01-15
Add new configuration flag for token exchange

zzooeeyy:
For
- https://github.com/Shopify/develop-app-access/issues/147

### What this PR does
- Add new configuration flag to enable token exchange workflow
- We'll need to rename `wip_token_exchange` configuration once it's ready to be used. 

### Checklist

- [NA] Update `CHANGELOG.md` if the changes would impact users
- [NA] Update `README.md`, if appropriate.
- [NA] Update any relevant pages in `/docs`, if necessary
- [NA] For security fixes, the [Disclosure Policy](https://github.com/Shopify/shopify_app/blob/master/SECURITY.md#disclosure-policy) must be followed.


zzooeeyy:
> Works for me, though I don't love the wip in the config name. Could we say unstable instead?
> Also, I assume the actual functionality will come later? 😄

Yes, actual functionality will come later, that's why I'm inclined to name it "wip" since it's actually "work in progress", "unstable" to me means something that's already implemented and it's _partially_ working, just unstable..

PR #1788, opened 2024-02-08
Always register webhooks with offline sessions

### LINKED ISSUE #1708: Invalid API key or access token (unrecognized login or wrong password)
# Issue summary
After omniauth success and `ShopifyApp::WebhooksManagerJob` execute then I got error response
```
 {"errors":"[API] Invalid API key or access token (unrecognized login or wrong password)", "error_reference":"If you report this error, please include this id: a050f299-40bb-4604-8c82-69c78c9645ed."}
```

- `shopify_api` version: 13.1.0
- `shopify_app` version: 21.6.0
- Ruby version: 3.0.4
- Operating system: MacOS for local development, Amazon Linux for production

## Expected behavior

Install webhook without any error

## Actual behavior

Raise error from `ShopifyApp::WebhooksManagerJob` 

## Steps to reproduce the problem

1. Install app with Test your app
2. Omniauth successfully 
3. The error raised

paulomarg:
### What this PR does

Fixes #1708

Currently, we use the session we just created in the OAuth callback to trigger webhook registration. When using online tokens, that registration might not kick in before the token has expired.

To solve that, we can register webhooks using offline tokens, which won't expire.

- Install an app in a store, but don't run the webhook manager job
- Alter the user access token in the database
- Perform the job

Before this change, that would break. Now, it should work the same way.

### Things to focus on

1. Does this have any impact on performance?
2. Is it enough?

### Checklist

### DIFF
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -130,11 +130,11 @@ module ShopifyApp
     end
 
     def perform_post_authenticate_jobs(session)
-      # Ensure we use the offline session to install webhooks and scripttags
-      offline_session = session.online? ? shop_session : session
+      # Ensure we use the shop session to install webhooks and scripttags
+      session_for_shop = session.online? ? shop_session : session
 
-      install_webhooks(offline_session)
-      install_scripttags(offline_session)
+      install_webhooks(session_for_shop)
+      install_scripttags(session_for_shop)
 
       perform_after_authenticate_job(session)
     end
--- a/app/controllers/shopify_app/callback_controller.rb
+++ b/app/controllers/shopify_app/callback_controller.rb
@@ -130,8 +130,12 @@ module ShopifyApp
     end
 
     def perform_post_authenticate_jobs(session)
-      install_webhooks(session)
-      install_scripttags(session)
+      # Ensure we use the offline session to install webhooks and scripttags
+      offline_session = session.online? ? shop_session : session
+
+      install_webhooks(offline_session)
+      install_scripttags(offline_session)
+
       perform_after_authenticate_job(session)
     end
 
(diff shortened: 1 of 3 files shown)


test/controllers/callback_controller_test.rb
```diff
@@ -296,20 +289,40 @@ class CallbackControllerTest < ActionController::TestCase
...
+      shop_storage.store(@stubbed_session)
+
+      ShopifyApp::WebhooksManager.expects(:queue).with(SHOP_DOMAIN, "token")
+
+      get :callback, params: @callback_params
+      assert_response 302
+    rescue => e
+      shop_storage.clear
+      raise e
+    end
```
sle-c: i think using `ensure` is more suitable in this usecase

```suggestion
    ensure
      shop_storage.clear
    end
```

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -130,8 +130,12 @@ def user_access_scopes_strategy
     end
 
     def perform_post_authenticate_jobs(session)
-      install_webhooks(session)
-      install_scripttags(session)
+      # Ensure we use the offline session to install webhooks and scripttags
+      offline_session = session.online? ? shop_session : session
```
sle-c, line 134: i think the name `offline_session` is a bit confusing because I'd assume `shop_session` to be online but then it got assigned to `offline_session`. How about `present_session`?

sle-c: `current_session` might be used already

sle-c (approved):
minor comments, overall LGTM

app/controllers/shopify_app/callback_controller.rb
paulomarg: I went with `session_for_shop` as `shop_session` is the right name here, but it was taken 😄

PR #1789, opened 2024-02-09
add debug message to callback rescue

uurcank:
This adds a log at debug level to the callback rescue. It does not feel right to quietly rescue and redirect here.


paulomarg (approved):
Totally agreed. We should ideally never have errors silently going away.

PR #1794, opened 2024-02-19
Fix and Refactor splash page URL Handling to Respect Nested URLs

uurcank:
This PR addresses an inconsistency in the splash_page_with_params method within the controller concern, where the method previously defaulted to the application's root path, disregarding the nested URL setting [as explained here](https://github.com/Shopify/shopify_app/blob/main/docs/shopify_app/engine.md).

Previously, when a nested_url was enabled and the app was accessed from the Shopify dashboard or the page was refreshed, the user was incorrectly redirected back to the app's root URL, instead of the intended Shopify app root (the nested URL).

### DIFF
--- a/app/controllers/concerns/shopify_app/ensure_authenticated_links.rb
+++ b/app/controllers/concerns/shopify_app/ensure_authenticated_links.rb
@@ -18,7 +18,7 @@ module ShopifyApp
         embedded: params[:embedded],
       )
     end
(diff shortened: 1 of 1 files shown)


paulomarg (commented):
This is great, thank you for your contribution! If you could please fix up the rubocop error, we'll get this merged.

PR #1284, opened 2021-06-15
Packaging for release 18.0.2

jesalerno84:
18.0.2 (Jun 15, 2021)
----------
* Added careers link to readme. [#1274](https://github.com/Shopify/shopify_app/pull/12274)

### Checklist

Before submitting the PR, please consider if any of the following are needed:

- [x] Update `CHANGELOG.md` if the changes would impact users
- [x] Update `README.md`, if appropriate.
- [x] Update any relevant pages in `/docs`, if necessary
- [x] For security fixes, the [Disclosure Policy](https://github.com/Shopify/shopify_app/blob/master/SECURITY.md#disclosure-policy) must be followed.

PR #1801, opened 2024-03-05
bump dependency to latest major version of the API

nelsonwittwer:
### What this PR does

Bumps the dependency on the `shopify_api` gem to the latest major version that was recently released. For those looking to install the latest version of the version of the ruby API there was a bundler dependency error as documented in [shopify_app #1284](https://github.com/Shopify/shopify-api-ruby/issues/1284)

This also follows the direction of the API gem and requires Ruby 3.0 or greater as 2.7 is no longer supported.

To verify this worked as expected I used one of my test apps and installed the latest version of the API and pointed to this local branch:

```ruby
gem "shopify_api", "~> 14.0"
gem "shopify_app", path: "../../shopify_app"
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 Unreleased
 ----------
+* ⚠️ [Breaking] Bumps minimum supported Ruby version to 3.0. Bumps `shopify_api` to 14.0 [1801](https://github.com/Shopify/shopify_app/pull/1801)
 * Make type param for webhooks route optional. This will fix a bug with CLI initiated webhooks.[1786](https://github.com/Shopify/shopify_app/pull/1786)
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     name: Ruby ${{ matrix.version }}
     strategy:
       matrix:
-        version: ['2.7', '3.0', '3.1', '3.2']
+        version: ['3.0', '3.1', '3.2']
```
paulomarg, line 15: I think this means we're also dropping v2 support from this gem, which would make it a breaking change right? Should we also update the Gemfile / add a changelog entry to that effect?

paulomarg (approved):
Just the one nit, but makes sense to me.

PR #1814, opened 2024-03-12
Packaging for release 22.00.1

lizkenyon:


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 Unreleased
 ----------
+
+22.00.1 (March 12, 2024)
```
paulomarg, line 4: Just wondering - is the double `00` expected? I don't think it matters but just curious why it would let that happen :)

Eianaftyio: No 😞

Eianaftyio: @bjk

PR #1819, opened 2024-03-26
Extract post authenticate tasks from CallbackController

zzooeeyy:
### What this PR does
- https://github.com/Shopify/temp-project-mover-Archetypically-20250319145332/issues/252

- Extract common logic to run post authenticate task from CallbackController so that it can be used after the token exchange flow. 
- Added PostAuthenticateTasks to run the authenticated task
- This is configurable  as `config.custom_post_authenticate_tasks`, if that's not set, the default `config.post_authenticate_tasks` will return `PostAuthenticateTasks` class
- Updated docs on post authenticate tasks
- Added deprecation warnings

##### Deprecation warning in server start from configuration
deprecation

##### Deprecation reminder in tests logs
test

##### When `config.custom_post_authenticate_tasks` is set, it will be used to handle post authenticate tasks
post-authenticate-jobs

- I also tested not setting the custom_post_authenticate_tasks to ensure the methods in CallbackController to perform post authenticate tasks are done. 

### Checklist

### DIFF
--- a/docs/shopify_app/authentication.md
+++ b/docs/shopify_app/authentication.md
@@ -12,7 +12,7 @@ See [*Getting started with session token authentication*](https://shopify.dev/do
 
 * [OAuth callback](#oauth-callback)
   * [Customizing callback controller](#customizing-callback-controller)
-* [Run jobs after the OAuth flow](#run-jobs-after-the-oauth-flow)
+* [Run jobs after the OAuth flow](#post-authenticate-tasks)
 * [Rotate API credentials](#rotate-api-credentials)
 * [Making authenticated API requests after authorization](#making-authenticated-api-requests-after-authorization)
 


lib/shopify_app/configuration.rb
```diff
@@ -29,6 +29,9 @@ class Configuration
     attr_writer :login_callback_url
     attr_accessor :embedded_redirect_url
 
+    # customize post authenticate tasks
+    attr_accessor :custom_post_authenticate_tasks
```
paulomarg, line 33: Nit: would it make more sense to have a `post_authenticate_tasks` setting that defaults to the new class rather than a `custom` setting?

test/controllers/callback_controller_test.rb
```diff
@@ -333,6 +352,32 @@ class CallbackControllerTest < ActionController::TestCase
       assert_response 302
     end
 
+    test "#callback calls post_authenticate_tasks if custom_post_authenticate_tasks is set" do
```
paulomarg, line 355: Should we have a very simple test here to ensure that the default behaviour gets called (even if we have to stub it)?

test/shopify_app/configuration_test.rb
```diff
@@ -2,6 +2,18 @@
 
 require "test_helper"
 
+module Shopify
+  class CustomPostAuthenticateTasks
```
paulomarg, line 6: This is just a question for my own benefit: will any class that defines a `perform` method do here? Should we check if it extends a job class instead to standardize things a bit more?

paulomarg (commented):
Had a few thoughts but most of them are just me trying to understand things better so take them with a grain of salt :)

lib/shopify_app/configuration.rb
zzooeeyy: yess that does make sense, but in this pre-deprecation case we _don't_ want to use the post_authenticate_tasks with a default if it's not manually set since we still want to call the old methods. I think it does make sense to have a single config with a default once we roll out the deprecation.

test/shopify_app/auth/post_authenticate_tasks_test.rb
```diff
@@ -0,0 +1,128 @@
...
+    ShopifyApp::WebhooksManager.expects(:queue).with(SHOP_DOMAIN, "offline_token")
+
+    ShopifyApp::Auth::PostAuthenticateTasks.perform(@offline_session)
+  end
+
+  test "#perform doesn't run the WebhooksManager if no webhooks are configured" do
+    ShopifyApp.configure do |config|
+      config.webhooks = []
+    end
+    ShopifyApp::WebhooksManager.add_registrations
```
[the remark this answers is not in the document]
zzooeeyy: ooops that's my oversight, I copied straight from CallbackControllTest LOL I'll also remove it from that test case

https://github.com/Shopify/shopify_app/blob/main/test/controllers/callback_controller_test.rb#L180

test/shopify_app/configuration_test.rb
paulomarg: I also don't think it matters too much, but I remember the documentation said it could "run this later", which I took to mean it would deal with a `perform_later` type of thing. But just for the sake of standardization, I don't think it's necessary.

app/controllers/shopify_app/callback_controller.rb
```diff
@@ -23,7 +23,16 @@ def callback
 
       return respond_with_user_token_flow if start_user_token_flow?(api_session)
 
-      perform_post_authenticate_jobs(api_session)
+      if ShopifyApp::VERSION < "23.0"
```
paulomarg, line 26: Nice that we can just compare it this way, makes this very easy to understand :)

docs/Upgrading.md
```diff
@@ -40,6 +40,17 @@ We also recommend the use of a staging site which matches your production enviro
...
 
+## Unreleased
+#### (v23.0.0) - Deprecated methods in CallbackController
+The following methods from `ShopifyApp::CallbackController` have been deprecated in `v23.0.0`
+- `perform_after_authenticate_job`
+- `install_webhooks`
+- `perform_post_authenticate_jobs`
+
+If you have overwritten these methods in your callback controller to modify the behavior of the inherited `CallbackController`, you will need to
+update your app to use configurable option `config.custom_post_authenticate_tasks` instead. See [post authenticate tasks](https://github.com/Shopify/shopify_app/blob/main/docs/shopify_app/authentication.md#post-authenticate-tasks)
```
paulomarg, line 51: Tip: you can use relative links here to avoid running into absolute URL errors:

```suggestion
update your app to use configurable option `config.custom_post_authenticate_tasks` instead. See [post authenticate tasks](/docs/shopify_app/authentication.md#post-authenticate-tasks)
```

lib/shopify_app/configuration.rb
paulomarg: Hmm yeah, fair. It wouldn't be ideal to change the setting name in the future though - so I think we should stick with what we add in this PR, whichever way we prefer to do it.

lib/shopify_app/configuration.rb
```diff
@@ -126,6 +131,48 @@ def user_access_scopes
...
+    end
+
+    private
+
+    def log_callback_controller_method_deprecation
+      return if Rails.env.test?
+
+      # TODO: Remove this before releasing v23.0.0
+      message = <<~EOS
+        ================================================
```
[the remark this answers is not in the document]
paulomarg: Fair. I think this comment is big enough that most folks would see it.

When I suggested we log this conditionally, I meant that ideally we wouldn't log it if there is no usage of the deprecated setup, i.e. if the app is configured in the "new" way, we just skip this to avoid confusion.

test/controllers/callback_controller_test.rb
```diff
@@ -47,6 +54,7 @@ class CallbackControllerTest < ActionController::TestCase
         "AppleWebKit/537.36 (KHTML, like Gecko)"\
         "Chrome/69.0.3497.100 Safari/537.36"
       ShopifyApp::SessionRepository.stubs(:store_session)
+      ShopifyApp.configuration.custom_post_authenticate_tasks = nil
```
paulomarg, line 57: Should this be the global test config value, so that other tests start with this as an assumption?

zzooeeyy: Not necessary I only did it to "reset" the configuration state, but I can move that to the teardown so it doesn't become and assumption for other tests!

paulomarg (approved):
Just a general question, but if this is the way, LGTM

PR #1822, opened 2024-04-02
Retrying block if encountered 401 error

zzooeeyy:
- https://github.com/Shopify/develop-app-access/issues/146

📹 [Demo video](https://videobin.shopify.io/v/Xr28Lg)

### What this PR does
- Handles 401 error in `around_action` in token exchange concern `activate_shopfiy_session`. 
- It'll retry once by performing token exchange to retrieve new access token. 
- Subsequent errors will be thrown..

⚠️ One caveat is that if the app already rescues errors in their controller action, then it won't be retried since the error will be caught in the yield block (in the controller action).. 
```ruby
# my_controller.rb

# Will retry:
def count
  ShopifyAPI::Product.count
end

# Will not retry
def count_and_rescue
  ShopifyAPI::Product.count
rescue =>
  # not raising again
end
```
### Things to focus on

1. 
2. 
3. 

### Checklist


rachel-carvalho:
We'll create an API wrapper and handle errors differently, so I'm moving this back to draft.

lib/shopify_app/admin_api/with_token_refetch.rb
```diff
@@ -0,0 +1,36 @@
...
+
+module ShopifyApp
+  module AdminAPI
+    module WithTokenRefetch
+      def with_token_refetch(session, session_token)
+        retrying = false if retrying.nil?
+        yield
+      rescue ShopifyAPI::Errors::HttpResponseError => error
+        if error.code == 401 && !retrying
+          retrying = true
```
zzooeeyy, line 11: Does it make more sense to move the delete session logic to here instead of the concern? If apps use this wrapper method and it failed, it the session should be deleted right?? 🤔 
```
ShopifyApp::SessionRepository.delete_session(session.id)
```

zzooeeyy (commented):
Thanks for the clean up, I just have some questions! ❤️

lib/shopify_app/admin_api/with_token_refetch.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      def copy_session_attributes(from:, to:)
+        to.shop = from.shop
+        to.state = from.state
+        to.access_token = from.access_token
+        to.scope = from.scope
+        to.associated_user_scope = from.associated_user_scope
+        to.expires = from.expires
+        to.associated_user = from.associated_user
+        to.shopify_session_id = from.shopify_session_id
```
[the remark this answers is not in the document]
paulomarg: Yes, I agree - the session should be able to clone itself if that's something we'll want to do.

lib/shopify_app/admin_api/with_token_refetch.rb
```diff
@@ -0,0 +1,36 @@
...
+    module WithTokenRefetch
+      def with_token_refetch(session, session_token)
+        retrying = false if retrying.nil?
+        yield
+      rescue ShopifyAPI::Errors::HttpResponseError => error
+        if error.code == 401 && !retrying
+          retrying = true
+          ShopifyApp::Logger.debug("Encountered 401 error, exchanging token and retrying with new access token")
+          new_session = ShopifyApp::Auth::TokenExchange.perform(session_token)
+          copy_session_attributes(from: new_session, to: session)
```
paulomarg, line 14: Should we just reassign `session = new_session` instead, and completely override it?

lib/shopify_app/controller_concerns/token_exchange.rb
```diff
@@ -3,38 +3,43 @@
...
+      retrieve_session_from_token_exchange if current_shopify_session.blank? || should_exchange_expired_token?
 
       begin
         ShopifyApp::Logger.debug("Activating Shopify session")
         ShopifyAPI::Context.activate_session(current_shopify_session)
-        yield
+        with_token_refetch(current_shopify_session, session_token, &block)
+      rescue ShopifyAPI::Errors::HttpResponseError => error
+        if error.code == 401
+          ShopifyApp::Logger.debug("Admin API returned a 401 Unauthorized error, deleting current access token.")
```
paulomarg, line 17: This might also happen for the SFAPI with `unauthenticated_` scopes, right?

lib/shopify_app/controller_concerns/token_exchange.rb
```diff
@@ -46,64 +51,13 @@ def current_shopify_domain
...
-          ShopifyApp::SessionRepository.store_session(session)
-        rescue ActiveRecord::RecordNotUnique
-          ShopifyApp::Logger.debug("Session not stored due to concurrent token exchange calls")
-        end
-      end
-
-      session
+      ShopifyApp::Auth::TokenExchange.perform(session_token)
+    rescue ShopifyAPI::Errors::InvalidJwtTokenError
+      # respond_to_invalid_session_token
```
paulomarg, line 60: This is silently ignoring this error - is that something we want to do?

test/shopify_app/auth/token_exchange_test.rb
```diff
@@ -0,0 +1,175 @@
...
+    @offline_session = build_offline_session
+    @online_session = build_online_session
+  end
+
+  test "#perform exchanges offline token then stores it when only shop session store is configured" do
+    ShopifyAPI::Auth::TokenExchange.expects(:exchange_token).with(
+      shop: @shop,
+      session_token: @session_token,
+      requested_token_type: OFFLINE_ACCESS_TOKEN_TYPE,
+    ).returns(@offline_session)
```
paulomarg, line 26: Do we have to explicitly say we only expect one call here?

paulomarg (commented):
I think this is great! We should also make sure to document the new public-facing aspects of these changes - ideally with some examples of how folks might use them.

lib/shopify_app/admin_api/with_token_refetch.rb
rachel-carvalho: We need to update the same instance because of possible subsequent calls in the same request that might be using it. Performing token exchange saves the new session to the db but in memory we still hold the original values.

An example would be to fetch a product, then updating it, performing 2 calls. If the app does this in separate places with separate `with_token_refetch` calls, it would end up causing token exchange to happen twice, because `current_active_session` (or even just their `session` variable) would still contain the old access token with its old expiry, scopes etc when it performs the second API call.

test/shopify_app/auth/token_exchange_test.rb
rachel-carvalho: Mocha's `expects` fails if not called [exactly once](https://mocha.jamesmead.org/Mocha/Mock.html#expects-instance_method), but there's also a `.once` method for readability which we can use to make it more explicit.

docs/shopify_app/sessions.md
```diff
@@ -20,14 +20,15 @@ Sessions are used to make contextual API calls for either a shop (offline sessio
...
+      - [Re-fetching an access token when API returns Unauthorized](#re-fetching-an-access-token-when-api-returns-unauthorized)
   - [Access scopes](#access-scopes)
     - [`ShopifyApp::ShopSessionStorageWithScopes`](#shopifyappshopsessionstoragewithscopes)
     - [`ShopifyApp::UserSessionStorageWithScopes`](#shopifyappusersessionstoragewithscopes)
   - [Migrating from shop-based to user-based token strategy](#migrating-from-shop-based-to-user-based-token-strategy)
   - [Migrating from `ShopifyApi::Auth::SessionStorage` to `ShopifyApp::SessionStorage`](#migrating-from-shopifyapiauthsessionstorage-to-shopifyappsessionstorage)
 
 ## Sessions
-#### Types of session tokens
+#### Types of session access tokens
```
zzooeeyy, line 31: I think you'll need to update the md link here too
https://github.com/Shopify/shopify_app/pull/1822/files#diff-f65ce70479de09e34a7129d47ee16835650bc8e6e4a625ff6431ab7dc804a092R11

docs/shopify_app/sessions.md
```diff
@@ -203,6 +204,66 @@ user.with_shopify_session do
...
+
+If the error is being rescued in the action, it's still possible to perform token exchange and retry again with the `with_token_refetch` method provided by `EnsureHasSession`.
+
+```ruby
+class MyController < ApplicationController
+  include ShopifyApp::EnsureHasSession
+
+  def index
+    client = ShopifyAPI::Clients::Graphql::Admin.new(session: current_shopify_session)
+    with_token_refetch(current_shopify_session, session_token) do # Unauthorized errors raised in this block will initiate token exchange and the block will be retried once
```
zzooeeyy, line 230: hmm looking at the example here, do you think it might be confusing to devs on the difference between `current_shopify_session` and `session_token`? There are so many cases where we just use "session" to describe something, should we rename `session_token` to be like `session_jwt_token` to be more explicit on which one we're talking about?

docs/shopify_app/sessions.md
```diff
@@ -203,6 +204,66 @@ user.with_shopify_session do
 end
 ```
 
+#### Re-fetching an access token when API returns Unauthorized
```
zzooeeyy, line 207: Do we need to add a big disclaimer (at least for now until we have a new section for the new auth strategy) that this is only applicable for the new embedded auth strategy?

🤔 And should we actually park this doc change until we're ready to release the new auth strategy? and add it when we add the "release" docs describing this new strategy? Not sure how often this gem gets released, but it might be weird showing docs with features that hasn't been released yet?

lib/shopify_app/admin_api/with_token_refetch.rb
```diff
@@ -7,29 +7,21 @@ def with_token_refetch(session, session_token)
...
           retry
-        else
-          ShopifyApp::Logger.debug("Encountered error: #{error.code} - #{error.response.inspect}, re-raising")
-          raise
         end
-      end
-
-      private
-
-      def copy_session_attributes(from:, to:)
```
zzooeeyy: I wonder if we keep this for now with so the CI can pass until a new version of shopify-ruby-api is released? then we can clean it up?

zzooeeyy (commented):
Thanks! Some more 💭

docs/shopify_app/sessions.md
```diff
@@ -203,6 +204,66 @@ user.with_shopify_session do
...
+
+When using `ShopifyApp::EnsureHasSession` and the `new_embedded_auth_strategy` configuration, any **unhandled** Unauthorized `ShopifyAPI::Errors::HttpResponseError` will cause the app to perform token exchange to fetch a new access token from Shopify and the action to be retried once.
+
+```ruby
+class MyController < ApplicationController
+  include ShopifyApp::EnsureHasSession
+
+  def index
+    client = ShopifyAPI::Clients::Graphql::Admin.new(session: current_shopify_session)
+    client.query(options) # when this raises an Unauthorized error from Shopify, EnsureHasSession will retry the action once after performing token exchange
```
paulomarg, line 217: Total nit, but having the comment after the line makes it a bit funky to read in MD mode (even at max width) because it's such a long line. Could we move them up / break them so it's easier to read without scrolling?

image

docs/shopify_app/sessions.md
paulomarg, line 230: Also, should we call out that the session will be updated if it does refetch it?

test/shopify_app/auth/token_exchange_test.rb
```diff
@@ -0,0 +1,177 @@
...
+
+    ShopifyAPI::Auth::TokenExchange.expects(:exchange_token).with(
+      shop: @shop,
+      session_token: @session_token,
+      requested_token_type: ONLINE_ACCESS_TOKEN_TYPE,
+    ).returns(@online_session)
+
+    assert_nil ShopifyApp::SessionRepository.load_session(@online_session.id)
```
paulomarg, line 53: Nit: should we check that the offline session _also_ got stored here?

test/shopify_app/controller_concerns/token_exchange_test.rb
```diff
@@ -54,58 +63,28 @@ class TokenExchangeControllerTest < ActionController::TestCase
...
+        ShopifyApp::SessionRepository.store_session(@offline_session)
+      end
 
       ShopifyAPI::Context.expects(:activate_session).with(@offline_session)
 
       get :index, params: { shop: @shop }
     end
   end
 
-  test "Exchanges online token if user session store is configured" do
```
paulomarg, line 81: Should we keep this test? This logic is still in the token exchange class, right?

paulomarg (commented):
Had some nits but looking really good!

docs/shopify_app/sessions.md
rachel-carvalho: Yeah I agree it's very confusing. I noticed we call it `id_token` in the URL and also in [App Bridge](https://shopify.dev/docs/api/app-bridge-library/apis/id-token), so I'm leaning towards `id_token` here to keep consistence, especially because those are the two sources of the token we're using here.

test/shopify_app/auth/token_exchange_test.rb
ragalie: Came here to say the same thing :)

ragalie (approved):
LGTM, thanks!

paulomarg (approved):
Just one last nit :)

lib/shopify_app/controller_concerns/token_exchange.rb
```diff
@@ -46,75 +45,22 @@ def current_shopify_domain
...
-          ShopifyApp::SessionRepository.store_session(session)
-        rescue ActiveRecord::RecordNotUnique
-          ShopifyApp::Logger.debug("Session not stored due to concurrent token exchange calls")
-        end
-      end
-
-      session
+      @current_shopify_session = nil
+      ShopifyApp::Auth::TokenExchange.perform(shopify_id_token)
+      # TODO: Rescue JWT validation errors when bounce page is ready
```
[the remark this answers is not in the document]
zzooeeyy: Yesss

PR #1823, opened 2024-04-03
Handle invalid session token

zzooeeyy:
- closes https://github.com/Shopify/develop-app-access/issues/145

### What this PR does
- Handle invalid session token errors 
- Redirect to app-bridge-next bounce page if session token is invalid & Auth header doesn't exist
- Respond with 401 if auth header exists and session token is invalid. If the app's front-end is using CDN version of app bridge next, it'll automatically be retried. otherwise, the 401 error will be exposed.

### Tophatting-

* 📹 [Redirecting to bounce page on initial app render](https://videobin.shopify.io/v/5rOD2g) (we are not parsing `id_token` from URL param yet, so session token will be invalid initially, this redirect shouldn't be hit as often once we are using `id_token`)  
* For testing - I created a temp global variable count that raises an "invalid session token" error on **every other try**. 
  * 📹 [App configured with app bridge from CDN](https://videobin.shopify.io/v/rzQElp) - The initial `create` failed from invalid session token, since `X-Shopify-Retry-Invalid-Session-Request` is set from the response, app bridge will retry `create` with new session token, thus passing the second time
  * 📹 [App configured with old version of app bridge NOT using the CDN](https://videobin.shopify.io/v/bjkoE1) - Response for `create` returns 401 in every other request from my test setup.. 

### Checklist

- [ will update once feature is released ] Update `CHANGELOG.md` if the changes would impact users
- [will update once feature is released  ] Update `README.md`, if appropriate.
- [will update once feature is released  ] Update any relevant pages in `/docs`, if necessary

### DIFF
--- a/app/controllers/concerns/shopify_app/ensure_installed.rb
+++ b/app/controllers/concerns/shopify_app/ensure_installed.rb
@@ -17,7 +17,11 @@ module ShopifyApp
 
       before_action :check_shop_domain
 
-      unless ShopifyApp.configuration.use_new_embedded_auth_strategy?
+      if ShopifyApp.configuration.use_new_embedded_auth_strategy?
+        include ShopifyApp::TokenExchange
+        include ShopifyApp::EmbeddedApp
+        around_action :activate_shopify_session
+      else
         # TODO: Add support to use new embedded auth strategy here when invalid
         # session token can be handled by AppBridge app reload
         before_action :check_shop_known
--- a/app/controllers/concerns/shopify_app/ensure_installed.rb
+++ b/app/controllers/concerns/shopify_app/ensure_installed.rb
@@ -22,8 +22,6 @@ module ShopifyApp
         include ShopifyApp::EmbeddedApp
         around_action :activate_shopify_session
       else
-        # TODO: Add support to use new embedded auth strategy here when invalid
-        # session token can be handled by AppBridge app reload
         before_action :check_shop_known
         before_action :validate_non_embedded_session
       end
--- a/lib/shopify_app/controller_concerns/token_exchange.rb
+++ b/lib/shopify_app/controller_concerns/token_exchange.rb
@@ -5,8 +5,17 @@ module ShopifyApp
     extend ActiveSupport::Concern
     include ShopifyApp::AdminAPI::WithTokenRefetch
 
+    INVALID_SHOPIFY_ID_TOKEN_ERRORS = [
+      ShopifyAPI::Errors::CookieNotFoundError,
+      ShopifyAPI::Errors::InvalidJwtTokenError,
+    ].freeze
+
     def activate_shopify_session(&block)
-      retrieve_session_from_token_exchange if current_shopify_session.blank? || should_exchange_expired_token?
+      begin
+        retrieve_session_from_token_exchange if current_shopify_session.blank? || should_exchange_expired_token?
+      rescue *INVALID_SHOPIFY_ID_TOKEN_ERRORS
+        return respond_to_invalid_shopify_id_token
+      end
 
       begin
         ShopifyApp::Logger.debug("Activating Shopify session")
@@ -47,9 +56,6 @@ module ShopifyApp
     def retrieve_session_from_token_exchange
       @current_shopify_session = nil
       ShopifyApp::Auth::TokenExchange.perform(shopify_id_token)
-      # TODO: Rescue JWT validation errors when bounce page is ready
-      # rescue ShopifyAPI::Errors::InvalidJwtTokenError
-      #   respond_to_invalid_shopify_id_token
     end
 
     def shopify_id_token
@@ -61,23 +67,26 @@ module ShopifyApp
     end
 
     def respond_to_invalid_shopify_id_token
-      # TODO: Implement this method to handle invalid session tokens
+      return redirect_to_bounce_page if request.headers["HTTP_AUTHORIZATION"].blank?
 
-      # if request.xhr?
-      # response.set_header("X-Shopify-Retry-Invalid-Session-Request", 1)
-      # unauthorized_response = { message: :unauthorized }
-      # render(json: { errors: [unauthorized_response] }, status: :unauthorized)
-      # else
-      # patch_session_token_url = "#{ShopifyAPI::Context.host}/patch_session_token"
-      # patch_session_token_params = request.query_parameters.except(:id_token)
+      response.set_header("X-Shopify-Retry-Invalid-Session-Request", 1)
+      unauthorized_response = { message: :unauthorized }
+      render(json: { errors: [unauthorized_response] }, status: :unauthorized)
+    end
 
-      # bounce_url = "#{ShopifyAPI::Context.host}#{request.path}?#{patch_session_token_params.to_query}"
+    def redirect_to_bounce_page
+      patch_session_token_url = "#{ShopifyApp.configuration.root_url}/patch_session_token"
+      patch_session_token_params = request.query_parameters.except(:id_token)
 
-      # # App Bridge will trigger a fetch to the URL in shopify-reload, with a new session token in headers
-      # patch_session_token_params["shopify-reload"] = bounce_url
+      bounce_url = "#{request.path}?#{patch_session_token_params.to_query}"
 
-      # redirect_to("#{patch_session_token_url}?#{patch_session_token_params.to_query}", allow_other_host: true)
-      # end
+      # App Bridge will trigger a fetch to the URL in shopify-reload, with a new session token in headers
+      patch_session_token_params["shopify-reload"] = bounce_url
+
+      redirect_to(
+        "#{patch_session_token_url}?#{patch_session_token_params.to_query}",
+        allow_other_host: true,
+      )
     end
 
     def online_token_configured?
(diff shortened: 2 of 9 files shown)


lib/shopify_app/controller_concerns/token_exchange.rb
```diff
@@ -5,8 +5,18 @@ module TokenExchange
...
+    INVALID_SHOPIFY_ID_TOKEN_ERRORS = [
+      ShopifyAPI::Errors::CookieNotFoundError,
+      ShopifyAPI::Errors::InvalidJwtTokenError,
+    ].freeze
+
     def activate_shopify_session(&block)
-      retrieve_session_from_token_exchange if current_shopify_session.blank? || should_exchange_expired_token?
+      begin
+        retrieve_session_from_token_exchange if current_shopify_session.blank? || should_exchange_expired_token?
+      rescue *INVALID_SHOPIFY_ID_TOKEN_ERRORS => e
```
rachel-carvalho, line 16: Could we rescue these errors for the entire `activate_shopify_session` method? `with_token_refetch` might also end up raising this error, since it performs token exchange.

test/routes/sessions_routes_test.rb
```diff
@@ -25,6 +25,10 @@ class SessionsRoutesTest < ActionController::TestCase
     assert_routing "/logout", { controller: "shopify_app/sessions", action: "destroy" }
   end
 
+  test "patch_shopify_id_token routes to sessions#patch_shopify_id_token" do
```
rachel-carvalho, line 28: Nit: this file is testing `session` routes (in the sense of log in/log out/`request.session`, not session token I guess 😅). Maybe we could create a new file?

test/shopify_app/controller_concerns/token_exchange_test.rb
```diff
@@ -178,13 +183,48 @@ class TokenExchangeControllerTest < ActionController::TestCase
     end
   end
 
+  [
+    ShopifyAPI::Errors::InvalidJwtTokenError,
+    ShopifyAPI::Errors::CookieNotFoundError,
```
rachel-carvalho, line 188: Maybe we shouldn't respond to invalid cookies with a bounce page now that we'll be using the method that exclusively checks `shopify_id_token`.

test/shopify_app/controller_concerns/token_exchange_test.rb
```diff
@@ -178,13 +183,48 @@ class TokenExchangeControllerTest < ActionController::TestCase
...
+  ].each do |invalid_shopify_id_token_error|
+    test "Redirects to bounce page if Shopify ID token is invalid with #{invalid_shopify_id_token_error}" do
+      ShopifyApp.configuration.root_url = "/my-root"
+      ShopifyAPI::Utils::SessionUtils.stubs(:current_session_id).raises(invalid_shopify_id_token_error)
+      request.headers["HTTP_AUTHORIZATION"] = nil
+
+      params = { shop: @shop, my_param: "for-keeps", id_token: "dont-include-this-id-token" }
+      expected_redirect_url = "/my-root/patch_shopify_id_token?my_param=for-keeps&shop=my-shop.myshopify.com"
+      expected_redirect_url += "&shopify-reload=%2Freloaded_path%3Fmy_param%3Dfor-keeps%26shop%3Dmy-shop.myshopify.com"
```
rachel-carvalho, line 197: Nit: maybe we could URL encode the param for better readability of the test?
```suggestion
      reload_url = CGI.escape("/reloaded_path?my_param=for-keeps&shop=#{@shop}")
      expected_redirect_url = "/my-root/patch_shopify_id_token?my_param=for-keeps&shop=my-shop.myshopify.com&shopify-reload=#{reload_url}"
```

lib/shopify_app/controller_concerns/token_exchange.rb
```diff
@@ -61,23 +68,28 @@ def id_token_header
...
 
-      # # App Bridge will trigger a fetch to the URL in shopify-reload, with a new session token in headers
-      # patch_session_token_params["shopify-reload"] = bounce_url
+      # App Bridge will trigger a fetch to the URL in shopify-reload, with a new session token in headers
+      patch_shopify_id_token_params["shopify-reload"] = bounce_url
 
-      # redirect_to("#{patch_session_token_url}?#{patch_session_token_params.to_query}", allow_other_host: true)
-      # end
+      redirect_to(
+        "#{patch_shopify_id_token_url}?#{patch_shopify_id_token_params.to_query}",
```
rachel-carvalho, line 90: I wish there was a way to use the Rails URL helper method `patch_shopify_id_token_path` here which already does most of the URL assembly work for us, but the way our tests are setup using `with_routing` it's not including engine's routes, so I think it would be a bit too much work for it to be worth it.

lib/shopify_app/controller_concerns/token_exchange.rb
```diff
@@ -5,8 +5,18 @@ module TokenExchange
     extend ActiveSupport::Concern
     include ShopifyApp::AdminAPI::WithTokenRefetch
 
+    INVALID_SHOPIFY_ID_TOKEN_ERRORS = [
+      ShopifyAPI::Errors::CookieNotFoundError,
```
paulomarg, line 9: Would we ever have a cookie present for token exchange? If it's only meant to run for embedded apps there should never be any cookies involved, right?

paulomarg (approved):

app/controllers/concerns/shopify_app/ensure_installed.rb
```diff
@@ -17,9 +17,11 @@ module EnsureInstalled
 
       before_action :check_shop_domain
 
-      unless ShopifyApp.configuration.use_new_embedded_auth_strategy?
-        # TODO: Add support to use new embedded auth strategy here when invalid
-        # session token can be handled by AppBridge app reload
+      if ShopifyApp.configuration.use_new_embedded_auth_strategy?
+        include ShopifyApp::TokenExchange
+        include ShopifyApp::EmbeddedApp
+        around_action :activate_shopify_session
```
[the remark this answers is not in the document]
zzooeeyy: Yess, so `EnsureInstalled` concern itself implements the [installed_shop_session ](https://github.com/Shopify/shopify_app/blob/handle-invalid-session-token/app/controllers/concerns/shopify_app/ensure_installed.rb#L41-L43) method, so it'll still be able to be used. Having TokenExchange enabled, they'll just get both 'current_shopify_session` and `installed_shop_session` !

test/routes/sessions_routes_test.rb
zzooeeyy: hmmm I think it belongs here since it's just testing the routes, and we did add the new route to patch session token 🤔

lib/shopify_app/controller_concerns/token_exchange.rb
zzooeeyy: Yea cookies are never involved.. but the way `ShopifyAPI::current_session_id` is setup, it'll throw a "CookieNotFoundError" if we pass in a nil auth header (which will be the case for the initial server load when we're not using id_token from URL params), so once I change this to use the new util method, we won't be catching this `CookieNotFoundError` but `MissingJwtTokenError` instead.

zzooeeyy: Yea exactly :(

lib/shopify_app/controller_concerns/token_exchange.rb
zzooeeyy: I was pairing with Mike and we thought that's not worth it since we already check for validity before we call token exchange, it'd just be adding extra work for the slightest chance that it might expire within that timeframe... There are leeways in-place to ensure we don't encounter these problems..?

rachel-carvalho: Oh but what if we don't perform token exchange there because `current_shopify_session` exists and then do during `with_token_refetch`? In that case it wouldn't be double validation/rescue, right?

test/routes/sessions_routes_test.rb
rachel-carvalho: I just mean there are many files testing routes like `test/routes/callback_routes_test.rb` and `test/routes/webhooks_routes_test.rb`, each testing different route groups. The `session_routes_test.rb` file is testing login/logout, so I think we _could_ have an `id_token_routes_test.rb`, wdyt?

lib/shopify_app/controller_concerns/token_exchange.rb
zzooeeyy: hmmm I think this will cause a "Double render/redirect" error since this is around the controller action.. so if we handle the error and respond with retry or redirect to bounce page and the controller action also renders... 🤔

test/routes/sessions_routes_test.rb
rachel-carvalho: Oh sorry I missed it was in the same controller. I think they're different but not sure if it warrants its own controller, never mind this!

lib/shopify_app/controller_concerns/token_exchange.rb
rachel-carvalho: If the action raises a token exchange error it will likely not have rendered yet, but you're right that it _could_ happen if the app doesn't follow the pattern of rendering/redirecting as the last thing in the action 🤔 

I looked up and saw there's a `performed?` [method](https://api.rubyonrails.org/classes/ActionController/Metal.html#method-i-performed-3F) we can call in the controller to avoid a double render/redirect. What if we bounce unless performed?

PR #1610, opened 2022-12-13
Remove Logged output for rescued JWT exceptions 

klenotiw:
### What this PR does

fixes: https://github.com/Shopify/shopify_app/issues/1567

We decided to remove this logged output. If we feel like we need it again we should consider putting it in a lower log level like `:debug` or `:info` so it would be easier for devs to turn off.

### Things to focus on

1. Do we like this approach?

### Checklist

Before submitting the PR, please consider if any of the following are needed:

- [x] Update `CHANGELOG.md` if the changes would impact users
- [x] Update `README.md`, if appropriate.
- [x] Update any relevant pages in `/docs`, if necessary
- [x] For security fixes, the [Disclosure Policy](https://github.com/Shopify/shopify_app/blob/master/SECURITY.md#disclosure-policy) must be followed.

PR #1832, opened 2024-04-22
Support id_token from URL params

zzooeeyy:
- https://github.com/Shopify/develop-app-access/issues/289

### What this PR does
* Bumps `shopify_api` to `14.3.0`
* Support `id_token` from URL param 
  * Extracted controller concern `WithShopifyIdToken`
      * This concern provides a method `shopify_id_token` to retrieve the Shopify Id token from either the authorization header or the URL param `id_token`.
  * `ShopifyApp::JWTMiddleware` supports retrieving session token from URL param `id_token`
  * `ShopifyApp::JWTMiddleware` returns early if the app is not embedded to avoid unnecessary JWT verification
  * `LoginProtection` now uses `WithShopifyIdToken` concern to retrieve the Shopify Id token, thus accepting the session token from the URL param `id_token`
* Marking `ShopifyApp::JWT` to be deprecated in version 23.0.0, use `ShopifyAPI::Auth::JwtPayload` instead.

- Ensure existing behaviour remains the same besides accepting id_token from param. Apps shouldn't raise more errors from JWT parsing.
- 📹 [LoginProtection still behaves the same, session token can be retrieved from id_token](https://videobin.shopify.io/v/bjQre0)
- 📹 [TokenExchange no longer bounces to patch session token on initial load since id_token is part of the URL params](https://videobin.shopify.io/v/xdRR15)

### Checklist

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -92,29 +92,6 @@ Edit `config/initializer/shopify_app.rb` and ensure the following configurations
 + config.shop_session_repository = 'Shop'
 ```
 
-#### Inspect server logs
-
-If you have checked the configurations above, and the app is still using cookies, then it is possible that the `shopify_app` gem defaulted to relying on cookies. This would happen when your browser allows third-party cookies and a session token was not successfully found as part of your request.
-
-In this case, check the server logs to see if the session token was invalid:
-
-```los
-[ShopifyApp::JWT] Failed to validate JWT: [JWT::<Error>] <Failure message>
-```
-
-*Example*
-
-```
-[ShopifyApp::JWT] Failed to validate JWT: [JWT::ImmatureSignature] Signature nbf has not been reached
-```
-
-**Note:** In a local development environment, you may want to temporarily update your `Gemfile` to point to a local instance of the `shopify_app` library instad of an installed gem. This will enable you to use a debugging tool like `byebug` to debug the library.
-
-```diff
-- gem 'shopify_app', '~> 14.2'
-+ gem 'shopify_app', path: '/path/to/shopify_app'
-```
-
 ### My app can't make requests to the Shopify API
 
 > **Note:** Session tokens cannot be used to make authenticated requests to the Shopify API. Learn more about authenticating your backend requests to Shopify APIs at [Shopify API authentication](https://shopify.dev/concepts/about-apis/authentication).
--- a/docs/Upgrading.md
+++ b/docs/Upgrading.md
@@ -51,6 +51,10 @@ If you have overwritten these methods in your callback controller to modify the 
 update your app to use configurable option `config.custom_post_authenticate_tasks` instead. See [post authenticate tasks](/docs/shopify_app/authentication.md#post-authenticate-tasks)
 for more information.
 
+#### (v23.0.0) - Deprecated "ShopifyApp::JWT" class
+The `ShopifyApp::JWT` class has been deprecated in `v23.0.0`. Use [ShopifyAPI::Auth::JwtPayload](https://github.com/Shopify/shopify-api-ruby/blob/main/lib/shopify_api/auth/jwt_payload.rb)
+class from the `shopify_api` gem instead.
+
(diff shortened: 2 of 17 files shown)


docs/Troubleshooting.md
```diff
@@ -92,29 +92,6 @@ Edit `config/initializer/shopify_app.rb` and ensure the following configurations
 + config.shop_session_repository = 'Shop'
 ```
 
-#### Inspect server logs
```
paulomarg, line 95: I don't know if we want to remove all of this, are we sure these logs are irrelevant now?

docs/Upgrading.md
```diff
@@ -51,6 +51,10 @@ If you have overwritten these methods in your callback controller to modify the
 update your app to use configurable option `config.custom_post_authenticate_tasks` instead. See [post authenticate tasks](/docs/shopify_app/authentication.md#post-authenticate-tasks)
 for more information.
 
+#### (v23.0.0) - Deprecated "ShopifyApp::JWT" class
```
paulomarg, line 54: I'd also highlight that the API of the new one is a superset of the old one, so apps can safely search replace them if they ever reference the class directly, or not have to update at all if we return this token at any point.

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -273,10 +274,10 @@ def user_session_expected?
       online_token_configured?
     end
 
-    def load_current_session(auth_header: nil, cookies: nil, is_online: false)
+    def load_current_session(shopify_id_token: nil, cookies: nil, is_online: false)
       return ShopifyAPI::Context.load_private_session if ShopifyAPI::Context.private?
 
-      session_id = ShopifyAPI::Utils::SessionUtils.current_session_id(auth_header, cookies, is_online)
+      session_id = ShopifyAPI::Utils::SessionUtils.current_session_id(shopify_id_token, cookies, is_online)
```
paulomarg, line 280: Just to triple check: this will filter out the `Bearer ` part if present, right?

lib/shopify_app/controller_concerns/with_shopify_id_token.rb
paulomarg, line 1: Not to be fixed in this PR, but I kinda feel like it'd be cool to have 2 tiers of concerns - the ones we export and want folks to use, and the ones we use internally.

lib/shopify_app/middleware/jwt_middleware.rb
```diff
@@ -24,21 +25,24 @@ def call_next(env)
...
     def set_env_variables(token, env)
-      jwt = ShopifyApp::JWT.new(token)
+      jwt = ShopifyAPI::Auth::JwtPayload.new(token)
 
+      env["jwt.token"] = token
       env["jwt.shopify_domain"] = jwt.shopify_domain
       env["jwt.shopify_user_id"] = jwt.shopify_user_id
       env["jwt.expire_at"] = jwt.expire_at
+    rescue ShopifyAPI::Errors::InvalidJwtTokenError
+      # ShopifyApp::JWT did not raise any exceptions, ensuring behaviour does not change
```
paulomarg, line 44: Should we just log debug this so it's not a completely silent `rescue`?

paulomarg (approved):

lib/shopify_app/controller_concerns/with_shopify_id_token.rb
zzooeeyy: Yea it's slightly confusing for sure, since there's no good "prevention" to stop users from including these private concerns...

lib/shopify_app/middleware/jwt_middleware.rb
zzooeeyy: The logging was explicitly removed from this issue https://github.com/Shopify/shopify_app/pull/1610, so I'd like to keep the behaviour the same..

paulomarg: Fair point!

PR #1833, opened 2024-04-25
EnsureBilling and EnsureAuthenticatedLinks works with token exchange

zzooeeyy:
- Fixes: https://github.com/Shopify/develop-app-access/issues/295

### What this PR does
* Making sure EnsureBilling and EnsureAuthenticatedLinks can call methods defined in token exchange, by extracting some common helper methods for re-use
* Fixed an infinite redirect loop caused by handling billing API error -
  * Steps to reprduce:
    1. Open app
    2. EnsureHasSession checks if billing is valid
    3. Billing API responds with billing error 
    4. EnsureBilling handles billing error by redirecting to shop /login with shop and host params
    5. /login starts auth again
    6. Auth CallbackController handles the callback, but it also checks billing, which could still return error - thus redirecting back to /login again.  ↺
   * Fix was to only redirect to shop /login withOUT shop and host params so that /login does not re-start auth.
* TokenExchangeConcern -
  *  If id_session is invalid, redirect to embed the app first before redirecting to bounce page.

### Before - Billing API error sent app into infinite redirect loop
https://github.com/Shopify/shopify_app/assets/102243935/24cb3077-3bb2-4c34-a620-2b69d67c003f

### After - Billing API error does not send app into infinite redirect loop
https://github.com/Shopify/shopify_app/assets/102243935/527b509e-f844-41c9-94a8-5b5379a94a57

### Token exchange works with EnsureBIlling and can load app after confirming charge
https://github.com/Shopify/shopify_app/assets/102243935/e71b810d-653a-40f3-9f49-2dc34860618d

### Checklist

### DIFF
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -79,7 +79,7 @@ module ShopifyApp
       response.set_header(ACCESS_TOKEN_REQUIRED_HEADER, "true")
     end
 
-    def add_top_level_redirection_headers(ignore_response_code: false)
+    def add_top_level_redirection_headers(url: nil, ignore_response_code: false)
       if request.xhr? && (ignore_response_code || response.code.to_i == 401)
         ShopifyApp::Logger.debug("Adding top level redirection headers")
         # Make sure the shop is set in the redirection URL
@@ -94,7 +94,7 @@ module ShopifyApp
           end
         end
 
-        url = login_url_with_optional_shop
+        url ||= login_url_with_optional_shop
 
         ShopifyApp::Logger.debug("Setting Reauthorize-Url to #{url}")
         RedirectForEmbedded.add_app_bridge_redirect_url_header(url, response)
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -79,14 +79,7 @@ module ShopifyApp
       response.set_header(ACCESS_TOKEN_REQUIRED_HEADER, "true")
     end
 
-    def jwt_expire_at
-      expire_at = request.env["jwt.expire_at"]
-      return unless expire_at
-
-      expire_at - 5.seconds # 5s gap to start fetching new token in advance
-    end
-
-    def add_top_level_redirection_headers(url: nil, ignore_response_code: false)
+    def add_top_level_redirection_headers(ignore_response_code: false)
       if request.xhr? && (ignore_response_code || response.code.to_i == 401)
         ShopifyApp::Logger.debug("Adding top level redirection headers")
         # Make sure the shop is set in the redirection URL
@@ -101,23 +94,14 @@ module ShopifyApp
           end
         end
 
-        url ||= login_url_with_optional_shop
+        url = login_url_with_optional_shop
 
         ShopifyApp::Logger.debug("Setting Reauthorize-Url to #{url}")
-        response.set_header("X-Shopify-API-Request-Failure-Reauthorize", "1")
-        response.set_header("X-Shopify-API-Request-Failure-Reauthorize-Url", url)
+        RedirectForEmbedded.add_app_bridge_redirect_url_header(url, response)
       end
     end
 
     protected
-
-    def jwt_shopify_domain
-      request.env["jwt.shopify_domain"]
-    end
-
-    def jwt_shopify_user_id
-      request.env["jwt.shopify_user_id"]
-    end
 
     def host
       params[:host]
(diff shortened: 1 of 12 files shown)


lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -79,14 +79,7 @@ def signal_access_token_required
...
 
-    def jwt_expire_at
-      expire_at = request.env["jwt.expire_at"]
-      return unless expire_at
-
-      expire_at - 5.seconds # 5s gap to start fetching new token in advance
-    end
-
-    def add_top_level_redirection_headers(url: nil, ignore_response_code: false)
+    def add_top_level_redirection_headers(ignore_response_code: false)
```
zzooeeyy, line 82: `url` argument is not used anywhere anymore, removing this. 
Extracted helper `RedirectForEmbedded.add_app_bridge_redirect_url_header(url, response)` to be reused elsewhere

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -79,14 +79,7 @@ def signal_access_token_required
       response.set_header(ACCESS_TOKEN_REQUIRED_HEADER, "true")
     end
 
-    def jwt_expire_at
-      expire_at = request.env["jwt.expire_at"]
-      return unless expire_at
-
-      expire_at - 5.seconds # 5s gap to start fetching new token in advance
-    end
```
zzooeeyy, line 87: Extracted these 'jwt' helpers to be inside `WithShopifyIdToken` concern for reusability

lib/shopify_app/controller_concerns/login_protection.rb
```diff
@@ -101,24 +94,15 @@ def add_top_level_redirection_headers(url: nil, ignore_response_code: false)
...
 
     protected
 
-    def jwt_shopify_domain
-      request.env["jwt.shopify_domain"]
-    end
-
-    def jwt_shopify_user_id
-      request.env["jwt.shopify_user_id"]
-    end
```
zzooeeyy, line 120: Extracted these 'jwt' helpers to be inside `WithShopifyIdToken` concern for reusability

lib/shopify_app/controller_concerns/token_exchange.rb
```diff
@@ -83,8 +89,36 @@ def redirect_to_bounce_page
       )
     end
 
+    def missing_embedded_param?
+      !params[:embedded].present? || params[:embedded] != "1"
+    end
+
+    def redirect_to_embed_app
```
paulomarg, line 96: Just out of curiosity - won't we have other places in the code that need to do this as well? Seems to me like this might also be extractable if that's the case.

lib/shopify_app/controller_concerns/token_exchange.rb
```diff
@@ -83,8 +89,36 @@ def redirect_to_bounce_page
...
+      ShopifyApp::Logger.debug("Redirecting to embed app")
+
+      host = if params[:host]
+        params[:host]
+      elsif params[:shop]
+        Base64.encode64("#{sanitized_shop_name}/admin")
```
paulomarg, line 102: This is a little bit of an assumption - I think there are some cases where we might not want this (requests aren't coming from the admin), but we can deal with those if / when they happen.

test/shopify_app/controller_concerns/with_shopify_id_token_test.rb
```diff
@@ -120,6 +120,36 @@ def setup
...
+    with_application_test_routes do
+      request.env["jwt.shopify_user_id"] = expected_user_id
+      get :index
+
+      assert_equal expected_user_id, @controller.jwt_shopify_user_id
+    end
+  end
+
+  test "#jwt_expire_at returns jwt.expire_at - 5 seconds from request env" do
+    expected_expire_at = Time.now.to_i
```
paulomarg, line 144: Are we locking down the time here? I wonder if this could lead to flaky tests if the second happens to shift during the run.

lib/shopify_app/controller_concerns/login_protection.rb
zzooeeyy: hmmm yea, although this method added more checks around _when_ to add the headers. I think I'll just add the existing behaviour back to avoid breaking behaviour for apps

lib/shopify_app/controller_concerns/token_exchange.rb
paulomarg: There are cases where `host` isn't necessarily `shop/admin`, so we're basically assuming that will be the case whenever we hit this path.

paulomarg (approved):

elioncho:
Hi! I am getting a redirect loop when I approve the billing charge. My home controller is a subclass of the AuthenticatedController. My AuthenticatedController has EnsureAuthenticatedLinks and EnsureHasSession. Once the redirection loop stops I can go ahead and open the app on the store. What could be happening here?

CleanShot 2024-06-12 at 16 52 49@2x
CleanShot 2024-06-12 at 16 52 49@2x

kirillplatonov:
@zzooeeyy why do we need `EnsureAuthenticatedLinks` with token exchange? It was used to redirect to splash screen which no longer seems to be necessary. Or am I missing something?

PR #1834, opened 2024-04-29
Add new auth strategy docs & set token exchange as default configuration from rails generator

zzooeeyy:
### What this PR does
- https://github.com/Shopify/temp-project-mover-Archetypically-20250319145332/issues/260
- https://github.com/Shopify/temp-project-mover-Archetypically-20250319145332/issues/261

- Added documentations for token exchange embedded auth strategy (mostly copied from the Remix app).
  - README section
  - Changelog 
  - Session & Authentication guides
  - Upgrading guide

- Modified the rails configuration generator to enable this new auth strategy config by default if it's an embedded app.

## Before and after with the configuration on/off 
- The embedded app is using previous OAuth with callbacks in "authorization code grant" flow when `config.new_embedded_auth_strategy = false`.
- The embedded app is using the new zero redirect OAuth with "token exchange" flow when `config.new_embedded_auth_strategy = true`

### New app install 
| Auth code flow (old)  | New embedded app auth (new) |
| ------------- | ------------- |
| <video src="https://github.com/Shopify/shopify_app/assets/102243935/3a626fab-7ac5-4687-9885-734dc8172dcb">  | <video src="https://github.com/Shopify/shopify_app/assets/102243935/37bfa6ba-9b07-4ddc-9b9b-6832fbd25576">|
| There are multiple redirects back and forth from app to Shopify to retrieve both the offline and online access tokens.  Took about 15s from "install" to home page being displayed. | There are no redirects happening at all from the app after the install. Home page displayed in about 5s  |

### Adding new access scope
| Auth code flow (old)  | New embedded app auth (new) |
| ------------- | ------------- |
| <video src="https://github.com/Shopify/shopify_app/assets/102243935/8e61d6e7-1e84-48c5-91a0-abff220a6782">  | <video src="https://github.com/Shopify/shopify_app/assets/102243935/93226f75-008f-42c8-8ca6-873c1ce192e0">|
| When a new scope has been deployed, the old authorization code flow can't pick it up until the server has been restarted. Then the user has to go through many redirects to get the new access tokens with the new access scopes. | When a new scope has been deployed, the new scope grant will be automatically prompted. Once the user grants approval to the new scopes, the operation will continue. 

### Expired online access token
| Auth code flow (old)  | New embedded app auth (new) |
| ------------- | ------------- |
| <video src="https://github.com/Shopify/shopify_app/assets/102243935/413f6713-8dd6-46cf-8e47-bd1c50064823">  | <video src="https://github.com/Shopify/shopify_app/assets/102243935/3375d3cc-9fa6-434e-be7f-cda679da9d02">|
| When the online access token has expired. The app redirected multiple times to retrieve the new access tokens, and the XHR operation from the front end didn't get completed. | When the online access token has expired, the new auth strategy automatically fetches the new access token before handling the actual API request with no redirects. |

### Invalidated online access token
| Auth code flow (old)  | New embedded app auth (new) |
| ------------- | ------------- |
| <video src="https://github.com/Shopify/shopify_app/assets/102243935/a5ba36d0-8b4a-4b49-aa3f-40d291362963">  | <video src="https://github.com/Shopify/shopify_app/assets/102243935/10dbfe16-ddb4-4300-b4ae-b9abb712402b">|
| When the online access token has been invalidated, the app could not handle XHR requests until the app has been reloaded and redirected multiple times for OAuth to retrieve new access tokens | When the online access token has been invalidated, the server handles the 401 HTTP unauthorized error and fetches new access token through token exchange before retrying the controller action again. The result is a seamless server side handling of this error with zero redirects. |

### Checklist

### DIFF
--- a/docs/shopify_app/authentication.md
+++ b/docs/shopify_app/authentication.md
@@ -37,7 +37,7 @@ with [token exchange](#token-exchange) instead of the authorization code grant f
 
 OAuth process by exchanging the current user's [session token (shopify id token)](https://shopify.dev/docs/apps/auth/session-tokens) for an
 [access token](https://shopify.dev/docs/apps/auth/access-token-types/online.md) to make
-authenticated Shopify API queries. This can replace authorization code grant flow completely if you also take advantage of [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation).
+authenticated Shopify API queries. This will replace authorization code grant flow completely when your app is configured with [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation).
 
 To enable token exchange authorization strategy, you can follow the steps in ["New embedded app authorization strategy"](/README.md#new-embedded-app-authorization-strategy).
 Upon completion of the token exchange to get the access token, [post authenticated tasks](#post-authenticate-tasks) will be run.
@@ -60,12 +60,7 @@ Learn more about [Shopify managed installation](https://shopify.dev/docs/apps/au
 Using token exchange will ensure that the access token retrieved will always have the latest access scopes granted by the user.
 
 ## Authorization Code Grant Flow
-> [!NOTE]
-> If you are building an embedded app, we **strongly** recommend using [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation)
-with [token exchange](#token-exchange) instead of the authorization code grant flow.
-
-
-Authorization code grant flow is the previous OAuth flow that requires the app to redirect the user 
+Authorization code grant flow is the OAuth flow that requires the app to redirect the user 
 to Shopify for installation/authorization of the app to access the shop's data. It is still required for apps that are not embedded.
 
 To perform [authorization code grant flow](https://shopify.dev/docs/apps/auth/get-access-tokens/authorization-code-grant), you app will need to handle
@@ -79,8 +74,6 @@ The ShopifyApp gem then redirect the merchant to Shopify, to ask for permission 
 for detailed implementation)
 
 ### OAuth callback
-
->️ **Note:** In Shopify App version 8.4.0, we have extracted the callback logic in its own controller. If you are upgrading from a version older than 8.4.0 the callback action and related helper methods were defined in `ShopifyApp::SessionsController` ==> you will have to extend `ShopifyApp::CallbackController` instead and port your logic to the new controller.
 
 Shopify will redirect the merchant back to your app's callback URL once they approve the app installation.
 Upon completing the OAuth flow, Shopify calls the app at `ShopifyApp.configuration.login_callback_url`. (This was provided to Shopify in the OAuth begin URL parameters)
--- a/docs/shopify_app/sessions.md
+++ b/docs/shopify_app/sessions.md
@@ -223,12 +223,14 @@ end
 ```
 
 If the error is being rescued in the action, it's still possible to make use of `with_token_refetch` provided by `EnsureHasSession` so that a new access token is fetched and the code is executed again with it. This will also update the session parameter with the new attributes.
+This block should be used to wrap the code that makes API queries, so your business logic won't be retried.
 
 ```ruby
 class MyController < ApplicationController
   include ShopifyApp::EnsureHasSession
 
   def index
+    # Your app's business logic
     with_token_refetch(current_shopify_session, shopify_id_token) do
       # Unauthorized errors raised within this block will initiate token exchange.
       # `with_token_refetch` will store the new access token and use it
@@ -237,6 +239,7 @@ class MyController < ApplicationController
       client = ShopifyAPI::Clients::Graphql::Admin.new(session: current_shopify_session)
       client.query(options)
     end
+    # Your app's business logic continues
   rescue => error
     # app's specific error handling
   end
@@ -331,6 +334,8 @@ When the configuration flag `check_session_expiry_date` is set to true, the user
 ## Migrating from shop-based to user-based token strategy
 
 1. Run the `user_model` generator as [mentioned above](#user-online-token-storage).
+  - The generator will ask whether you want to migrate the User model to include `access_scopes` and `expires_at` columns. `expires_at` field is useful for detecting when the user session has expired and trigger a re-auth before an operation. It can reduce
+   API failures for invalid access tokens. Configure the [expiry date check](#expiry-date) to complete this feature.
 2. Ensure that both your `Shop` model and `User` model includes the [necessary concerns](#available-activesupportconcerns-that-contains-implementation-of-the-above-methods)
 3. Update the configuration file to use the new session storage.
 
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ These routes are configurable. See the more detailed [*Engine*](/docs/shopify_ap
 
 To learn more about how this gem authenticates with Shopify, see [*Authentication*](/docs/shopify_app/authentication.md).
 
-### New embedded app authorization strategy
+### New embedded app authorization strategy (Token Exchange)
 
 > [!TIP]
 > If you are building an embedded app, we **strongly** recommend using [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation)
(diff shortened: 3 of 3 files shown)


docs/shopify_app/authentication.md
```diff
@@ -10,17 +10,80 @@ See [*Getting started with session token authentication*](https://shopify.dev/do
...
+    - Access scope changes are handled by Shopify when you use [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation)
+2. [Authorization Code Grant Flow](#authorization-code-grant-flow)
+    - Suitable for non-embedded apps
+    - Installations, and access scope changes are managed by the app
+
+## Token Exchange
+
+OAuth process by exchanging the current user's [session token (shopify id token)](https://shopify.dev/docs/apps/auth/session-tokens) for an
+[access token](https://shopify.dev/docs/apps/auth/access-token-types/online.md) to make
+authenticated Shopify API queries. This can replace authorization code grant flow completely if you also take advantage of [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation).
```
paulomarg, line 40: This makes it sound like managed install is optional, but that's not the case, right?

docs/shopify_app/authentication.md
```diff
@@ -10,17 +10,80 @@ See [*Getting started with session token authentication*](https://shopify.dev/do
...
+
+To enable token exchange authorization strategy, you can follow the steps in ["New embedded app authorization strategy"](/README.md#new-embedded-app-authorization-strategy).
+Upon completion of the token exchange to get the access token, [post authenticated tasks](#post-authenticate-tasks) will be run.
```
paulomarg, line 43: Is there any difference in behaviour as to when post auth tasks are run, when compared to auth code? If so, this might be a good place to call that out.

docs/shopify_app/authentication.md
```diff
@@ -10,17 +10,80 @@ See [*Getting started with session token authentication*](https://shopify.dev/do
...
+Learn more about [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation).
+Using token exchange will ensure that the access token retrieved will always have the latest access scopes granted by the user.
+
+## Authorization Code Grant Flow
+> [!NOTE]
+> If you are building an embedded app, we **strongly** recommend using [Shopify managed installation](https://shopify.dev/docs/apps/auth/installation#shopify-managed-installation)
+with [token exchange](#token-exchange) instead of the authorization code grant flow.
+
+
+Authorization code grant flow is the previous OAuth flow that requires the app to redirect the user
```
paulomarg, line 68: I would try to avoid temporal references here as they will eventually become outdated.

docs/shopify_app/authentication.md
```diff
@@ -10,17 +10,80 @@ See [*Getting started with session token authentication*](https://shopify.dev/do
...
+### Begin OAuth
+ShopifyApp automatically redirects the user to Shopify to complete OAuth to install the app when the `ShopifyApp.configuration.login_url` is reached.
+Behind the scenes the ShopifyApp gem starts the process by calling `ShopifyAPI::Auth::Oauth.begin_auth` to build the 
+redirect URL with necessary parameters like the OAuth callback URL, scopes requested, type of access token (offline or online) requested, etc.
+The ShopifyApp gem then redirect the merchant to Shopify, to ask for permission to install the app. (See [ShopifyApp::SessionsController.redirect_to_begin_oauth](https://github.com/Shopify/shopify_app/blob/main/app/controllers/shopify_app/sessions_controller.rb#L76-L96)
+for detailed implementation)
+
+### OAuth callback
 
 >️ **Note:** In Shopify App version 8.4.0, we have extracted the callback logic in its own controller. If you are upgrading from a version older than 8.4.0 the callback action and related helper methods were defined in `ShopifyApp::SessionsController` ==> you will have to extend `ShopifyApp::CallbackController` instead and port your logic to the new controller.
```
paulomarg, line 83: This is a fairly old warning, I wonder if we still need to have the version reference here, or if we want to make it more "permanent".

docs/shopify_app/sessions.md
```diff
@@ -20,25 +19,24 @@ Sessions are used to make contextual API calls for either a shop (offline sessio
...
+- **Shop** ([offline access](https://shopify.dev/docs/apps/auth/access-token-types/offline))
   - Access token is linked to the store
   - Meant for long-term access to a store, where no user interaction is involved
   - Ideal for background jobs or maintenance work
-- **User** ([online access](https://shopify.dev/docs/apps/auth/oauth/access-modes#online-access))
+- **User** ([online access](https://shopify.dev/docs/apps/auth/access-token-types/online))
   - Access token is linked to an individual user on a store
   - Meant to be used when a user is interacting with your app through the web
 
-⚠️  [Read more about Online vs. Offline access here](https://shopify.dev/apps/auth/oauth/access-modes).
```
paulomarg, line 39: I kind of liked this sentence, any chance we can keep it around if folks want to understand a bit more?

docs/shopify_app/sessions.md
```diff
@@ -203,6 +203,97 @@ user.with_shopify_session do
...
+```
+
+If the error is being rescued in the action, it's still possible to make use of `with_token_refetch` provided by `EnsureHasSession` so that a new access token is fetched and the code is executed again with it. This will also update the session parameter with the new attributes.
+
+```ruby
+class MyController < ApplicationController
+  include ShopifyApp::EnsureHasSession
+
+  def index
+    with_token_refetch(current_shopify_session, shopify_id_token) do
```
paulomarg, line 232: If we're using methods provided by `EnsureHasSession`, do we still need to explicitly pass the session and id token as params? Maybe we could do that internally and keep the params as an override (or not?), that way calling this is a bit simpler.

docs/shopify_app/sessions.md
```diff
@@ -203,6 +203,97 @@ user.with_shopify_session do
...
+
+    # If this call raises an Unauthorized error from Shopify, EnsureHasSession
+    # will execute the action again after performing token exchange.
+    # It will store and use the newly retrieved access token for this and any subsequent calls.
+    client.query(options)
+  end
+end
+```
+
+If the error is being rescued in the action, it's still possible to make use of `with_token_refetch` provided by `EnsureHasSession` so that a new access token is fetched and the code is executed again with it. This will also update the session parameter with the new attributes.
```
paulomarg, line 225: I can see this being a bit confusing for people in the sense that the code might be run twice. Should we explicitly state that only the query should be executed inside the block, and no business logic? Should we have a way to do this in the client itself, something like

```rb
def index
  client = Client.new
  client.query({ **options, refetch_id_token: shopify_id_token })
end
```

I'm mostly just bikeshedding here, wondering if we can go a little bit further to simplify the app's code.

paulomarg (approved):
I had a few thoughts, but nothing blocking.

lib/generators/shopify_app/install/templates/shopify_app.rb.tt
```diff
@@ -4,6 +4,8 @@ ShopifyApp.configure do |config|
   config.scope = "<%= @scope %>" # Consult this page for more scope options:
                                   # https://help.shopify.com/en/api/getting-started/authentication/oauth/scopes
   config.embedded_app = <%= embedded_app? %>
+  config.new_embedded_auth_strategy = <%= embedded_app? %>
```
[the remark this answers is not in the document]
zzooeeyy: Yea I thought about enabling it, but I thought it might be confusing  for apps that _aren't_ using online tokens.... 
And Paulo brings up a good point, I'm not sure why we have this check anyways, it should just not get a new token automatically if there are no online tokens or if online token doesn't have an expiry date column, and wouldn't need a config for it?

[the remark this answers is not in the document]
zzooeeyy: And I just found this change to the template -- https://github.com/Shopify/shopify-app-template-ruby/pull/24

Looks offline tokens are preferred instead of using online token?

docs/shopify_app/sessions.md
paulomarg: Ah that makes sense, then this is perfect.

docs/shopify_app/authentication.md
paulomarg: Right, but I meant more in the sense of whether there are any scenarios where the auth code flow would trigger the tasks and this won't. I guess there aren't any?

PR #1844, opened 2024-05-06
Don't delete shop session on 401 error

zzooeeyy:
### What this PR does
Not deleting the shop session when 401 error has occurred in token exchange retry.

### Checklist


paulomarg (approved):
Is this behaviour present in remix too? Might be something we want to change there too, as folks make a good point about being overzealous in cleaning up data.

PR #1860, opened 2024-06-11
LoginProtection: Extract private method set_session_return_to

jaredbeck:
### What this PR does

Apps which use cookies for session storage, and thus have a 4kB session data limit, may choose to override this method to prevent excessively-long `query` strings from causing a CookieOverflow error.

Currently, apps must override the entire `redirect_to_login` method. To make gem updates easier, we want to override a smaller area.

This PR does not change functionality.

### Things to focus on

Confirm that the method extraction was performed correctly, ie. without changing functionality.

### Checklist

Please let me know if any of the following are necessary in this case. It's not immediately clear to me if they are necessary.

### DIFF
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -122,10 +122,17 @@ module ShopifyApp
           query = Rack::Utils.parse_nested_query(referer.query)
           query = query.merge(sanitized_params).to_query
         end
-        session[:return_to] = query.blank? ? path.to_s : "#{path}?#{query}"
+        session[:return_to] = return_to_url(path, query)
         ShopifyApp::Logger.debug("Redirecting to #{login_url_with_optional_shop}")
         redirect_to(login_url_with_optional_shop)
       end
+    end
+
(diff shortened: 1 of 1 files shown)


lizkenyon (commented):
Hi there 👋 

Thanks for taking the time to open this PR! ⭐ 

The team was wondering if there was a way we could get this working by default so not everyone who runs into this issue has to overide it, while still allowing folks to override if our default logic doesn't work for them.

In particular we were wondering in the scenario the URL is to long what are you setting the return URL to?

We also need to be careful because because session[:return_to]  is used in other places. If we extract setting session[:return_to] to a separate method, we need to make sure to use session[:return_to] = "override value" to avoid other methods to break when accessing session[:return_to] .

Then we could possible do something like this

In the `redirect_to_login` 
```
session[:return_to] = return_to_url(path, query)
```

And
```
def return_to_url(path, query)
   if Context.embedded?
  query.blank? ? path.to_s : "#{path}?#{query}"
  else
       // shorten URL appropriately 
  end
end
```

How does this sound to you!

jaredbeck:
How does this look?

PR #1861, opened 2024-06-11
Remove JWT middleware

### LINKED ISSUE #1744: Refactor JWT to use `around_action` when needed instead of middleware used with every request
## Overview/summary

We currently use a [JWT middleware](https://github.com/Shopify/shopify_app/blob/main/lib/shopify_app/engine.rb#L31) to parse the JWT for every request that is made to a rails app that uses this gem.  For actions that don't need to make API calls to Shopify we are still spending time logging and trying to parse a token that doesn't exist. For applications that have a large amount of traffic, this inefficiency adds up. 

We should explore moving that logic to our controller concerns as part of a `around_action` or `before_action` filter so we only parse JWT when we know it'll be there.

danielpgross:
### What this PR does

Closes #1744

Removes the JWT middleware, since JWT authentication was moved to `shopify-api-ruby` in v19 and then deprecated here.

### Things to focus on

1. 
2. 
3. 

### Checklist


lib/shopify_app/controller_concerns/with_shopify_id_token.rb
```diff
@@ -5,31 +5,30 @@ module WithShopifyIdToken
     extend ActiveSupport::Concern
 
     def shopify_id_token
-      @shopify_id_token ||= id_token_from_request_env || id_token_from_authorization_header || id_token_from_url_param
+      @shopify_id_token ||= id_token_from_authorization_header || id_token_from_url_param
```
tgwizard, line 8: If there's no token in header or URL param, this will memoize nil and keep evaluating the `id_token_from_authorization_header || id_token_from_url_param` part on future calls.

Perhaps we should write this as

```suggestion
      return @shopify_id_token if defined?(@shopify_id_token)
      @shopify_id_token = id_token_from_authorization_header || id_token_from_url_param
```

tgwizard: (there are similar concerns with the memoization in the other methods here, not sure how important this is).

tgwizard (approved):
Very nice! This LGTM, though others should review too.

Admittedly, I'm having troubles following the code that actually ensures a shopify app verifies the ID token is present and valid. The code I'm reading seems to indicate we do some kind of token exchange - but I thought we fully relied on the JWT, without any external calls.

Could you please walk me through the flow?

danielpgross:
> Very nice! This LGTM, though others should review too.
> 
> Admittedly, I'm having troubles following the code that actually ensures a shopify app verifies the ID token is present and valid. The code I'm reading seems to indicate we do some kind of token exchange - but I thought we fully relied on the JWT, without any external calls.
> 
> Could you please walk me through the flow?

Thanks @tgwizard! Will walk you through it according to my understanding:

1. An app includes the `EnsureHasSession` concern in a given controller to restrict that controller to requests that are authenticated with a valid session -- think of this as the entry point to the flow.
2. `EnsureHasSession` includes `LoginProtection` and sets up `activate_shopify_session` from there as an `around_action`: https://github.com/Shopify/shopify_app/blob/ef54cdc1742d77e5410218e72bfb864fa2623861/app/controllers/concerns/shopify_app/ensure_has_session.rb#L16
3. `activate_shopify_session` calls `current_shopify_session`, which calls `load_current_session`and in turn `ShopifyAPI::Utils::SessionUtils.current_session_id`
https://github.com/Shopify/shopify_app/blob/ef54cdc1742d77e5410218e72bfb864fa2623861/lib/shopify_app/controller_concerns/login_protection.rb#L25 
https://github.com/Shopify/shopify_app/blob/ef54cdc1742d77e5410218e72bfb864fa2623861/lib/shopify_app/controller_concerns/login_protection.rb#L56
https://github.com/Shopify/shopify_app/blob/ef54cdc1742d77e5410218e72bfb864fa2623861/lib/shopify_app/controller_concerns/login_protection.rb#L264
4. Now we're in `ShopifyAPI` land, and `session_id_from_shopify_id_token` gets called:
https://github.com/Shopify/shopify-api-ruby/blob/a64f7dd391306033ca58399b42fa32207ba6dc34/lib/shopify_api/utils/session_utils.rb#L23
5. The JWT finally gets decoded and then a session ID is generated from it: 
https://github.com/Shopify/shopify-api-ruby/blob/a64f7dd391306033ca58399b42fa32207ba6dc34/lib/shopify_api/utils/session_utils.rb#L48

lib/shopify_app/controller_concerns/with_shopify_id_token.rb
```diff
@@ -5,31 +5,30 @@ module WithShopifyIdToken
...
+    end
+
+    def jwt_payload
+      @jwt_payload ||= shopify_id_token.present? ? ShopifyAPI::Auth::JwtPayload.new(shopify_id_token) : nil
```
tgwizard, line 12: Is there some way we can skip parsing the JWT token twice, once here, and once in https://github.com/Shopify/shopify-api-ruby/blob/a64f7dd391306033ca58399b42fa32207ba6dc34/lib/shopify_api/utils/session_utils.rb#L48?

(That could be a follow-up optimization).

lib/shopify_app/controller_concerns/with_shopify_id_token.rb
```diff
@@ -5,31 +5,30 @@ module WithShopifyIdToken
...
     end
 
     def jwt_shopify_domain
-      request.env["jwt.shopify_domain"]
+      @jwt_shopify_domain ||= jwt_payload.present? ? ShopifyApp::Utils.sanitize_shop_domain(jwt_payload.dest) : nil
     end
 
     def jwt_shopify_user_id
-      request.env["jwt.shopify_user_id"]
+      jwt_payload&.sub&.to_i
```
zzooeeyy, line 20: I think we can just use `jwt_payload&.shopify_user_id` here (and `jwt_payload&.shopify_domain`, `jwt_payload&.expire_at` ). Similar to how the [middleware used to behave](https://github.com/Shopify/shopify_app/pull/1861/files#diff-f5062bb2e1d5a0819d14760dd0ce26827465e4811baa9ce69736b22c6f0b31fdL41). These helpers had been added to [ShopifyAPI::Auth::JwtPayload](https://github.com/Shopify/shopify-api-ruby/commit/6c9c2f80628dcea0e3ff33a42b8c61023747981b) to avoid having to parse these everytime.

danielpgross: Nice, I didn't realize these were available. Updated.

lib/shopify_app/controller_concerns/with_shopify_id_token.rb
danielpgross: Good idea. Looks like that would be a change in the `shopify-api-ruby` repo, so will leave it as a follow-up.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 Unreleased
 ----------
+- ⚠️ [Breaking] Removes `ShopifyApp::JWTMiddleware`. Any existing app code relying on decoded JWT contents set from `request.env` should instead include the `WithShopifyIdToken` concern and call its respective methods. [#1861](https://github.com/Shopify/shopify_app/pull/1861)
+  - Impacted values:
+    - `request.env["jwt.token"]`
+    - `request.env["jwt.shopify_domain"]`
+    - `request.env["jwt.shopify_user_id"]`
+    - `request.env["jwt.expire_at"]`
```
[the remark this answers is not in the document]
danielpgross: Done 👍

PR #1873, opened 2024-07-04
Improve embedded requests detection with Sec-Fetch-Dest header

kirillplatonov:
### What this PR does

Token exchange concern uses `embedded` param to determine whether request should be redirected to embedded or bounce page should be displayed: https://github.com/Shopify/shopify_app/blob/main/lib/shopify_app/controller_concerns/token_exchange.rb#L67

It works well when the app just loaded and Shopify provided all necessary params with id_token. But subsequent requests in app won't have `embedded` param in the URL. Meaning, that in case of invalid token error, the app will try to redirect user to embedded within iframe and it will lead to browser error:
image

To solve this problem, I found an alternative way to detect embedded using `Sec-Fetch-Dest` browser header. It's a nice fallback to `embedded` param and should help us correctly detect embedded requests and redirect them properly.

References:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-Fetch-Dest
- https://caniuse.com/?search=Sec-Fetch-Dest

### Checklist


kirillplatonov:
cc @lizkenyon

PR #1878, opened 2024-07-10
Fix bug where locale is not read from session

nickpith:
### What this PR does

The changes in https://github.com/Shopify/shopify_app/pull/1711 introduced a bug where single-page apps that don't include the locale param in subsequent requests were unable to properly set the locale that was persisted in the user session. Instead they would use the default locale.

See https://github.com/Shopify/shopify_app/pull/1711/files#r1664570299 for background.

### Things to focus on

Nothing in particular since this is a very small change. A new test case was added to make sure it was accounted for.

### Checklist


awd:
Thank you @nickpith - LGTM :shipit:

PR #1884, opened 2024-07-24
Re-add deprecated files to prevent major release

paulomarg:
### What this PR does

In #1861, we (correctly) removed some files related to JWT that should no longer be used in `shopify_app`. However, by removing those files we'd be forced to make a major release, which is still a bit too close to the last one.

This PR adds back the code that was removed (but will be unused internally) so it doesn't break existing apps. It will log deprecations for them instead.

Log message will look like this
```
================================================
=> Upcoming deprecations in v23.0:
* 'CallbackController::perform_after_authenticate_job' and related methods 'install_webhooks', 'perform_after_authenticate_job'
* will be removed from CallbackController in the next major release. If you need to customize
* post authentication tasks, see https://github.com/Shopify/shopify_app/blob/main/docs/shopify_app/authentication.md#post-authenticate-tasks

* ShopifyApp::JWTMiddleware will be removed, use ShopifyApp::WithShopifyIdToken instead.
================================================
```

Use this branch in the Gemfile of an app using the ruby template, and set it up to use `ShopifyApp::JWTMiddleware`. 

### Things to focus on

1. Is it ok to not have tests? The feature will be unchanged until the next major release.
2. Can you see any fringe cases where it won't work?

### DIFF
--- a/lib/shopify_app/configuration.rb
+++ b/lib/shopify_app/configuration.rb
@@ -169,9 +169,9 @@ module ShopifyApp
       # TODO: Remove this before releasing v23.0.0
       message = <<~EOS
         ================================================
-        => Upcoming deprecations in v23.0:
+        => Upcoming changes in v23.0:
         * 'CallbackController::perform_after_authenticate_job' and related methods 'install_webhooks', 'perform_after_authenticate_job'
-        * will be removed from CallbackController in the next major release. If you need to customize
+        * are deprecated and  will be removed from CallbackController in the next major release. If you need to customize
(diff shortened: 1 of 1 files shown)


lib/shopify_app/configuration.rb
```diff
@@ -163,16 +163,18 @@ def post_authenticate_tasks
...
 
-    def log_callback_controller_method_deprecation
+    def log_v23_deprecations
       return unless Rails.env.development?
 
       # TODO: Remove this before releasing v23.0.0
       message = <<~EOS
         ================================================
-        => Upcoming deprecation in v23.0:
+        => Upcoming deprecations in v23.0:
```
lizkenyon, line 172:
```suggestion
        => Upcoming changes in v23.0:
```

? These are currently deprecated.

PR #1890, opened 2024-07-29
Add the `unified_admin_domain` configuration option for the unified admin domain.

MatWrz:
### What this PR does

Currently, the install path assumes that the installation is done on `admin.shopify.com` if not in a spin environment. This prevent installation of the app against a local shopify server or mock server.

The PR provides a `unified_admin_domain` that is used when building the installation URL will allow for installs on non-spin and non-production environments.

### Checklist

### DIFF
--- a/test/controllers/concerns/embedded_app_test.rb
+++ b/test/controllers/concerns/embedded_app_test.rb
@@ -3,7 +3,7 @@
 require "test_helper"
 require "action_view/testing/resolvers"
 
-class EmbeddedAppTest < ActionController::TestCase
+class EmbeddedAppTest < ActionDispatch::IntegrationTest
   class BaseTestController < ActionController::Base
     abstract!
 
@@ -28,9 +28,11 @@ class EmbeddedAppTest < ActionController::TestCase
     def redirect_to_embed
       redirect_to_embed_app_in_admin
     end
-  end
 
-  tests EmbeddedAppTestController
+    def current_shopify_domain
+      nil
+    end
+  end
 
   setup do
     Rails.application.routes.draw do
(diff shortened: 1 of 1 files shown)


lib/shopify_app/controller_concerns/frame_ancestors.rb
```diff
@@ -8,7 +8,7 @@ module FrameAncestors
       content_security_policy do |policy|
         policy.frame_ancestors(-> do
           domain_host = current_shopify_domain || "*.#{::ShopifyApp.configuration.myshopify_domain}"
-          "#{ShopifyAPI::Context.host_scheme}://#{domain_host} https://admin.shopify.com"
+          "#{ShopifyAPI::Context.host_scheme}://#{domain_host} https://admin.#{::ShopifyApp.configuration.unified_admin_domain}"
```
paulomarg, line 11: Should we use the utils method here instead for consistency?
```suggestion
          "#{ShopifyAPI::Context.host_scheme}://#{domain_host} #{ShopifyApp::Utils::unified_admin_domain}"
```

paulomarg, line 11: Also, should we add a test to `test/controllers/concerns/embedded_app_test.rb` to ensure that the ancestors are being set correctly?

paulomarg (approved):
Couple of nits, nothing blocking.

lib/shopify_app/controller_concerns/frame_ancestors.rb
paulomarg: Totally missed that it wasn't public 😅 - I am slightly concerned now that having both `unified_admin_path` and `unified_admin_domain` is going to lead to confusion. Maybe keep it this way for now and we think up a way of combining the two methods.

test/controllers/concerns/embedded_app_test.rb
```diff
@@ -3,7 +3,7 @@
 require "test_helper"
 require "action_view/testing/resolvers"
 
-class EmbeddedAppTest < ActionController::TestCase
+class EmbeddedAppTest < ActionDispatch::IntegrationTest
```
MatWrz, line 6: This was needed for Rails to populate the `Content-Security-Policy` header
It is also encouraged by the Rails doc when testing controllers. (https://api.rubyonrails.org/classes/ActionController/TestCase.html)

>Rails discourages the use of functional tests in favor of integration tests (use [ActionDispatch::IntegrationTest](https://api.rubyonrails.org/classes/ActionDispatch/IntegrationTest.html)).

>New Rails applications no longer generate functional style controller tests and they should only be used for backward compatibility. Integration style controller tests perform actual requests, whereas functional style controller tests merely simulate a request. Besides, integration tests are as fast as functional tests and provide lot of helpers such as as, parsed_body for effective testing of controller actions including even [API](https://api.rubyonrails.org/classes/ActionController/API.html) endpoints.

lib/shopify_app/controller_concerns/frame_ancestors.rb
MatWrz: I've added some tests for the content security policy header.

PR #1898, opened 2024-08-20
Set CurrentTenant to nil when gem is present

rachel-carvalho:
### What this PR does

This makes sure the `CurrentTenant` is set to `nil` if gem is present whenever `ShopifyApp` ends the request in a before action, which happens:

- When there's no current session when using auth code flow
- When there's no `id_token` or if it's invalid when using token exchange

#### Steps to reproduce the error:
1. create a new app with the CLI
2. make sure it has `write_products` scope and `shopify app deploy`
3. make sure `new_embedded_auth_strategy` is `false`
4. add the line to set the current tenant to `HomeController#index`, `ProductsController#count` and `ProductsController#create`
5. add the security gem
6. `bundle` and start the server
7. install the app
8. delete the row created in the `shops` table
9. click on `Populate 5 products`
10. 👀 see the current tenant exception in the server logs 💥 

#### Seeing the fix
1. point `shopify_app` in the Gemfile to this branch with `gem "shopify_app", github: "Shopify/shopify_app", ref: "fix_current_tenant_error"`
2. `bundle` and restart the server
3. open the app again
4. delete the row created by install in the `shops` table
5. click on `Populate 5 products`
6. 👀 see how it correctly redirects to get a new access token ✨ 

### Things to focus on

I couldn't find a nicer way to test the path of `defined?(ClassName)`, I'd appreciate if anyone has better suggestions.

Anything else I'm missing?

### Checklist

### DIFF
--- a/test/shopify_app/controller_concerns/login_protection_test.rb
+++ b/test/shopify_app/controller_concerns/login_protection_test.rb
@@ -618,6 +618,35 @@ class LoginProtectionControllerTest < ActionController::TestCase
     end
   end
 
+  test "#redirect_to_login sets current tenant to nil before redirecting when ShopifySecurityBase::CurrentTenant is defined" do
+    ::ShopifyAPI::Utils::SessionUtils.stubs(:current_session_id).returns(nil)
+    ShopifyApp::Logger.stubs(:debug)
+
+    with_shopify_security_base_current_tenant do |current_tenant_class, nil_tenant|
+      with_application_test_routes do
+        current_tenant_class.expects(:tenant=).with(nil_tenant)
+        ShopifyApp::Logger.expects(:debug)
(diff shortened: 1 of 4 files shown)


test/shopify_app/controller_concerns/login_protection_test.rb
```diff
@@ -656,4 +685,22 @@ def with_custom_login_url(url)
   ensure
     ShopifyApp.configure { |config| config.login_url = original_url }
   end
+
+  def with_shopify_security_base_current_tenant
+    shopify_security_base_module = Module.new
+    ShopifyApp::LoginProtection.const_set("ShopifySecurityBase", shopify_security_base_module)
```
rachel-carvalho, line 691: I couldn't find a nicer way to stub the existence of a class than using `const_set` then `remove_const` (which for some reason is private even though `const_set` isn't). Open to other ideas if anyone has them!

PR #1924, opened 2024-11-26
Drop log_v23_deprecations console warning

tamcgoey:
### What this PR does

We've seen this console warning for 8 months with each app boot in dev, thinking we could remove the warning. 

Especially since if there's breaking changes they are could be raised in the changelog + via exceptions.

### Checklist


Arkham:
Shouldn't we drop the warning when releasing v23? or am I missing something/

tamcgoey:
We could, just that this has been logging for 8 months now and v23 hasn't been released any time since then.

It's just annoying we see this error every time we boot up the app.

Arkham:
I see! Let me touch base with the team and see why we haven't released v23 yet :)

PR #1806, opened 2024-03-06
Remove script tag manager

nelsonwittwer:
### What this PR does

Removes `ScripttagManager`

#### Deprecation Messaging Error
We realize there was a communication error in messaging where deprecation logs messaged this resources would be `deprecated` in version 22 not `deleted`.  Given this log as been sounding off for two years now we felt that we should remove it rather than punting for the next major release.

We apologize for the confusion this caused and you find yourself affected. Refer to the [upgrading doc](https://github.com/Shopify/shopify_app/blob/main/docs/Upgrading.md) for instructions on how reimplement this functionality if you find yourself in the position of needing to upgrade to the latest version of this gem while still using the script tag manager.

### Reviewer's guide to testing

Installed on a local app with a dev store to ensure no errors were raised.

### Checklist

Before submitting the PR, please consider if any of the following are needed:

- [x] Update `CHANGELOG.md` if the changes would impact users
- [x] Update `README.md`, if appropriate.
- [x] Update any relevant pages in `/docs`, if necessary
- [x] For security fixes, the [Disclosure Policy](https://github.com/Shopify/shopify_app/blob/master/SECURITY.md#disclosure-policy) must be followed.

PR #1948, opened 2025-03-10
[minor] Scrip tags manager

lizkenyon:
### What this PR does
* This PR adds the a script tag manager to the `shopify_app` gem.
* This functionality was [removed in version 22 of this package](https://github.com/Shopify/shopify_app/pull/1806/files).
* Script tags are still a valid way for modify the content of vintage themes.

* This PR has several differences between the previous script tag manager
    * Not the manager adds script tags with GraphQL instead of REST
    * This adds the ability to specify whether or not the script tags should be cached
    * You can now specify a template to check for before adding the script tag
        * Script tags should not be used if the theme has a valid template
     * Change the name from `scripttag_manager` to `script_tag_manager`

https://github.com/user-attachments/assets/59a230c4-5bd4-4fef-902f-02c594846958

### Things to focus on

1. Is the documentation clear?
2. Has functionality been restored?
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@ Unreleased
 ----------
 - Adds a `script_tag_manager` that will automatically create script tags when the app is installed. [1948](https://github.com/Shopify/shopify_app/pull/1948)
 
+22.5.2 (March 14, 2025)
+----------
+- Add support for a new trusted Shopify domain [#1949](https://github.com/Shopify/shopify_app/pull/1949)
+- Fix broken links to OAuth access scopes documentation [#1944](https://github.com/Shopify/shopify_app/pull/1944)
+- Remove console deprecation warnings from boot [#1924](https://github.com/Shopify/shopify_app/pull/1924)
+- Removes extraneous tag in "Install app" form [#1927](https://github.com/Shopify/shopify_app/pull/1927)
+
 22.5.1 (December 11, 2024)
 ----------
 - Fix Rails [CVE-2024-54133](https://github.com/rails/rails/commit/3da2479cfe1e00177114b17e496213c40d286b3a) [1929](https://github.com/Shopify/shopify_app/pull/1929)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    shopify_app (22.5.1)
+    shopify_app (22.5.2)
       activeresource
       addressable (~> 2.7)
       jwt (>= 2.2.3)
@@ -122,7 +122,7 @@ GEM
     mocha (2.0.2)
       ruby2_keywords (>= 0.0.5)
     multi_xml (0.6.0)
-    net-imap (0.4.17)
+    net-imap (0.4.19)
       date
       net-protocol
     net-pop (0.1.2)
@@ -155,7 +155,7 @@ GEM
       pry (~> 0.13)
     public_suffix (5.0.1)
     racc (1.8.1)
-    rack (2.2.10)
+    rack (2.2.13)
     rack-test (2.1.0)
       rack (>= 1.3)
     rails (6.1.7.9)
--- a/lib/shopify_app/utils.rb
+++ b/lib/shopify_app/utils.rb
@@ -8,6 +8,7 @@ module ShopifyApp
         "myshopify.io",
         "myshopify.com",
         "spin.dev",
+        "shop.dev",
       ].freeze
 
       def sanitize_shop_domain(shop_domain)
--- a/lib/shopify_app/version.rb
+++ b/lib/shopify_app/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module ShopifyApp
-  VERSION = "22.5.1"
+  VERSION = "22.5.2"
 end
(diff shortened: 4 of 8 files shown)


lib/shopify_app/managers/script_tags_manager.rb
```diff
@@ -0,0 +1,370 @@
...
+        return nil if themes.empty?
+
+        themes.first
+      rescue => e
+        ShopifyApp::Logger.warn("Failed to fetch active theme: #{e.message}")
+        nil
+      end
+    end
+
+    def template_supports_app_blocks?(theme_id, template_type)
```
sle-c, line 136: can we break down this method to smaller ones and handle exceptions in each method instead of multiple calls here?

lib/shopify_app/managers/script_tags_manager.rb
```diff
@@ -0,0 +1,370 @@
...
+        }
+      QUERY
+
+      begin
+        response = client.query(query: query, variables: variables)
+
+        if response.body["data"]["scriptTagCreate"]["userErrors"].any?
+          errors = response.body["data"]["scriptTagCreate"]["userErrors"]
+          error_messages = errors.map { |e| "#{e["field"]}: #{e["message"]}" }.join(", ")
+          raise ::ShopifyApp::CreationFailed, "ScriptTag creation failed: #{error_messages}"
```
sle-c, line 284: some of the exception we raise above are not wrapped around a class like this one. Should we wrap them in error classes like this so it's more explicit?

lib/shopify_app/managers/script_tags_manager.rb
```diff
@@ -0,0 +1,370 @@
...
+      QUERY
+
+      begin
+        response = client.query(query: query)
+
+        # Check for errors in the response
+        if response.body["errors"].present?
+          ShopifyApp::Logger.warn("GraphQL error fetching script tags: #{response.body["errors"].map do |e|
+                                                                           e["message"]
+                                                                         end.join(", ")}")
```
sle-c, line 350: this formatting is kinda strange 🤔

lib/shopify_app/managers/script_tags_manager.rb
```diff
@@ -0,0 +1,370 @@
...
+          return []
+        end
+
+        # Handle nil data or missing structure
+        return [] unless response.body["data"] &&
+          response.body["data"]["scriptTags"] &&
+          response.body["data"]["scriptTags"]["edges"]
```
sle-c, line 357: shouldn't the structure be guaranteed here as well?

test/dummy/config/initializers/shopify_app.rb
```diff
@@ -13,7 +13,7 @@ def self.call
       config.myshopify_domain = "myshopify.com"
       config.api_version = ShopifyAPI::LATEST_SUPPORTED_ADMIN_VERSION
       config.billing = nil
-      config.scripttags = nil
+      config.script_tags = nil
```
sle-c, line 16: wondering if this is gonna make this incompatible with previous versions 🤔

lib/shopify_app/managers/script_tags_manager.rb
sle-c: also we should handle `userErrors` response

lib/shopify_app/managers/script_tags_manager.rb
lizkenyon: I wrapped all of the other errors with the `ShopifyAPI::Errors::InvalidGraphqlRequestError`. Looking at the rest of the library code, we aren't super consistent with this type of scenarios. But this seemed like that closest fit.

lib/shopify_app/managers/script_tags_manager.rb
lizkenyon: Fixed.

lib/shopify_app/managers/script_tags_manager.rb
```diff
@@ -0,0 +1,328 @@
...
+        },
+      }
+
+      response = client.query(query: SCRIPT_TAG_CREATE_MUTATION, variables: variables)
+
+      if response.body["data"]["scriptTagCreate"]["userErrors"].any?
+        errors = response.body["data"]["scriptTagCreate"]["userErrors"]
+        error_messages = errors.map { |e| "#{e["field"]}: #{e["message"]}" }.join(", ")
+        raise ::ShopifyApp::CreationFailed, "ScriptTag creation failed: #{error_messages}"
+      end
```
sle-c, line 282: can we handle `response.body["errors"]` too?

lib/shopify_app/managers/script_tags_manager.rb
```diff
@@ -0,0 +1,328 @@
...
+    rescue ShopifyAPI::Errors::HttpResponseError => e
+      raise ::ShopifyApp::CreationFailed, e.message
+    end
+
+    def delete_script_tag(tag)
+      client = graphql_client
+
+      variables = { id: tag["id"] }
+
+      client.query(query: SCRIPT_TAG_DELETE_MUTATION, variables: variables)
```
sle-c, line 294: can we handle `response.body["errors"]` for this query?

sle-c (approved):
some minor comments, otherwise LGTM

PR #1974, opened 2025-05-06
Exchange id_token when it is present

### LINKED ISSUE #1886: Token exchange not triggered after app reinstall
I'm using token exchange auth and managed installs. 

When I install the app for the first time - the new `Shop` record is created and `ShopifyApp::Auth::TokenExchange` is called.

When I re-install the app - the `Shop` record is already present in DB and `ShopifyApp::Auth::TokenExchange` not called automatically so `shopify_token` is not updated and `post_authenticate_jobs` are not called.

## Expected behavior

`ShopifyApp::Auth::TokenExchange` should be called after app re-install.

## Actual behavior

`ShopifyApp::Auth::TokenExchange` not called after re-install.

## Demo

https://github.com/user-attachments/assets/8acbbaaf-a059-439a-87e7-79cc2e2599e5

jduff:
### What this PR does

Fixes #1886

There are a couple of cases myself and others have ran into where the access token or scopes are not updated when using token exchange:
- When the user uninstalls and reinstalls the app (and you do not hard delete the shop for data retention reasons)
- When managed app install scopes are updated

In my testing when these cases occur we get an `id_token` back in the url params and this is a good indication that something has changed. Based on #1886 it seems like others have found this works for them as well. Triggering `retrieve_session_from_token_exchange` in this case causes all session storage and (I believe but didn't test) after authenticate triggers to run.

I haven't looked too much into the authentication header but it's possible that checking `shopify_id_token` is present would be a better approach to take into account that header having the id_token. I'm not sure what situations that header is present in through and my assumption is that most of those cases we don't necessarily need to update the scopes/access_token and the cases where we do want to update them it will be included in the url params.

### Things to focus on

1. Make sure my assumptions about when this should trigger are correct.
2. Check if it would make sense to trigger this when the authentication header is present as well.

### Checklist

### DIFF
--- a/lib/shopify_app/controller_concerns/token_exchange.rb
+++ b/lib/shopify_app/controller_concerns/token_exchange.rb
@@ -17,7 +17,7 @@ module ShopifyApp
     ].freeze
 
     def activate_shopify_session(&block)
-      retrieve_session_from_token_exchange if current_shopify_session.blank? || should_exchange_expired_token?
+      retrieve_session_from_token_exchange if should_exchange_token?
 
       ShopifyApp::Logger.debug("Activating Shopify session")
       ShopifyAPI::Context.activate_session(current_shopify_session)
@@ -27,6 +27,10 @@ module ShopifyApp
     ensure
       ShopifyApp::Logger.debug("Deactivating session")
       ShopifyAPI::Context.deactivate_session
+    end
+
+    def should_exchange_token?
+      current_shopify_session.blank? || should_exchange_expired_token? || params[:id_token].present?
     end
 
     def should_exchange_expired_token?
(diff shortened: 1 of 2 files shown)


kirillplatonov:
I don't think it's a good permanent solution. `id_token` is always included in params when the app is opened in Shopify Admin, regardless of installation/scope changes. With these changes, we'll end up making token exchange every time when the app is opened. It will slow down app loading by 200-300ms, which will impact perf metrics and our abilities to maintain Build for Shopify status.

As mentioned in #1886, I still believe we need some proper indication from App Bridge, that the app needs to update token after installation/scopes changes. The easiest solution would be new parameter (eg `reload_token=1`). This way, we could make token exchange only when it's **really** needed.

Issue #1710, opened 2023-08-02
Sorbet host check fails


resistorsoftware:
# Issue summary

Installing a new app and the Sorbet is failing to find the host even though it is set. 

So what is the culprit here? ENV['HOST'] exists as https://something.herokuapp.com/ for the Shopify App context setup.

- `shopify_api` version:  12.5.0
- `shopify_app` version:   21.4.1
- Ruby version:  3.2
- Operating system:  *nix

```
// Paste any relevant logs here
```
Filter chain halted as :check_shop_known rendered or redirected
Completed 302 Found in 3ms (ActiveRecord: 0.7ms | Allocations: 753)
FATAL -- : [efae465b-28b7-4517-8731-1aa969182bf2]
TypeError (Passed nil into T.must):
sorbet-runtime (0.5.10827) lib/types/_types.rb:222:in must' shopify_api (12.5.0) lib/shopify_api/context.rb:175:in host_scheme'

## Expected behavior

App should install

## Actual behavior

errors out on Sorbet type check on host

resistorsoftware:
For me this error only comes up if I accidentally add a trailing slash like my example above. That slash kills it. I removed the slash, it works.

codev009:
Removing the trailing slash isn't working for me.

resistorsoftware:
Are you using App Bridge 4? I think you can ignore host issues once you go with that. I no longer set any HOST in my localhost and it works fine with App Bridge 4.

Issue #1526, opened 2022-10-05
Upgrade docs are inadequate


jagthedrummer:
### Description

I'm trying to upgrade to v19, which seems to break almost everything, and the docs are completely inadequate. They basically say "lots of stuff moved to shopify_api, so go read that changelog" and that changelog basically says "we changed nearly everything, read the new library code and adjust your app accordingly".

### Steps to Reproduce

1. Have an app that works with v18
2. Upgrade to v19
3. 💥 

**Expected behavior:**

There should be docs that explicitly state which objects/methods have changed and it should suggest the new syntax to use.

**Actual behavior:**

Developers are left to fend for themselves.

**Reproduces how often:**

100%

### Browsers

N/A

### Gem versions

19

### Additional Information

Having been a Shopify app developer for 7+ years I find this to be extremely hostile to developers, and it makes me question whether our time enriching the Shopify ecosystem is appreciated by Shopify.

### Security

- [x] I have redacted any private information from my logs or code snippets.

asecondwill:
Yikes!  now i dont want to update anything!

matthewford:
+1 the docs are terrible

slatem:
+1 - absolutely horrible experiencing updating from 18 -> 19+

yogodoshi:
I had the same issue. I'm now trying to create a new app from scratch and moving the code over but I'm also having issues with this approach, even with a new app I'm getting errors that are difficult to debug like:

```
Processing by HomeController#index as HTML
Redirected to https://4ebe-2104-14d-5c59-4153-5c2-66ba-1ec3-1197.sa.ngrok.io/login
Filter chain halted as :check_shop_domain rendered or redirected
Completed 302 Found in 16ms (ActiveRecord: 0.0ms | Allocations: 1944)


  
TypeError (Passed `nil` into T.must):
  
sorbet-runtime (0.5.10626) lib/types/_types.rb:218:in `must'
shopify_api (12.2.1) lib/shopify_api/context.rb:157:in `host_scheme'
sorbet-runtime (0.5.10626) lib/types/private/methods/call_validation.rb:256:in `bind_call'
sorbet-runtime (0.5.10626) lib/types/private/methods/call_validation.rb:256:in `validate_call'
sorbet-runtime (0.5.10626) lib/types/private/methods/_methods.rb:275:in `block in _on_method_added'
[...]

PR #1998, opened 2025-09-19
Use ENV.fetch to get HOST instead of relying on useless Sorbet error

sshaw:
### What this PR does

The auto-generated initializer has various starting time checks to prevent app misconfiguration before running. This did not apply to ShopifyAPI::Context's host: keyword. If the HOST env var was not present, a unhelpful "TypeError (Passed `nil` into T.must)" error was raised by Sorbet. Now if there is no HOST env var present, the app will fail to run, like it does with other checks, and the user will receive a helpful message.

As mentioned in #1710 and #1526 

Sorbet 🤮

### Things to focus on

In generated project: `config/inititalizers/shopify_app.rb`

### Checklist

> - [ ] Update `CHANGELOG.md` if the changes would impact users 

This should be done by the maintainer at release time, not by contributors.

### DIFF
--- a/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
+++ b/lib/generators/shopify_app/install/templates/shopify_app.rb.tt
@@ -47,7 +47,7 @@ Rails.application.config.after_initialize do
       api_key: ShopifyApp.configuration.api_key,
       api_secret_key: ShopifyApp.configuration.secret,
       api_version: ShopifyApp.configuration.api_version,
-      host: ENV['HOST'],
+      host: ENV.fetch('HOST'),
       scope: ShopifyApp.configuration.scope,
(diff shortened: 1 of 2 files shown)


lib/generators/shopify_app/install/templates/shopify_app.rb.tt
```diff
@@ -47,7 +47,7 @@ Rails.application.config.after_initialize do
       api_key: ShopifyApp.configuration.api_key,
       api_secret_key: ShopifyApp.configuration.secret,
       api_version: ShopifyApp.configuration.api_version,
-      host: ENV['HOST'],
+      host: ENV.fetch('HOST'),
```
sshaw, line 50: Maybe this is only a 90% improvement because what does Sorbet regurgitate when it's an empty string? A good ol user-raised `ArgumentError` or `TypeError` with a nice message is best but, this is quick and simple.

PR #2014, opened 2025-10-16
Allow custom handler to be passed when registering a webhook

jonfrench:
### What this PR does

When an app uses a custom WebhooksController instead of using the Gem provided WebhooksController, the app assumes responsibility for receiving and processing web hook requests, and thus does not need to register ActiveJob `handlers` which match the name of the webhook's topic as required by WebhooksController.

There is a bug in `WebhookManager` which requires the app to [maintain an ActiveJob class matching the name of the topic](https://github.com/Shopify/shopify_app/blob/main/lib/shopify_app/managers/webhooks_manager.rb#L79) even if the app uses a custom WebhooksController.

In this Pull Request, I add an additional `WebhookManager` webhooks configuration property called `handler` which is the name of the ActiveJob class to enqueue if using the default WebhooksController or `nil` to unset the `handler`. The latter feature - the ability to set `handler: nil` will fix the bug.

Because this introduces a new configuration property but maintains backward compatibility with current configuration (_not_ setting the `handler` property behaves exactly the same), I rely on unit tests to assert the change.

### Things to focus on

1. 
2. 
3. 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Unreleased
 - Handle invalid token when adding redirection headers [#1945](https://github.com/Shopify/shopify_app/pull/1945)
 - Handle invalid record error for concurrent token exchange calls [#1966](https://github.com/Shopify/shopify_app/pull/1966)
 - Add App Bridge CDN URL to CSP script-src directive for embedded apps to ensure compatibility with strict CSP configurations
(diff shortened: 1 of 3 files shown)


jonfrench:
Arg. I just realized that you cannot register a http webhook in the WebhookRegistry without a handler:

https://github.com/Shopify/shopify-api-ruby/blob/8b20cf374f41f5c06ce423d173e67cefdb4cb068/lib/shopify_api/webhooks/registry.rb#L36

PR #2020, opened 2025-10-28
Fix loading issues with modern rails versions

that-jill:
### What this PR does

The jobs used here are referenced in custom ActiveJob serializers in other apps (specifically, merchant-to-merchant). Due to how they were being included and loaded here and some changes in Rails 8.1 (registering custom serializers earlier), those serializers were not being properly registered. Updating to a more modern version of rails and allowing rails to handle to load paths fixed the problem.

It's worth saying here: rails maintainers do not intend for you to load things like ActiveRecord::Base or ActiveJob::Base during initialization. Doing so is not a good practice and may result in unintended bugs / side effects.

### Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    shopify_app (22.5.2)
+    shopify_app (22.6.0)
       addressable (~> 2.7)
       rails (>= 7.1, < 9)
       redirect_safely (~> 1.0)
--- a/lib/shopify_app/version.rb
+++ b/lib/shopify_app/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module ShopifyApp
-  VERSION = "22.5.2"
+  VERSION = "22.6.0"
 end


shopify_app.gemspec
```diff
@@ -10,25 +10,27 @@ Gem::Specification.new do |s|
...
-  s.add_runtime_dependency("rails", "> 5.2.1")
+  s.add_runtime_dependency("rails", "> 7")
   s.add_runtime_dependency("redirect_safely", "~> 1.0")
   s.add_runtime_dependency("shopify_api", ">= 14.7.0", "< 15.0")
   s.add_development_dependency("byebug")
+  s.add_development_dependency("csv")
   s.add_development_dependency("jwt", ">= 2.2.3")
   s.add_development_dependency("minitest")
   s.add_development_dependency("mocha")
+  s.add_development_dependency("mutex_m")
```
sle-c: curious why we need to require `csv` and `mutex_m` as dev dependency?

shopify_app.gemspec
```diff
@@ -10,25 +10,27 @@ Gem::Specification.new do |s|
...
   s.summary     = "This gem is used to get quickly started with the Shopify API"
 
-  s.required_ruby_version = ">= 3.1"
+  s.required_ruby_version = ">= 3.2"
 
   s.metadata["allowed_push_host"] = "https://rubygems.org"
 
   s.add_runtime_dependency("addressable", "~> 2.7")
-  s.add_runtime_dependency("rails", "> 5.2.1")
+  s.add_runtime_dependency("rails", "> 7")
```
lizkenyon: Could you clarify what is the minimum version we would need to support to fix the issue?

In the changelog we specify rails 7.1

lizkenyon (commented):
Just a few clarifying questions! 

Could you clarify what you mean by merchant to merchants apps?

Let's add to the [upgrading guide](https://github.com/Shopify/shopify_app/blob/main/docs/Upgrading.md#unreleased) a link to the rails documentation for changes that may be required for apps that are upgrading their ruby version.

In our app template we are already using rails 7.1, I tested this branch there, and things are working as expected.

shopify_app.gemspec
lizkenyon: Thanks for taking the initiative here! I can take it from here! 

Thank you for clarifying the intended version.

I tested this with our app template and there were no problems with it.

PR #2058, opened 2026-02-26
fix: replace CGI.parse with Rack::Utils.parse_query for Ruby 4.0 compatibility

### LINKED ISSUE #2057: Ruby 4.0 compatibility: CGI.parse removed in LoginProtection#return_address_with_params
## Summary

`LoginProtection#return_address_with_params` calls `CGI.parse` which was removed from Ruby's standard library in Ruby 4.0. This causes a `NoMethodError` on every OAuth callback, making authentication completely broken on Ruby 4.0.

## Affected versions

Confirmed broken in `shopify_app 22.5.2` and `23.0.1` (per the source). Ruby 4.0.1.

## Location

[`lib/shopify_app/controller_concerns/login_protection.rb`](https://github.com/Shopify/shopify_app/blob/main/lib/shopify_app/controller_concerns/login_protection.rb)

```ruby
def return_address_with_params(params)
  uri = URI(base_return_address)
  uri.query = CGI.parse(uri.query.to_s)       # ← NoMethodError on Ruby 4.0
    .symbolize_keys
    .transform_values { |v| v.one? ? v.first : v }
    .merge(params)
    .to_query
  uri.to_s
end
```

## Root cause

`CGI` was separated from Ruby's stdlib into a standalone gem in Ruby 3.4 (with a deprecation warning) and fully removed in Ruby 4.0. See [ruby/cgi#82](https://github.com/ruby/cgi/issues/82).

## Suggested fix

Replace with `Rack::Utils.parse_query`, which is already available as a transitive dependency of Rails, returns a plain hash directly (no array-unwrapping needed), and is idiomatic for query string parsing in a Rack/Rails context:

```ruby
def return_address_with_params(params)
  uri = URI(base_return_address)
  uri.query = Rack::Utils.parse_query(uri.query.to_s)
    .symbolize_keys
    .merge(params)
    .to_query
  uri.to_s
end
```

This is cleaner than the original because `Rack::Utils.parse_query` returns string values directly (not arrays), so the `.transform_values` unwrapping step is no longer needed.

### Alternatives considered

- **Add `gem 'cgi'` to the gemspec**: Would restore `CGI.parse` without code changes, but adds an external dependency purely to keep an awkward API that wraps every value in an array.
- **`URI.decode_www_form(...).to_h`**: Pure stdlib, no extra deps, but more verbose than the Rack alternative.

## Workaround

Apps running Ruby 4.0 can monkey-patch the method in an initializer:

```ruby
# config/initializers/shopify_app_ruby4_compat.rb
module ShopifyApp
  module LoginProtection
    private

    def return_address_with_params(params)
      uri = URI(base_return_address)
      uri.query = URI.decode_www_form(uri.query.to_s).to_h
                    .transform_keys(&:to_sym)
                    .merge(params)
                    .to_query
      uri.to_s
    end
  end
end
```

bvogel:
## Summary

Fixes #2057

`CGI.parse` was removed from Ruby's standard library in Ruby 4.0, causing a `NoMethodError` in `LoginProtection#return_address_with_params` on every OAuth callback. This breaks authentication entirely on Ruby 4.0 — the callback 500s before the session can be stored.

## Change

Replace `CGI.parse` with `Rack::Utils.parse_query`, which is already available as a transitive dependency of Rails (no new dependency introduced).

```ruby
# Before
uri.query = CGI.parse(uri.query.to_s)
  .symbolize_keys
  .transform_values { |v| v.one? ? v.first : v }
  .merge(params)
  .to_query

# After
uri.query = Rack::Utils.parse_query(uri.query.to_s)
  .symbolize_keys
  .merge(params)
  .to_query
```

`Rack::Utils.parse_query` returns a plain hash with string values directly, so the `.transform_values` unwrapping step (which existed to flatten `CGI.parse`'s array-of-strings format) is no longer needed — the result is also slightly simpler.

## Behaviour difference

For Shopify URL params, none. The only theoretical difference is duplicate keys: `CGI.parse` would collect them into an array, while `Rack::Utils.parse_query` keeps the last value. Shopify-generated query strings never have duplicate parameter names.

## Testing

Existing test coverage for `return_address_with_params` should confirm correctness. Manually verified on Ruby 4.0.1 + Rails 8.1.

### DIFF
--- a/lib/shopify_app/controller_concerns/login_protection.rb
+++ b/lib/shopify_app/controller_concerns/login_protection.rb
@@ -225,9 +225,8 @@ module ShopifyApp
 
     def return_address_with_params(params)
       uri = URI(base_return_address)
-      uri.query = CGI.parse(uri.query.to_s)
+      uri.query = Rack::Utils.parse_query(uri.query.to_s)
         .symbolize_keys
-        .transform_values { |v| v.one? ? v.first : v }
         .merge(params)
         .to_query
       uri.to_s


bvogel:
back-porting this to the v22 branch would help a lot, as we still need to migrate to v23 for our project

PR #2059, opened 2026-03-02
Validate host param in generated HomeController template

amcaplan:
## Summary
- Align generated home controller template with the existing validation in `redirect_to_embed_app_in_admin`

## Test plan
- [x] Existing tests pass
- [x] Generator test updated

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 Unreleased
 ----------
+- [Patch] Validate host param in generated HomeController template to prevent open redirect
 - [Patch] Fix sorbet errors in generated webhook handlers
 
 23.0.1 (December 22, 2025)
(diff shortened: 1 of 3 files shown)


byrichardpowell (approved):

lizkenyon (approved):
TY! Looks good. Could you add a post in the CHANGELOG.md

We should also make a similar fix in the [template](https://github.com/Shopify/shopify-app-template-ruby/blob/main/web/app/controllers/home_controller.rb) itself.

amcaplan:
@lizkenyon I added a changelog entry. Also opened a PR [here](https://github.com/Shopify/shopify-app-template-ruby/pull/198) for the app template.

PR #2064, opened 2026-03-25
Update controller concern docs

zzooeeyy:
Rephrasae `EnsureInstalled` as "Installation Check Only" and `EnsureHasSession` as "Authenticated Requests" to make clear that `EnsureInstalled` does not authenticate the requester.


lizkenyon (approved):
Thank you for the improvement here! ⭐

PR #2073, opened 2026-04-30
Fix declarative webhook generator route path

### LINKED ISSUE #1977: Missing WebhookHandler
Hey,

I'm trying to set up Webhooks, following the documentation:
https://github.com/Shopify/shopify_app/blob/main/docs/shopify_app/webhooks.md#consuming-app-specific-webhooks-events

It seems that the recommended approach is to update `shopify.app.toml` and then run:

```ruby
rails g shopify_app:add_declarative_webhook --topic carts/update --path webhooks/carts_update
```

For any Webhook to subscribe to, which according to the documentation should set up a controller and a job.

When I run a test Webhook via the CLI, I get this error though:

```
ShopifyAPI::Errors::NoWebhookHandler (No webhook handler found for topic: carts/update.)
```

afurm:
### What this PR does

Fixes #1977.

This updates the `shopify_app:add_declarative_webhook` generator so the generated Rails route matches the configured webhook `--path`.

Previously, running:

```bash
rails g shopify_app:add_declarative_webhook --topic products/update --path webhooks/product_update
```

could add only the basename route under an existing `namespace :webhooks`, which meant apps using the default `/api` route scope served `/api/webhooks/product_update` instead of `/webhooks/product_update`. Shopify then posted to the configured `/webhooks/product_update` path, which could fall through to `ShopifyApp::WebhooksController` and raise `ShopifyAPI::Errors::NoWebhookHandler`.

The generator now creates an explicit route for the configured path and inserts it before the root-mounted ShopifyApp engine.

```bash
bundle exec rake test TEST=test/generators/add_declarative_webhook_generator_test.rb
```

### Things to focus on

1. The generated route uses the full configured webhook path.
2. The route is inserted before `mount ShopifyApp::Engine, at: "/"`.
3. The generator test covers the route path and ordering.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 Unreleased
 ----------
+- [Patch] Fix declarative webhook generator routes to match the configured webhook path
 - [Patch] Validate host param in generated HomeController template to prevent open redirect
 - [Patch] Fix sorbet errors in generated webhook handlers
 
--- a/lib/generators/shopify_app/add_declarative_webhook/add_declarative_webhook_generator.rb
+++ b/lib/generators/shopify_app/add_declarative_webhook/add_declarative_webhook_generator.rb
@@ -31,14 +31,27 @@ module ShopifyApp
       end
 
       def add_webhook_route
-        route = "\t\t\tpost '#{file_name}', to: '#{file_name}#receive'\n"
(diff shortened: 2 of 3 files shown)


kennyadsl:
I proposed an alternative approach in #2076, which should fix the issue as well.

PR #2087, opened 2026-09-22
Simplify Gardener Slack notifications to linked titles

lizkenyon:
## What changed

- Gardener Slack notifications now post the repo name and escaped, linked title.
- Link and media unfurling are disabled.
- Existing labeling behavior is unchanged.
- Uses the title-only shape from [Shopify/cli#8540](https://github.com/Shopify/cli/pull/8540), with a repo prefix for the shared channel.

Workflow-only change; package and runtime behavior is unchanged.

## Verified

- YAML parsing
- Synthetic title-only payload check
- `git diff --check`

Closes shop/issues#86822


Suleimanlatrsh (approved):
Tested, works great. Thanks!

